PR #7, opened 2018-09-11
Add pdalpra/gh4s to the repos list

pdalpra:
It's still a very young project, I would understand if you'd prefer to test on more mature repositories ;)


fthomas:
@pdalpra Your project turned out to be interesting because scala-steward failed to update sttp to version 1.3.3 becaus of sttp's unusual artifact names.

PR #23, opened 2018-09-14
@ maximn's repos

maximn:


fthomas:
@maximn, my sbt launcher fails to load coding-interview-questions-scala and TaxCalculator with errors like this:
```
error: bad symbolic reference. A signature in Implicits.class refers to term Remove
in package sbt which is not available.
It may be completely missing from the current classpath, or the version on
the classpath might be incompatible with the version used when compiling Implicits.class.
```
I think this is because of the old sbt versions these projects use. scala-steward would be able to check for updates if they would use a newer sbt version.

maximn:
Upgraded both of them to SBT 1.2.3

fthomas:
:+1:

PR #29, opened 2018-09-17
Add ChristopherDavenport Libraries

ChristopherDavenport:
My Public Libraries. I can start with a smaller set if this is too many.


fthomas:
I don't mind, it is @scala-steward who has to do all the hard work :-)

PR #37, opened 2018-09-18
Add http4s.g8 template

rossabaker:
This won't be a steward to the template itself, but for the template's build.


fthomas:
It would be great if we could keep the templates up-to-date too.

Issue #46, opened 2018-09-25
Support nonstandard project layout.


hepin1989:
Like the Play framework, which put the real sbt project in `/framework` folder.

PR #90, opened 2018-10-16
Add akka/akka-samples repo

raboof:


fthomas:
@raboof Scala steward right now requires a sbt project in the top-level directory of a repository, so for the time being it won't find any updates here. #46 tracks this feature request.

PR #113, opened 2018-10-29
add "--recursive" option when "git clone"

xuwei-k:


fthomas:
Thanks, @xuwei-k! I'm currently on vacation and will merge when I'm back.

fthomas:
@all-contributors please add @xuwei-k for code

fthomas:
@xuwei-k Do you remember what the motivation for this changes was? Do some repos require submodules to be present so that the build can be loaded? I'm asking because it caused some issues (#339 and #1779) and was wondering why we clone submodules in the first place.

fthomas:
> Do some repos require submodules to be present so that the build can be loaded?

This seems to be the case, see #810 for example.

PR #133, opened 2018-11-06
Add Finch

travisbrown:
@vkostyukov, @rpless, @sergeykolbasov, what do you think? I've started using this for circe and some other projects and while it's not perfect yet, I definitely think it's going to be helpful.


vkostyukov:
I'm 👍on this! Wanted to do that for a long time. Thank you, Travis!

PR #142, opened 2018-11-07
Add More ChristopherDavenport Libraries

ChristopherDavenport:
A few more that I've built that it would be nice to keep externally updated.


fthomas:
keep ’em coming :-)

Issue #123, opened 2018-11-04
Reduce workload


fthomas:
Currently scala-steward checks every repository for new dependencies on every run. Checking for updates is a costly operation since it means cloning the repository, starting sbt, and checking for library and plugin updates. Even for small repositories which are up-to-date this process takes roughly a minute on my server. This is not viable if scala-steward should be able to handle 1000+ repositories which I very much like to achieve.

My current idea to reduce the overall workload is the following:

I propose to skip checking repositories where scala-steward would not open new PRs or update existing PRs. This would require to persist some infos about the repos scala-steward has worked on. E.g if we know of each repo
 * the current SHA1 of the base branch
 * the current list of library and plugin dependencies
 * all PRs scala-steward has created

we should be able to decide if we need to process a repository without cloning it and running `sbt dependencyUpdates`. The SHA1 can be used check if the upstream repository has changed and our existing PRs need to be updated or if the list of dependencies has changed.
Before working on any repository we would need to check if any of the libraries or plugins scala-steward knows about has a newer version and then only process repositories whose dependencies have updates.

fthomas:
I just checked the runtime of the latest run:
```
2018-11-04 15:55:29,111 INFO   --- Total time: 3h 52m 54s 445ms ---
```
That is definitely too long for 154 repositories.

fthomas:
This has been implemented in the two PRs referencing this issue.

PR #147, opened 2018-11-10
Ignore updates which have existing up-to-date PRs

fthomas:
This is another part of #123 where we ignore updates if there is already an existing and up-to-date PR for an update. This should drastically reduce the number of repositories that need to be processed on each run.


fthomas:
Although this code is still kind of WIP, I'll merge to see how this whole idea works out.

Issue #149, opened 2018-11-13
Integration with scalafix for automated code fixes


Daenyth:
Is there a way to integrate scala-steward with scalafix, so that when a library maintainer supplied a scalafix migration from one version to another (such as the one in the works for fs2 0.10 to 1.0), that steward can automatically apply it in the PR along with the build change?

fthomas:
That's a great idea! How would one invoke the scalafix migration for fs2? There is nothing yet in scala-steward to do this but let's figure out what would be needed to do it.

Daenyth:
I'm not sure what's the best way to do it - I think for MVP, perhaps a config file on the steward side can list an index of `(artifactId, fromVersion, toVersion, scalaFixDependencyArtifactId, scalafixInvocationCommand)`. The dependency being for cases like the http4s 0.20 upgrade, where you will add a dependency to build.sbt in order to run the fix. It could also be used for the git address in the case of github-style fixes.

Once the infrastructure is in place to run it at all, the manual config driven approach could be revisited and an automatic method could be written.

fthomas:
Sounds like a plan. It seems the github-style fixes are easier to run since they don't require changes to the `build.sbt`. At least this is the impression I got from reading https://github.com/typelevel/cats/blob/master/scalafix/README.md which also suggests that github-style fixes only require that sbt-scalafix is added to the build.

Daenyth:
Github-style fixes seem like a good MVP target

olafurpg:
Awesome idea! For now, it's probably easiest to start with `github:org/repo` style rewrites but once https://github.com/scalacenter/scalafix/issues/905 is fixed it will be possible to run published scalafix linter rules with one-liners from the shell.

amarrella:
Awesome idea! I'll refactor the fs2 scalafix in the works to be github compatible for now so we can test it as well :)

amarrella:
Ok you can already test the fs2 rule here: 
```
sbt ";scalafixEnable;scalafix github:amarrella/fs2/v1?sha=672ea4f9" 
```

amarrella:
Maybe we could fork and try on this repo? https://github.com/pauljamescleary/scala-pet-store

fthomas:
Thanks, @amarrella. Let's make this a little bit more concrete. Say we make a list of updates and their corresponding rewrite rules using this data type:
```scala
case class Migration(
  groupId: String,
  artifactId: String,
  fromVersion: String,
  toVersion: String,
[...]

PR #153, opened 2018-11-15
scalafix integration

fthomas:
Towards #149. There is not much here yet, but I think it makes sense to open this PR anyway as place for further discussion.

/cc @amarrella @Daenyth


modules/core/src/main/scala/eu/timepit/scalasteward/scalafix/Migration.scala
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package eu.timepit.scalasteward.scalafix
+
+import cats.data.NonEmptyList
+
+final case class Migration(
+    groupId: String,
+    artifactIds: NonEmptyList[String],
+    newVersion: String,
```
ShaneDelmore: I'd put back the from version here. If someone is multiple versions behind they may need to take multiple passes to get up to date. For example a cats 0.8 -> 0.9 rule could be applied, then a 0.9 -> 1.0, but 0.8 -> 1.0 in one step may fail.

ShaneDelmore: Maybe minVersion?

fthomas: @ShaneDelmore That makes sense. Do scalafix rewrites specify somewhere what the minimal version is a rewrite can be applied to? Or is this implicit or sometimes encoded in the name of the rewrite? From the `fs2/v1` rewrite I can only assume that it will migrate code to version 1.0.0 of the library but not what the ranges of versions is at which I can safely apply this rewrite.

amarrella: No scalafix itself is not aware of the version. I think it's a good idea to specify the *from* version in this case since it's going to be automatic. Running it on a previous version won't do much damage, but probably the fix won't be complete either.

fthomas:
scala-steward just created its first PR using a scalafix migration!
:sparkles: https://github.com/fthomas/scala-pet-store/pull/13 :sparkles:

Unfortunately the scalafix changes are not nicely formatted so we probably need to run scalafmt before creating the PR.

fthomas:
The latest PR created with the changes on this branch is https://github.com/fthomas/scalafix-test/pull/4/files

PR #161, opened 2018-11-20
Suggestion: Add CLI to override config values

regadas:
Hi @fthomas I sketched this really quick and it might be a bit rough, let me know what you think!

### DIFF
--- a/modules/core/src/main/scala/eu/timepit/scalasteward/application/Cli.scala
+++ b/modules/core/src/main/scala/eu/timepit/scalasteward/application/Cli.scala
@@ -17,8 +17,8 @@
 package eu.timepit.scalasteward.application
 
 import caseapp._
-import cats.effect.Sync
 import cats.implicits._
+import eu.timepit.scalasteward.util.ApplicativeThrowable
 
 trait Cli[F[_]] {
   def parseArgs(args: List[String]): F[Cli.Args]
@@ -28,7 +28,6 @@ object Cli {
 
   final case class Args(
       workspace: String,
-      reposDir: Option[String],
       reposFile: String,
       gitAuthorName: String,
       gitAuthorEmail: String,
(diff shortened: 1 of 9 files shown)


modules/core/src/main/scala/eu/timepit/scalasteward/application/Cli.scala
```diff
@@ -0,0 +1,47 @@
...
+      reposDir: Option[String],
+      reposFile: String,
+      gitAuthorName: String,
+      gitAuthorEmail: String,
+      githubApiHost: String,
+      githubLogin: String,
+      gitAskPass: String
+  )
+
+  def create[F[_]](implicit F: Sync[F]): Cli[F] = new Cli[F] {
```
fthomas: It seems to me that `ApplicativeError[F, Throwable]` would be enough here. There is also a handy `ApplicativeThrowable` for the former in the `util` package object

modules/core/src/main/scala/eu/timepit/scalasteward/application/Config.scala
```diff
@@ -45,6 +45,8 @@ import scala.sys.process.Process
   */
 final case class Config(
     workspace: File,
+    reposFile: File,
+    reposDir: File,
```
fthomas: What is the reason for adding `reposDir`? Why would you want to override the default `workspace / "repos"`?

modules/core/src/main/scala/eu/timepit/scalasteward/application/Cli.scala
regadas: Good point!

modules/core/src/main/scala/eu/timepit/scalasteward/application/Config.scala
regadas: yeah, on a second thought I guess it doesn't make that much sense now 😄

PR #164, opened 2018-11-23
Add docker packaging

regadas:
@fthomas wdyt? 

I think later you can eventually remove `sbt-assembly` or even remove it now, let me know.


fthomas:
@regadas I've never used Docker and have only a superficial idea what it does. What would be the advantages over the the current fat JAR?

regadas:
@fthomas  fat jars are normally used for "portability", however because you merge everything in one single jar, if your not careful, can sometimes lead to classpath path issues, problems with transitive deps, etc. The best practice is to keep your jar "thin" and have your deps normally in a separate folder that you later add to the classpath. 

`sbt-native-packager` makes this all easy by creating bash scripts that do all the work (adds all deps to the classpath) and takes care of portability by allowing you to package your artifact in different ways (zip, tarball, docker).

Docker, in this case, will allow you to deploy and run `scala-steward` anywhere since the docker image provides everything needed. Docker provides a good isolation layer between your app and the system where it's running.

fthomas:
That are good points, thanks @regadas for the explanation! The isolation between scala-steward and the host system is especially useful since scala-steward is executing foreign code when it runs sbt in the repos.

I'll merge this now and remove `sbt-assembly` once I've verified that it works on my server.

regadas:
@fthomas 👍 😄

Issue #101, opened 2018-10-21
Propose sbt updates


fthomas:
Scala steward should also propose sbt updates and not only library and plugin upates.

jhnsmth:
I'd like to volunteer myself for this feature if no one else is already working on it

fthomas:
I'm not aware of anyone else who is working on this, so go for it!

jhnsmth:
Thanks for the quick answer. I will give it a try during next week.

PR #172, opened 2018-12-01
Update sbt to 1.2.7

fthomas:


hepin1989:
@fthomas it should be updated by scala-steward :)

fthomas:
@hepin1989 Yes, that would be great. There is already a ticket for that: #101

PR #174, opened 2018-12-01
Add Izumi Project monorepo

kaishh:


fthomas:
Once scala-steward finds an update it tries to apply it in all .sbt and .scala files. So if we're lucky it should get along with that scheme. We'll find out soon. :-)

PR #185, opened 2018-12-05
Add daandi's projects

daandi:
added nice-uuid and nice-url-encode-decode


fthomas:
@daandi It seems that all your dependencies and plugins are up-to-date. At least scala-steward couldn't find any updates.

Issue #180, opened 2018-12-04
Add config option to disable firejail


fthomas:
`ProcessAlg.execSandboxed` is an alternative to `ProcessAlg.exec` that runs programs in a sandbox. If one fully trusts the repos scala-steward is working on `execSandboxed` could be just an alias for `ProcessAlg.exec`. It would be great to have a config option for this case.

fthomas:
Fixed by #200.

PR #200, opened 2018-12-18
Config for sandbox execution

ArulselvanMadhavan:
My attempt to work on https://github.com/fthomas/scala-steward/issues/180 

I can get the tests to compile in my machine but the tests fail when they try to read files with test data like "create-fork.json"

Submitting this for initial feedback.

## TODO:
1. I need to figure out how to get the tests to pass in my machine
2. Need to write unit test for this new config flag.
3. I am using scalafmt 1.5.1 . I ran it against both src and test. Still I can't get past the scalafmtCheck in travis CI.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -36,7 +36,8 @@ object Cli {
       gitAskPass: String,
       signCommits: Boolean = false,
       whitelist: List[String] = Nil,
-      readOnly: List[String] = Nil
+      readOnly: List[String] = Nil,
+      execSandbox: Boolean = true
   )
 
   def create[F[_]](implicit F: ApplicativeThrowable[F]): Cli[F] = new Cli[F] {
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
@@ -52,7 +52,8 @@ final case class Config(
     gitAskPass: File,
     signCommits: Boolean,
     whitelistedDirectories: List[String],
-    readOnlyDirectories: List[String]
+    readOnlyDirectories: List[String],
+    execSandbox: Boolean
 ) {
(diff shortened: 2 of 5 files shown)


ArulselvanMadhavan:
retest

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
```diff
@@ -93,7 +95,11 @@ object SbtAlg {
         fileAlg.home.map(_ / ".sbt")
 
       def exec(command: Nel[String], repoDir: File): F[List[String]] =
-        ignoreOptsFiles(repoDir)(processAlg.execSandboxed(command, repoDir))
+        if (config.execSandbox) {
+          ignoreOptsFiles(repoDir)(processAlg.execSandboxed(command, repoDir))
+        } else {
+          ignoreOptsFiles(repoDir)(processAlg.exec(command, repoDir))
+        }
```
fthomas: Can we move this conditional to `ProcessAlg`? If `config.execSandbox` is false `ProcessAlg.exexSandboxed` should just be an alias for `ProcessAlg.exec`.

modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
```diff
@@ -108,7 +108,7 @@ class UpdateService[F[_]](
               ignUpdate =>
                 update.groupId == ignUpdate.groupId && update.nextVersion == ignUpdate.nextVersion &&
                   ignUpdate.artifactIds.exists(id => update.artifactId.startsWith(id))
-            )
+          )
         )
```
fthomas: If you run `sbt scalafmt` it should reformat the code such that `sbt scalafmtCheck` is happy.

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
ArulselvanMadhavan: Made the changes. 👍

PR #211, opened 2018-12-20
use monospace formatting for artifacts

mcanlas:


fthomas:
@mcanlas Could you paste the body of an existing PR with this change applied here? It would be nice to see how this looks like when rendered.

mcanlas:
Sample output of the body.

----

Updates `ch.qos.logback`:`logback-classic` from `1.2.0` to `1.2.3`.

fthomas:
This is obviously a matter of taste but I like the formatting we're using right now more. The interspersed monospace appears a little bit uneasy on the eye IMO.

If other people think this would be a nice change, please chime in. For now I'm going to close this.

PR #242, opened 2019-01-09
add repo: albuch/sbt-dependency-check

sullis:


fthomas:
@sullis Since you are not the owner of the added repo, I'm wondering if you've checked with them if they are OK with adding it here. I'd like to avoid flooding projects with pull requests without their consent.

fthomas:
I'm going to close this. Please reopen when you have maintainer approval.

Issue #48, opened 2018-09-25
Instructions on how to run manually for private project?


Daenyth:
I'd like to use this on my company's codebase to help save maintenance time. Is that possible?

fthomas:
There is now a `Config` class that contains all the things one wants to change when running Scala steward.

Philippus:
I tried this out. By default a fork is made of the project under the specified username, for our setup cloning and then raising the PR directly would be preferred.
I also got a `422 Unprocessable Entity` after the `Create PR`-step, not sure what could be the cause of that.

fthomas:
Note that #148 changed how the GiHub password / access token is passed to Git. The configuration of a token file has been replaced by an askpass program that Git invokes to retrieve the password.

fthomas:
#161 allows to pass configuration parameters as command-line options. So scala-steward does not need to be recompiled to change the configuration.

nafg:
What about just running it on a local project on my laptop? Specifically the logic that modifies the .sbt files.

fthomas:
scala-steward is not designed to run on single projects. I think the best option here would be to create some kind of sbt-steward plugin that shares the logic for modifying source files with scala-steward. If you'd like to work on this, I could create appropriate repos at https://github.com/scala-steward-org

nafg:
How coupled is the code right now? If the code is refactored so that part
is easy to move out, I could see myself doing it one day...

fthomas:
I'd say it is already easy to move that part out. All the logic that modifies source code is currently in `Update.replaceAllIn`.

blast-hardcheese:
It seems as though
> create a file ~/.github/tokens/$myLogin with a GitHub access token for the user myLogin
is no longer required as well

blast-hardcheese:
Ah, and the last issue was my `GITHUB_HOST` variable was incorrect (left off the `/v3/api` suffix). Everything is working great now, I'll open a PR with the alterations to the `docker run` command

tjheslin1:
I believe I'm having the same issue. @blast-hardcheese looking forward to your README update!

cchantep:
Hi, first thx for this great tool. I would be interested to understand how this would work with private repos, specially how would it fork?

fthomas:
@cchantep You have two options.

1. Use a user with appropriate permissions to read and fork you repos.
[...]

PR #253, opened 2019-01-11
Add option to keep credentials for updating private dependencies

Philippus:
When using scala-steward for updating private repositories (see #48) it makes sense to also be able to update private dependencies (requiring credentials). Applying this PR will add an option to keep the credentials that are now explicitly disabled with the `set every credentials := Nil`-command.


Philippus:
Thanks! One caveat is that if a project _only_ has updates for private dependencies it will not be picked up because the project will not be present in the `allUpdates` val created in this line: https://github.com/fthomas/scala-steward/blob/master/modules/core/src/main/scala/org/scalasteward/core/steward.scala#L35

fthomas:
You're right. But these [lines](https://github.com/fthomas/scala-steward/blob/e86f1ed32a92c5d6d5750b6935c9e21b303fe43e/modules/core/src/main/scala/org/scalasteward/core/steward.scala#L34-L39) are only on optimization when running scala-steward with a lot of repos to reduce the workload. You could replace them with `repos.traverse_(ctx.nurtureAlg.nurture)` and then scala-steward will check every repo for new updates.

fthomas:
@all-contributors please add @Philippus for code, test

Issue #236, opened 2019-01-08
Use a multi-step strategy for updating versions


fthomas:
Sometimes scala-steward updates version numbers of dependencies that are similar to the current update but not part of it. They are similar because these dependencies use the same version and their artifactIds have the same prefix. One example is https://github.com/http4s/http4s/pull/2343.

We could fix such errors by trying to update version numbers in two steps. The first step would try to update only lines that contain the exact groupId, artifactId, and version. If this step changes any lines we're done. If it doesn't yield any changes we would use the current strategy for updating version numbers. In the case of https://github.com/http4s/http4s/pull/2343 this strategy would have updated the version numbers of jawn-json4s and jawn-play but not of jawn-fs2 because it has a different groupId.

PR #276, opened 2019-01-27
Extend multi-step update strategy to all source files

fthomas:
This extends the multi-step update strategy outlined in #236 to all
source files. That means we first try an update strategy on all source
files and stop if it changed any file. Only if it didn't change any file,
we'll try the next strategy and so on.

Also included is a relaxed update strategy that splits artifactIds by
hyphens and uses each part as a search term for updating the version
number. This helps in cases where there is an update for a single update
(e.g. for circe-parser) but the version is defined more generically
(e.g. `val circeVersion = "0.11.1"`).

The code is not pretty but will be improved once this idea proves to be
useful.


fthomas:
I tried this change on some repositories and it seems to doing the right thing.

PR #286, opened 2019-02-01
Add yahoo/kafka-manager

ElfoLiNk:


fthomas:
@ElfoLiNk Have you checked with the maintainers of that repository if they are OK with adding it here? I'd like to avoid flooding projects with pull requests without the consent of their maintainers.

fthomas:
I'm going to close this for now but feel free to reopen it when the maintainers of kafka-manager decide that using Scala Steward is a good idea.

PR #301, opened 2019-02-14
Ignore updates via a repo-specific .scala-steward.conf file

### LINKED ISSUE #91: Configure version ranges (e.g binary compatibility)
Are there facilities in scala-steward to configure version ranges?

For example https://github.com/lightbend/ssl-config/pull/133 updates joda-time from 2.9.9 to 2.10.  Now, leaving aside that right this moment I don't know if that's a safe update for ssl-config, assuming we can't do that upgrade how do I inform scala-steward that for the moment we want only updates in the 2.9.x version series?

(Thanks for the bot, looks like a great addition to the ecosystem!)

fthomas:
This implements repository specific configuration via a `.scala-steward.conf` file to ignore or only allow specific updates. [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) is used as format for this file. If a repository has `.scala-steward.conf` file in its root directory, scala-steward will read it and ignore updates according to the content therein.

Here is an example of a `.scala-steward.conf` file:
```hocon
updates.ignore = [
  { groupId = "com.typesafe.akka", artifactId = "akka-actor" },
  { groupId = "org.scalaz", artifactId = "scalaz-zio", version = "0.5" }
]
updates.allow = [
  { groupId = "org.http4s", version = "0.18" }
]
```
The objects of the `ignore` and `allow` lists are patterns that are matched against each update. `groupId` and `artifactId` are matched exactly while the `version` pattern matches if it is a prefix of the version of the update. If a `groupId`, `artifactId`, or `version` pattern is not specified, it is treated like a wildcard.  So the example above instructs scala-steward to
 * ignore all updates of akka-actor
 * ignore all updates of scalaz-zio where the version begins with  `0.5`
 * ignore all updates of any http4s module except those where the version starts with `0.18`

Closes #91.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,6 +30,7 @@ lazy val core = myCrossProject("core")
       Dependencies.betterFiles,
       Dependencies.caseApp,
       Dependencies.catsEffect,
+      Dependencies.circeConfig,
       Dependencies.circeGeneric,
       Dependencies.circeParser,
       Dependencies.circeRefined,
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
@@ -30,9 +30,11 @@ import org.scalasteward.core.github.http4s.Http4sGitHubApiAlg
 import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
 import org.scalasteward.core.nurture.json.JsonPullRequestRepo
 import org.scalasteward.core.nurture.{EditAlg, NurtureAlg, PullRequestRepository}
+import org.scalasteward.core.repoconfig.RepoConfigAlg
 import org.scalasteward.core.sbt.SbtAlg
 import org.scalasteward.core.update.json.JsonUpdateRepository
 import org.scalasteward.core.update.{FilterAlg, UpdateRepository, UpdateService}
+
 import scala.concurrent.ExecutionContext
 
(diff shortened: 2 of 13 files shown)


fthomas:
This works and I'm happy with the implementation but I'd also like to think about if we can implement #91 with a few tweaks to `RepoConfig` by adding some kind of allow and deny lists for dependencies that also support version ranges and wildcards for artifactIds.

fthomas:
The last commit changed `ignoreDependencies` to `ignoredUpdates` and added `allowedUpdates` and also changed the syntax of these fields. `allowedUpdates` makes it possible to ignore updates that do not match a given version pattern. Here is an example what is currently possible:
```
allowedUpdates: [
  { groupId: "org.http4s", version: "0.17" },
  { groupId: "eu.timepit", artifactId: "refined", version: "0.8" }
]
ignoredUpdates: [ { groupId: "eu.timepit", artifactId: "singleton-ops" } ]
```
With this config, scala-steward will
 * only keep updates with groupId `org.http4s` whose version starts with `0.17`
 * only keep updates of refined whose version starts with `0.8`
 * ignore all updates of singleton-ops
 * keep all other updates that do match any of those patterns

The names of the fields might change before this is merged. Bikeshedding welcome. :-)

This now implements #91.

mwz:
It would be nice to add a note about this to README.

fthomas:
@mwz I agree, documentation would be nice. But I don't want to do it in this PR. It needs to be added later.

PR #318, opened 2019-02-26
Fix docker volume mounting in README.

mwz:
By mounting scala-steward directory into `/opt/docker` inside of the container, you overwrite all of the existing files including `/opt/docker/bin/scala-steward` and all the jars in `/opt/docker/lib/` which makes the container fail on startup (as the entry point script can no longer be executed).

I've updated the README to address this by mounting `$STEWARD_DIR` into `/opt/scala-steward` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,14 +87,14 @@ Or,
 ```bash
 sbt docker:publishLocal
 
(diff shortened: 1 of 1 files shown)


fthomas:
@all-contributors please add @mwz for doc

PR #337, opened 2019-03-08
Ability to inject environment variables to firejail

mwz:
Hi @fthomas,

We've been trying to get scala-steward running here at OVO and got it working fine for our private repositories without any private dependencies. Thanks for the awesome project 😍

Our projects don't have the credentials hardcoded in the sbt config (they come from environment variables injected in a CI pipeline) and therefore we can't make use of the `keepCredentials` setting for resolving private dependencies.

The changes introduced in this PR allow us to inject any environment variables to firejail via command line arguments. This way we're able to propagate credentials to our private repositories, e.g. Bintray and run scala-steward for private projects with private dependencies 😍 

Let me know what you think about the changes.

cc: @thomaska


fthomas:
Glad to hear that scala-steward works for you! :+1:

I'm wondering if it would also work if the environment variables are passed as `extraEnv` parameter to `exec`. This would also allow to inject them if `disableSandbox` is `true`. And if this works, it would be nice to use them not only in `execSandboxed` but also in `exec`.

mwz:
Ah yes, that's a good suggestion - I'll update the PR with those changes.

I'm having some issues with a few tests failing locally, so I'll also try to get to the bottom of it and either open a separate PR or an issue regarding this.

mwz:
I've made the changes for the additional env vars to also work with `exec`, I'll resolve the remaining conflicts tomorrow.

modules/core/src/main/scala/org/scalasteward/core/io/ProcessAlg.scala
```diff
@@ -32,16 +33,20 @@ trait ProcessAlg[F[_]] {
...
+        exec(command, cwd, envVars: _*)
       } else {
         val whitelisted = (cwd.pathAsString :: config.whitelistedDirectories)
           .map(dir => s"--whitelist=$dir")
         val readOnly = config.readOnlyDirectories
           .map(dir => s"--read-only=$dir")
-        exec(Nel("firejail", whitelisted ++ readOnly) ::: command, cwd)
+        val envVars = config.envVars
+          .map(envVar => s"--env=${envVar.name}=${envVar.value}")
+        exec(Nel("firejail", whitelisted ++ readOnly ++ envVars) ::: command, cwd)
```
fthomas: @mwz Have you tried passing the env vars as third parameter here instead of usig firejail's `--env`?
```scala
val envVars = config.envVars.map(EnvVar.unapply(_).get)
exec(Nel("firejail", whitelisted ++ readOnly) ::: command, cwd, envVars: _*)
```
If this works, this feature could be independent of firejail.

mwz: This should work as firejail will inherit any environment variables from the parent process.

I originally thought that when you run firejail it sandboxes a new shell session with any env variables exported in the current shell being discarded (that's why I used the `--env` flag) but apparently, it's not the case. I gave it some thought and in my opinion, the sandboxed execution model could be a bit more strict in such a way, that firejail should not inherit any environment variables from its parent process. I'm not sure how you run scala-steward yourself, but I have the github token (for my scala-steward user) exported as an env variable in the CI and it would make sense to me if it wasn't exposed in the sandbox.

I think this could be achieved by wrapping firejail it in the following command which clears the environment: `env -i bash -lc 'firejail ...'` with the additional environment variables injected explicitly to firejail via the use of `--env`.

I'm happy, for the purpose of this PR, to make the change you suggested, but I would be keen to discuss this a bit further and open another PR with the changes, if you agree with my idea.

fthomas: That's a good point I haven't thought about and I agree that `execSandboxed` should not inherit the environment from scala-steward.

So my current understanding is that environment variables injected via `--env-var` to scala-steward
 - are the only environment variables available to `execSandboxed`
 - and are added to the environment when calling `exec`.

Does that sound good to you?

It is your call if you want to do the changes in this PR or in a follow-up PR. I'm happy either way. :-)

mwz: > So my current understanding is that environment variables injected via `--env-var` to scala-steward
> 
> * are the only environment variables available to `execSandboxed`
> * and are added to the environment when calling `exec`.
> 
> Does that sound good to you?

Yes, that sounds good 👍 I've made the changes you suggested before, so this PR should be done. I'll follow up with another PR later to implement the rest.

fthomas:
@all-contributors please add @mwz for code, test, platform
@all-contributors please add @thomaska for code, test

fthomas:
@all-contributors please add @mwz for code, test, platform

fthomas:
@all-contributors please add @thomaska for code, test

fthomas:
@all-contributors please add @thomaska for code, test

PR #354, opened 2019-03-20
Decouple authorization from Http4sGitHubApiAlg

fthomas:


fthomas:
The main purpose of this change is to use `Http4sGitHubApiAlg` with other kinds of authorization.

PR #361, opened 2019-03-24
Add algebra for cloning and syncing GitHub repos

fthomas:


fthomas:
This change has two purposes:
- make `GitAlg` independent of anything related to GitHub (`checkAndSyncFork` used a data type from the GH API)
- have one place where `RepoOut.clone_url`s are modified before using them to clone repos

PR #367, opened 2019-03-26
attempt to include sbt-catalysts

kailuowang:
I am not sure how scala-steward works and if it will work for sbt-catalyts. 
The versions I cared most are stored [here](https://github.com/typelevel/sbt-catalysts/blob/master/src/main/scala/org/typelevel/TypelevelDeps.scala), not even in the `build.sbt` file or in `project`
It's fine if steward doesn't know how to work with these versions at least I imagine it will update the plugin's own deps.

### DIFF
--- a/repos.md
+++ b/repos.md
@@ -311,7 +311,6 @@
 - travisbrown/iteratee-twitter
 - twilio/guardrail
 - typelevel/cats
-- typelevel/sbt-catalysts
 - typelevel/cats-collections
 - typelevel/cats-effect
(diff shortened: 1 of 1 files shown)


fthomas:
scala-stewards uses sbt-updates' `dependencyUpdates` task under the hood, so you can test which dependencies it will update. If the dependencies in TypelevelDeps.scala are not part of the build (i.e. part of `libraryDependencies`), scala-steward won't propose updates for them.

Issue #365, opened 2019-03-26
.sbtopts temporary removal suppresses sbt.repositories


Slakah:
A common pattern in sbt is to override the repositories, by specifying the following in the `.sbtopts` of a project.

i.e.
`.sbtopts`:
```
-Dsbt.override.build.repos=true
-Dsbt.repository.config=sbt.repositories
```
The following code temporarily removes the `.sbtopts` file.
https://github.com/fthomas/scala-steward/blob/bf3191ed44ce9e39de7195b64b0bc2ff09c8e6ac/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala#L107

I can't think of a sensible solution.

Maybe we could add a whitelist for certain params?

I suppose a workaround, would be to define resolvers inside the build sbt definitions.

Thanks for the great tool!

fthomas:
I'm removing `.sbtopts` and `.jvmopts` because some projects use them to set `-Xms` values which are too high for my server. So these files are only removed because my server has not enough memory.

One easy way to keep these files when running scala-steward yourself would be to introduce a new config option which would change the behaviour of `ignoreOptsFiles`.

Slakah:
I've had a quick think about it, would you be open to a PR which would do the following?
* Introduce `FileAlg#editFileTemporarily`.
* Produce a whitelist of SBT_OPT flags, probably `-Dsbt.override.build.repos` and `Dsbt.repository.config`.
* Change `ignoreOptsFiles`, honour the SBT_OPT whitelist.

Another option might be to instead of introducing `editFileTemporarily`:
* Read the `.sbtopts` before it's temporarily removed.
* Extract whitelisted SBT_OPT flags.
* Supply whitelisted SBT_OPT flags to `SbtAlg#sbtCmd`.

I can't promise if/when I might find time to raise such a PR.

Again thanks for the help!

fthomas:
I would prefer to not parse `.sbtopts` files if not absolutely necessary. A new config option that disables `ignoreOptsFiles` would be much simpler. But maybe I don't understand your motivation for this change. Do you need this for a company internal instance of scala-steward you are running yourself or do you want that my public instance honors some flags defined in `.sbtopts`?

Slakah:
> I would prefer to not parse .sbtopts files if not absolutely necessary. A new config option that disables ignoreOptsFiles would be much simpler.

That's fair, I'll try and find time to raise a PR.

[...]

PR #380, opened 2019-04-02
#365 Add --ignore-opts-files option

Slakah:
Add a flag `--ignore-opts-files`. This flag will enable the temporary deletion of .sbtopts and .jvmopts files (previously this PR was to have a flag `--disable-ignore-opts-files` to do the opposite).

See: #365 

Please let me know if there's anything amiss or any tests that I should probably add. Thanks!

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -42,6 +42,7 @@ object Cli {
       readOnly: List[String] = Nil,
       disableSandbox: Boolean = false,
       doNotFork: Boolean = false,
+      ignoreOptsFiles: Boolean = false,
       keepCredentials: Boolean = false,
       envVar: List[EnvVar] = Nil
   )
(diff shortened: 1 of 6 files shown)


modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
```diff
@@ -42,6 +42,7 @@ object Cli {
       readOnly: List[String] = Nil,
       disableSandbox: Boolean = false,
       doNotFork: Boolean = false,
+      disableIgnoreOptsFiles: Boolean = false,
```
mwz: Perhaps `respectOptsFiles` would be a better name for this flag?

fthomas:
LGTM, thanks @Slakah! I'd be also okay with `--ignore-opts-files` and changing the default behavior when this flag is not present. I agree that removing those files is a special case and warrants an additional flag and the default should be not touching them at all.

Slakah:
> I'd be also okay with --ignore-opts-files and changing the default behavior when this flag is not present.

Done!

I've also added some tests, to check my changes actually work!

fthomas:
@all-contributors please add @Slakah for code, test

PR #409, opened 2019-04-16
Update repos.md

olivierschultz:


fthomas:
Be aware that scala-steward will not update libraries and plugins of the template but only those of the project.

PR #421, opened 2019-04-23
escape multi line bash documentation

mcanlas:
Makes copy pasting the examples less error-prone.


fthomas:
@all-contributors please add @mcanlas for doc

PR #431, opened 2019-04-28
Extend ignore comments tests and fix a tiny bug

kiranbayram:
Hey there,

after adding some different test lines for ignore-comments test, i noticed i should have trimmed before checking if it starts with `//`. But nonetheless, here is the PR for it. Sorry for the inconvenience :sweat_smile: 

Cheers


fthomas:
Thanks @kiranbayram!

Issue #377, opened 2019-04-01
Document how to structure dependencies to optimally work with scala-steward


yannick-cw:
I am wondering if there are things which prevent scala-steward from working properly and updating dependencies. Is there some documentation how to structure or NOT to structure the dependencies?

fthomas:
Currently there is no documentation about that. But you're right, there are ways which makes it easy for scala-steward to bump version numbers and ways which make it outright impossible to bump them. For example, the line where the version is defined needs to contain parts of the artifactId in front of the version. So if you'd define the version like this
```scala
val catsVersion =
  "1.6.0"
```
scala-steward would not recognize that `1.6.0` is Cats' version number.

kiranbayram:
This issue should be closed i guess. @yannick-cw you can check the examples at [ExampleTest](https://github.com/fthomas/scala-steward/blob/master/modules/core/src/test/scala/org/scalasteward/core/model/ExampleTest.scala)

yannick-cw:
@kiranbayram yes looks great :)

PR #456, opened 2019-05-01
#377 Add good and bad examples doc for dependency definitions

kiranbayram:
Hi there,

I've decided to put examples to a separate file to keep README file clean. And also this doesn't cover all possible working examples (since there're countless combinations) but i've put a link to UpdateTest if somebody wanted to more possibilities or edge cases which are handled.

Cheers


fthomas:
Many thanks for this, @kiranbayram! This is much appreciated. What do you think about having this documentation as actual test so that the good and bad examples are always in sync with the code? I fear that if this is not machine checked it will be outdated rather quickly.

kiranbayram:
Yes, you're probably right. I'll try to write an ExampleTest which covers these cases and we can refer to that in Readme

kiranbayram:
It would still be nice to create documentation automatically from this test's output but i don't know how .

fthomas:
We could use mdoc (https://scalameta.org/mdoc/) to have these examples as typechecked documentation.

kiranbayram:
Should we then probably open a feature request for trying that out?

fthomas:
Sounds good to me.

PR #470, opened 2019-05-06
Add Zalando to the list of companies

sergeykolbasov:
We have dozens of teams & hundreds of people using Scala in their daily work at Zalando, so it was inevitable to set up scala-steward for internal purposes.

As a side note, it was incredibly easy to configure it, so huge thanks!


fthomas:
You're welcome, and thanks for letting me know!

Issue #404, opened 2019-04-11
Add labels to pull requests


Fristi:
It would be nice to classify a pull request by adding

- Library update
- Sbt plugin update
- Test library update

Such that you could do automatic patch releases by using mergify (auto-merge) + rancher (auto-tag) + sbt-release-early. I would only release a new artifact if there is a Library update

fthomas:
I like this idea. Customisation of the labels (or disabling/enabling this feature) could be configured via `.scala-steward.conf`.

kiranbayram:
Hello there, are we talking about Github labels here? Push access seems to be required to add labels to issues (PRs are also issues in Github api context). That could be a problem.

fthomas:
Yes, I assumed GitHub labels. An alternative could be to just add "labels" to the PR body as plain text like this: `labels: library-update`. Other tools might be able to work with such labels too.

kiranbayram:
Putting labels into PR body as plain text should work, yes. I may give this a go in coming days if i get to have some free time.

MPV:
I added #1471 as follow-up issue for adding GitHub labels, in case there's others interested in that feature.

PR #475, opened 2019-05-08
Labels feature

kiranbayram:
Opening this PR to make it transparent that i'm working on this feature

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/model/Label.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/model/Label.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2018-2019 scala-steward contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.scalasteward.core.model
 
 sealed trait Label {
--- a/modules/core/src/main/scala/org/scalasteward/core/model/Label.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/model/Label.scala
@@ -15,6 +15,7 @@
  */
 
 package org.scalasteward.core.model
+import io.circe.{Decoder, Encoder}
 
 sealed trait Label {
   def name: String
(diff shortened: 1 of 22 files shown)


kiranbayram:
Hello @fthomas , I wanted to ask you before continuing if I should add a `labels` field to `Update` model or should I keep the labels outside and write a `LabelAlg` which returns labels for an `Update` ? What do you think ?

fthomas:
@kiranbayram How would a potential `LabelAlg` determine a label for an `Update`? I think at the time scala-steward found `Update`s for a project it has forgotten if it is an update for a library dependency, a test dependency, or for a a plugin and that information can't be recreated easily. So I would think that scala-steward need to somehow retain that info when it determines `Update`s for a project.

kiranbayram:
Then it sounds like we need to add labels field to Update model and use it while creating PR body

modules/core/src/main/scala/org/scalasteward/core/model/Label.scala
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.scalasteward.core.model
+import io.circe.{Decoder, Encoder}
+
+sealed trait Label {
```
Fristi: You could make this a `sealed abstract class Label(val name: String)` which save you some lines below

kiranbayram:
@Fristi Thanks for the review. I think, it might be useful to automatically identify a dependency type but maybe with another PR.

kiranbayram:
Hello @fthomas, could you please review ?

fthomas:
Do I understand it correctly that this PR basically allows users to define a mapping from dependencies to labels in `.scala-steward.conf` which scala-steward then uses to label PRs of matching dependencies?

kiranbayram:
Yes, that is right

kiranbayram:
I thought this was what you suggested in your comment on the issue

fthomas:
Sorry for the confusion, @kiranbayram. My understanding of #404 was this: scala-steward knows when it gets updates from sbt-updates if a dependency is a normal or test dependency (this is stored in `Update.Single.configurations`) and it could know which updates are for build dependencies and plugins. With this info we could label them automatically without the need for any configuration. With customisation of the labels I meant to configure the names of those labels via the config file.

kiranbayram:
Hmm, sorry from my side too. I should probably have communicated what i wanted to do before doing it 😊 I can try to adjust the PR once i get some free time

kiranbayram:
I think it'd be easier to implement this feature from scratch than updating this branch, that's why i'm going to close this PR.

MPV:
Did anyone (@kiranbayram) give this another try? ❤️

MPV:
Or was https://github.com/fthomas/scala-steward/pull/755 the merged solution for this? ❔👍

exoego:
If you mean #404 for "this", yes, I think the merged PR #755 was for it and superceded this PR.

MPV:
Ah yes, thanks @exoego ❤️ 

I opened #1471 to discuss also adding proper labels (not just into PR description).

Issue #313, opened 2019-02-22
GitLab support


daddykotex:
We'd like to leverage this awesome project to update the dependencies on our internal project dependencies. To do that, scala-steward must support GitLab. I've started to work on a native implementation.

I'm opening this MR just to leave a trace of that on the internet. That way, if someone else planned/was planning or is already doing it, we can coordinate better.

Thanks again for the project, it is really a breeze to work it.

fthomas:
:+1: 

@tjheslin1 is working on BitBucket support. I expect that there is quite a bit of overlap in your efforts.

daddykotex:
In fact, the changes in https://github.com/fthomas/scala-steward/compare/master...tjheslin1:bitbucket look very similar to mine. The naming is different but that's it. I think `vcs`, as @tjheslin1 use is better than what I had, so I'll rebase on his work and see where it leads me.

tjheslin1:
Hi @daddykotex. Glad my branch is able to help you. I look forward to seeing a PR!

I have gotten a bit distracted in the last week or two regarding this contribution, I'm glad you're working on it too! When trying to test locally I ran into the issue documented in #48. I was waiting for this to be closed as I saw something was in progress on a fork for this.

For BitBucket the reading up on the BitBucket api is done, unless there's a chance that changes to authentication needs to be made, which I was trying to run scala-steward locally for.

daddykotex:
I'm on my way to finish the Gitlab integration. I will test it a little bit here, in-house, and if everything operates smoothly, I'll open multiple PR one by one and try to split my changes to ease review.

kubukoz:
This would be totally awesome, and we'd benefit greatly from having scala-steward work in self-hosted GL. I want to help, so @daddykotex feel free to let me know if you need some help (even if only in the review).

daddykotex:
So there you go, with a little bit of pressure :)

https://github.com/daddykotex/scala-steward/tree/refactor/vcsapialg

This is a WIP for Gitlab support. I tried it on an internal repository but also on Gitlab public services. I don't think my tests covered the whole API so I suggest a few Gitlab users take an hour to try it on their own Gitlab instances (or their public repositories) to make sure it works fine.

Here is how to proceed:

1. Create a user on your Gitlab installation.
2. In this user's settings add a new Access Token with permissions to access the API.
3. Write a bash program that print the password, like that:

```bash
#!/bin/bash
echo "ACCESS_KEY"
```

4. Check out the code shared above
[...]

PR #497, opened 2019-05-18
Refactor to move github/data to vcs/data

daddykotex:
This is mostly changes to prevent the PR with the actual logic to be obfuscated. I move the data structures from github/data to vcs/data. I also rename the `GitHubRepoAlg` to `VCSRepoAlg`.

With this PR, we are assuming the data structures (used to talk to GitHub) will work for any VCS be it BitBucket and GitLab. In reality, it's a bit of a stretch. I've made it work with a bunch of shortcut (as a prototype it ran on one of our internal repository and it opened a bunch of PR 👍 ). But we'll need to discuss what needs to change in those structures in order to accommodate other VCS when I open the PR with the actual logic.

The actual changes would look like:

* a new package called org.scalasteward.core.gitlab (with a different `Url` class 
* custom encoder/decoders to go from Gitlab's JSON to the vcs data structures (if they're left as-is)
* a VCSApiAlg that covers Github and Gitlab
* a VCS Selector (if you will) to read the config and build the appropriate implementation of `VCSApiAlg`

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
@@ -21,7 +21,7 @@ import cats.effect.Sync
 import org.http4s.Uri
 import org.scalasteward.core.application.Cli.EnvVar
 import org.scalasteward.core.git.Author
-import org.scalasteward.core.github.data.AuthenticatedUser
+import org.scalasteward.core.vcs.data.AuthenticatedUser
 import org.scalasteward.core.util
 import scala.sys.process.Process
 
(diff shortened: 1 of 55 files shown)


daddykotex:
Referencing #313

fthomas:
Ok, let's do this! I agree that reusing the same data structures for different repository mangers might be problematic. Especially the `Out` classes currently mirror GitHub's API and if we want to support different features for different repository mangers in the future we probably also need different `Out` classes.

Btw, thanks @daddykotex for splitting your changes into this "rename" PR and a later "logic" PR. This makes it easier for me to review them.

daddykotex:
Wow amazing. Thanks to you for scala-steward.

If you want to make sure you are not breaking anything, we can work on a feature branch. I'll open the follow-up MR and explains the difference so we can all be on the same page about the differences between GitHub integration and Gitlab integration.

fthomas:
:+1: Sounds good to me. I've created the branch [topic/gitlab](https://github.com/fthomas/scala-steward/tree/topic/gitlab) for that.

PR #577, opened 2019-06-16
Add migration for http4s 0.20

fthomas:


fthomas:
Here is the referenced rule: https://github.com/http4s/http4s/blob/master/scalafix/rules/src/main/scala/fix/v0_20.scala

fthomas:
Here is this PR in action: https://github.com/fthomas/scalafix-test/pull/8/files

fthomas:
What is currently unclear to me is how to determine the list of artifactIds from looking at the rewrite rules. The easiest for me would be to run this migration when any artifactId of the update matches `http4s-.*` @amarrella do you have any insights which artifactIds we should list here?

amarrella:
I *think* you can safely target all artifacts, but the fix will probably mostly used for clients and server?

fthomas:
Thanks for the confirmation, @amarrella. The migration targets now all `http4s-.*` artifacts because I found it too tedious to list all applicable artifacts individually.

PR #596, opened 2019-06-19
Add Gemini Observatory repos to scala-steward

cquiroz:
Add Gemini Observatory repos to be tracked by scala-steward

/cc @tpolecat @swalker2m


cquiroz:
Thanks

Issue #603, opened 2019-06-21
Document the scalafix integration


fthomas:
Maybe a blog post announcing this would encourage people to write more Scalafix integrations.

fthomas:
Fixed by #609. I'm writing a blog post later.

PR #609, opened 2019-06-21
Documentation for the Scalafix integration

fthomas:
Taking a first stab at writing documentation for the Scalafix
integration.

Related to #603.


fthomas:
Reviews of this PR would be much appreciated.

Issue #622, opened 2019-06-24
java.io.IOException without any stacktrace when running scala-steward


anilkumarmyla:
We run scala-steward in our private repos, and it was working fine until recently when it started spewing IOException without any stack trace. The last successful run was on `2019-06-17 12:33:53` and next run on `2019-06-20 01:03:00` failed. We checkout latest code from master and run it, so there are a wide range of commits between the two timestamps to look for the commit that causes this. Error message from the run is as below, notice there is no stack trace for IOException making it hard to debug 

```
12:49:23 failed finding updates for ArtificialProject(ScalaVersion(2.12.8),SbtVersion(1.2.8),List(Dependency(ch.qos.logback,logback-classic,logback-classic,1.2.3,None,Some(test))),List())
12:49:23 java.io.IOException: [info] Loading settings for project global-plugins from sbt-updates.sbt,scala-steward.sbt ...
12:49:23 [info] Loading global plugins from /root/.sbt/1.0/plugins
12:49:23 [info] Loading settings for project updates-build from plugins.sbt ...
12:49:23 [info] Loading project definition from ...
12:49:23 [info] Loading settings for project updates from build.sbt ...
12:49:23 [info] Set current project to updates (in build ...)
12:49:23 [error] Not a valid command: dependencyUpdates
12:49:23 [error] Not a valid project ID: dependencyUpdates
12:49:23 [error] Expected ':'
12:49:23 [error] Not a valid key: dependencyUpdates (similar: dependencyClasspath, dependencyOverrides, dependencyPositions)
12:49:23 [error] dependencyUpdates
12:49:23 [error]                  ^ 
```

When we run scala-steward locally it works fine, but in our ci systems above error happens. Any idea what causes this and how to debug this?

fthomas:
`Not a valid command: dependencyUpdates` hints at a problem with the sbt-updates plugin. Could you look into the global sbt plugins directory (`~/.sbt/1.0/plugins/` and  `~/.sbt/0.13/plugins/`) and remove all `.sbt` files from Scala Steward: `sbt-updates.sbt`, `scala-steward.sbt`, and `StewardPlugin.scala`

anilkumarmyla:
Ok, we cache `~/.sbt` in our ci build. Clearing it and running again worked, probably some stale data in `~/.sbt` caused this. We can close this

PR #623, opened 2019-06-24
Use logger.error instead of println

### LINKED ISSUE #420: Usage of println instead of logger
There are around 6 usages of println in the codebase instead of using logger. Just wondering if that's intended or we can switch to logger. Willing to help, let me know!

anilkumarmyla:
Kinda fixes #420 
and helps debugging #622

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
@@ -84,8 +84,7 @@ final class UpdateService[F[_]](
             util.divideOnError(prj)(sbtAlg.getUpdatesForProject)(_.halve.toList.flatMap {
               case (p1, p2) => List(p1, p2)
             }) { (failedP: ArtificialProject, t: Throwable) =>
-              logger.error(t)(s"failed finding updates for $failedP error=${t.getMessage}")
-              F.pure(List.empty[Update.Single])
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
```diff
@@ -84,8 +84,7 @@ final class UpdateService[F[_]](
             util.divideOnError(prj)(sbtAlg.getUpdatesForProject)(_.halve.toList.flatMap {
               case (p1, p2) => List(p1, p2)
             }) { (failedP: ArtificialProject, t: Throwable) =>
-              println(s"failed finding updates for $failedP")
-              println(t)
+              logger.error(t)(s"failed finding updates for $failedP error=${t.getMessage}")
```
fthomas: `logger.error(...)` is a pure expression and needs to be composed via `>>` with the following expression to have an effect.

ChristopherDavenport: `logger.error(...).as(List.empty[Update.Single])` is even simpler since the result is just a pure.

fthomas: :+1: This is a nice application of the monad law: `flatMap(_ => pure(x)) <-> map(_ => x)`.

PR #645, opened 2019-07-01
Merge topic/gitlab into master

fthomas:
I'm opening this so that people can see the diff between topic/gitlab and master and to track pending tasks before this can be merged.


modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -84,8 +84,8 @@ final class NurtureAlg[F[_]](
   def processUpdate(data: UpdateData): F[Unit] =
     for {
       _ <- logger.info(s"Process update ${data.update.show}")
-      head = vcs.headFor(data.fork.show, data.update)
-      pullRequests <- gitHubApiAlg.listPullRequests(data.repo, head, data.baseBranch)
+      head = vcsSpecifics.sourceFor(data.repo, data.update)
```
fthomas: This does not do the right thing for GitHub repos since `head` is not created from `data.fork` (see #552 why this is needed). I'll try to change this later.

fthomas:
One thing I'd also like to change is the `vcs` acronym which is usually used as generic term for Git, Mercurial, Subversion, CVS, etc. but we are using it as generic term for GitHub, GitLab, and BitBucket. I'm not aware of a common generic term for these services but would propose to use `grm` (for Git repository manager). Other suggestions welcome!

Since this is a purely stylistic change, it should be done after merging this so we don't obfuscate the diff of this PR with a simple rename.

fthomas:
@daddykotex I've just verified that this branch still works with GitHub. So there is nothing left to do from my side before this can be merged.

daddykotex:
Sounds good to me!

All aboard the VCS diversity!

PR #649, opened 2019-07-02
Fix branch name

daddykotex:
During merge/rebase, I may have accidentally changed the
branch name used when creating the MR. On Gitlab, we need
`update/dependency-1.2.0` where as on GitHub we need:
`owner:update/dependency-1.2.0`.


modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -123,7 +123,8 @@ final class NurtureAlg[F[_]](
   def createPullRequest(data: UpdateData): F[Unit] =
     for {
       _ <- logger.info(s"Create PR ${data.updateBranch.name}")
-      requestData = NewPullRequestData.from(data, config.vcsLogin)
+      branchName = vcsSpecifics.headForListingPullRequests(data.fork, data.update)
+      requestData = NewPullRequestData.from(data, branchName, config.vcsLogin)
```
fthomas: This will use `owner/repo:update/dependency-1.2.0` as `branchName` in case of GitHub but it should be `owner:update/dependency-1.2.0`. It is unfortunate that GitHub requires two different `head` parameters for listing pull requests and creating new ones.

daddykotex: You are right, what I wanted here is `owner:update/dependency-1.2.0`. I've fixed it with a second method in VCS Specifics.

But maybe instead of using the interface VCSSpecifics that I added, we could replace it with functions in package object (like it was before) but that takes in the `VCSType` ?

fthomas: Sounds good to me too.

daddykotex:
Removed the whole `VCSSpecifics` thing

PR #665, opened 2019-07-06
Add ability to propose sbt version update

### LINKED ISSUE #101: Propose sbt updates

exoego:
Closes #101.
Steward will propose update if they found `sbt.version=X.Y.Z-SOMETHING`.
It does not update `SbtVersion` that used for conditional build like below

```scala
sbtSeries.value match {
    case "0.13" => SbtVersion("0.13.17")
    case "1.0"  => defaultSbtVersion
    case _      => defaultSbtVersion
}
```
since such build requires cautious update.

I could not find how to perform system-test against actual repos...
How should I perform ?

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
@@ -79,34 +79,27 @@ object UpdateHeuristic {
     getPrefixRegex = update => Some(s"${update.groupId}.*?")
   )
 
-  val sbt = UpdateHeuristic(
-    name = "sbt",
-    order = 2,
-    getSearchTerms = _ => List("sbt\\.version="),
-    getPrefixRegex = _ => None
-  )
-
   val original = UpdateHeuristic(
     name = "original",
-    order = 3,
+    order = 2,
     getSearchTerms = update => update.searchTerms.toList
   )
 
   val relaxed = UpdateHeuristic(
     name = "relaxed",
-    order = 4,
+    order = 3,
     getSearchTerms = update => util.string.extractWords(update.artifactId)
   )
 
   val sliding = UpdateHeuristic(
     name = "sliding",
-    order = 5,
+    order = 4,
     getSearchTerms = update => update.artifactId.sliding(5).take(5).filterNot(_ === "scala").toList
   )
 
   val groupId = UpdateHeuristic(
     name = "groupId",
-    order = 6,
+    order = 5,
     getSearchTerms = update =>
       update.groupId
         .split('.')
@@ -117,5 +110,5 @@ object UpdateHeuristic {
   )
 
   val all: Nel[UpdateHeuristic] =
-    Nel.of(strict, original, relaxed, sliding, groupId, sbt).sortBy(_.order)
+    Nel.of(strict, original, relaxed, sliding, groupId).sortBy(_.order)
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
@@ -79,27 +79,34 @@ object UpdateHeuristic {
     getPrefixRegex = update => Some(s"${update.groupId}.*?")
   )
 
+  val sbt = UpdateHeuristic(
+    name = "sbt",
+    order = 2,
+    getSearchTerms = _ => List("sbt\\.version="),
+    getPrefixRegex = _ => None
+  )
+
   val original = UpdateHeuristic(
     name = "original",
-    order = 2,
+    order = 3,
     getSearchTerms = update => update.searchTerms.toList
   )
 
   val relaxed = UpdateHeuristic(
     name = "relaxed",
-    order = 3,
+    order = 4,
     getSearchTerms = update => util.string.extractWords(update.artifactId)
   )
 
   val sliding = UpdateHeuristic(
     name = "sliding",
-    order = 4,
+    order = 5,
     getSearchTerms = update => update.artifactId.sliding(5).take(5).filterNot(_ === "scala").toList
   )
 
   val groupId = UpdateHeuristic(
     name = "groupId",
-    order = 5,
+    order = 6,
     getSearchTerms = update =>
       update.groupId
         .split('.')
@@ -110,5 +117,5 @@ object UpdateHeuristic {
   )
 
   val all: Nel[UpdateHeuristic] =
-    Nel.of(strict, original, relaxed, sliding, groupId).sortBy(_.order)
+    Nel.of(strict, original, relaxed, sliding, groupId, sbt).sortBy(_.order)
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
@@ -25,6 +25,7 @@ import org.scalasteward.core.dependency.Dependency
 import org.scalasteward.core.vcs.data.Repo
 import org.scalasteward.core.io.{FileAlg, FileData, ProcessAlg, WorkspaceAlg}
 import org.scalasteward.core.model.Update
+import org.scalasteward.core.sbt
 import org.scalasteward.core.sbt.command._
 import org.scalasteward.core.sbt.data.ArtificialProject
 import org.scalasteward.core.scalafix.Migration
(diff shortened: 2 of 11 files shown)


modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
```diff
@@ -79,27 +79,34 @@ object UpdateHeuristic {
     getPrefixRegex = update => Some(s"${update.groupId}.*?")
   )
 
+  val sbt = UpdateHeuristic(
+    name = "sbt",
+    order = 2,
```
exoego: I initially tried to set `order = 6` (largest), but could not.
Since the heuristic `original` has higher precedence than `sbt`.

exoego:
Ah, I did not aware about `SbtVersion("0.13.18")`.
Will add soon.

fthomas:
Thanks @exoego for working on this. I think a few more things are needed for this feature:

1. [`isSourceFile`](https://github.com/fthomas/scala-steward/blob/93eb7a62946c1f7ca20c28a41ac44e6341afa353/modules/core/src/main/scala/org/scalasteward/core/io/package.scala#L23-L28) needs to be updated to return `true` for `project/build.properties` files.
1. `SbtAlg.getUpdatesForRepo` needs to include `Update.Single` for sbt like you've used in the test (`Update.Single("org.scala-sbt", "sbt", "1.2.7", Nel.of("1.2.8")`). sbt-updates doesn't include updates for sbt, so we have to implement this ourself. In the first iteration of this feature I would propose to use `sbt.defaultSbtVersion` as latest for the 1.x series and add a new `sbt.latestSbtVersion_0_13` for the 0.13.x series. The sbt version of a `Repo` can then be queried via sbt or by reading `project/build.properties` (which I'd prefer because it is faster.) With this information you can now create an `Update.Single` and include it in `getUpdatesForRepo`.

I'm not sure if we then need a new `UpdateHeuristic`. Maybe the heuristics we already have are good enough. And to test it I would run Scala Steward with a test repo.

exoego:
Updated `SbtAlg.getUpdatesForRepo` and reverted `UpdateHeuristic` in https://github.com/fthomas/scala-steward/pull/665/commits/7676425c2a92523b57a74a775ffe97c93bfb77e2

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
```diff
@@ -110,8 +111,10 @@ object SbtAlg {
           maybeClearCredentials = if (config.keepCredentials) Nil else List(setCredentialsToNil)
           commands = maybeClearCredentials ++
             List(dependencyUpdates, reloadPlugins, dependencyUpdates)
-          lines <- exec(sbtCmd(commands), repoDir)
-        } yield parser.parseSingleUpdates(lines)
+          sourceUpdates = exec(sbtCmd(commands), repoDir).map(parser.parseSingleUpdates)
+          buildPropUpdates = proposeBuildPropertiesUpdate()
+          updates <- (sourceUpdates product buildPropUpdates).map(p => p._1 ::: p._2)
```
exoego: I do not know how to concate two `F[List[A]]` in `cats` so this concatenation may be funny.

fthomas: What about
```scala
  sourceUpdates <- ...
  buildPropUpdates <- ...
} yield sourceUpdates ++ buildPropUpdates
```
?

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
```diff
@@ -125,6 +128,31 @@ object SbtAlg {
...
+      def proposeBuildPropertiesUpdate(): F[List[Update.Single]] =
+        buildPropertiesFile.flatMap { prop =>
+          fileAlg
+            .readFile(prop)
+            .map { maybeContent =>
+              for {
+                content <- maybeContent
+                currentVer <- sbtVersionRegex.findFirstMatchIn(content).map(_.group(1))
+                if !latestSbtVersions.contains(currentVer)
+                newVer = if (currentVer.startsWith("0."))
```
fthomas: We need to ensure that `newVersion` is actually greater than `currentVer`. `model.Version` can be used to do that.

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
```diff
@@ -125,6 +128,31 @@ object SbtAlg {
...
         fileAlg.home.map(_ / ".sbt")
 
+      val buildPropertiesFile: F[File] =
+        fileAlg.home.map(_ / "project" / "build.properties")
+
+      final private val sbtVersionRegex = s"build.properties=(.+)".r
+      final private val latestSbtVersions = List(sbt.defaultSbtVersion, sbt.latestSbtVersion_0_13)
+        .map(_.value)
+
+      def proposeBuildPropertiesUpdate(): F[List[Update.Single]] =
```
fthomas: I think this should return `F[Option[Update.Single]]` to make clear that it returns at most one element. At the usage site it can then be converted to a list for easier concatenation with the other updates.

fthomas: I've another suggestion. If you extract the part of this function that transforms the content of `build.properties` to an update (`String => Option[Update.Single]`), that function has no effect in `F[_]` and can be tested more easily.

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
exoego: Done in https://github.com/fthomas/scala-steward/pull/665/commits/200b7c6de1f389af6287ba90468aa1eb304cda1f

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
exoego: Mad it `Option` in https://github.com/fthomas/scala-steward/pull/665/commits/2a407f257f2629154d9cae4f490af68a81cd437c
Extracted the method in https://github.com/fthomas/scala-steward/pull/665/commits/a3516c8edebf20a9e8821586905af4287bf9e7fe

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
exoego: Done in https://github.com/fthomas/scala-steward/pull/665/commits/1578a3dbe3128cdae517c687a5b85aa5c6e2b3cd

fthomas:
I just tested your branch with one of my repos and it works: https://github.com/fthomas/properly/pull/26
This is great!

I also pushed two minor fixes.

fthomas:
I tried this on a few more repos:
* https://github.com/ChristopherDavenport/agitation/pull/29/files
* https://github.com/ChristopherDavenport/base.g8/pull/11/files
* https://github.com/ChristopherDavenport/scalaz-task-effect/pull/73/files

And they all look good.

The only one that also updated some unrelated version number was this:
* https://github.com/daandi/nice-url-encode-decode/pull/8/files

I'm not sure yet if we should restrict sbt updates to `build.properties` files.

exoego:
It looks like `UpdateHeuristic.original` is used for sbt update.
Since `UpdateHeuristic.strict` requires `groupId` in the line, but `build.properties` dont' have `groupId`,
so `UpdateHeuristicoriginal` with a Regex like `(?i)(.*)((sbt).*?)\Q1.2.8\E(.?)` is used.
Such regex can matches `addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.2.8")` too.

> I'm not sure yet if we should restrict sbt updates to build.properties files.

I think restriction is needed.
Where is a good place for such file restriction ?

fthomas:
The file restriction would be computed in `EditAlg` and then passed to `FileAlg.findSourceFilesContaining` instead of using the static `isSourceFile`.

This would restrict sbt updates to only edit `project/build.properties` files and nothing else. I'm trying to think of situations where the sbt version might be used somewhere else which we would then miss.

exoego:
> The file restriction would be computed in EditAlg and then passed to FileAlg.findSourceFilesContaining instead of using the static isSourceFile.

Added such dynamic `Update`-based file filter in https://github.com/fthomas/scala-steward/pull/665/commits/94c1e18d45e430bcb56aac8d37d556f886157d90

> I'm trying to think of situations where the sbt version might be used somewhere else which we would then miss.

To use sbt version somewhere in `build.sbt`, I think users should use `sbtVersion` instead of literal.
So I assume it is not a big problem if we restrict sbt update only to `build.properties` in 1st iteration.

kubukoz:
https://github.com/kubukoz/sup/pull/84 it works, thanks for this update!

PR #671, opened 2019-07-07
Add markup to ignore lines

### LINKED ISSUE #192: Support markup to ignore single or multiple lines
see https://github.com/erikvanoosten/metrics-scala/pull/118

exoego:
Closes #192

Adding two markups to control lines to be ignored.

# Ignoring single-line
A line ends with `// scala-steward:off` is ignored solely.
```scala
"com.typesafe.akka" %% "akka-actor" % "2.4.0", // scala-steward:off
"com.typesafe.akka" %% "akka-testkit" % "2.4.0", // Update should happen on this line
```

# Ignoring multiple-lines
Once a line starts with `// scala-steward:off` found, Steward won't edit lines until `scala-steward:on` found.

```scala
// scala-steward:off
"com.typesafe.akka" %% "akka-actor" % "2.4.0",
"com.typesafe.akka" %% "akka-testkit" % "2.4.0",
// scala-steward:on
"com.typesafe.akka" %% "akka-stream" % "2.4.0", // Update should happen on this line
```

# Note for staying specific range

A range markup like `// scala-steward:only 2.4.x` was suggested in #192. 
I think we do not need such, since sbt already supports [version range in the form of `2.4.+` or `[2.4,2.5)`](https://www.scala-sbt.org/release/docs/Library-Dependencies.html#Ivy+revisions).
So, IIUC, if one want to stay in specific range, one can write:
```scala
// scala-steward:off
"com.typesafe.akka" %% "akka-actor" % "2.4.+",
"com.typesafe.akka" %% "akka-testkit" % "2.4.+",
// scala-steward:on
```

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/util/string.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/util/string.scala
@@ -60,39 +60,28 @@ object string {
   ): Option[MinLengthString[N]] =
     refineV[MinSize[N]](xs.reduceLeft(longestCommonPrefix)).toOption
 
-  final private val regexIgnoreMultiLinesBegins = "^\\s*//\\s*scala-steward:off".r
-
   /** Like `Regex.replaceSomeIn` but indicates via the return type if there
     * was at least one match that has been replaced.
     */
   def replaceSomeInOpt(
       regex: Regex,
       target: CharSequence,
+      splitter: String => Nel[(String, Boolean)],
       replacer: Regex.Match => Option[String]
   ): Option[String] = {
     var changed = false
     val replacer1 = replacer.andThen(_.map(r => { changed = true; r }))
     val targetString = target.toString
-    val result = if (!targetString.contains("scala-steward:off")) {
-      regex.replaceSomeIn(target, replacer1)
-    } else {
-      var ignoreLines = false
-      targetString.linesWithSeparators
-        .map(s => {
-          if (ignoreLines) {
-            if (s.contains("scala-steward:on")) {
-              ignoreLines = false
-              regex.replaceSomeIn(s, replacer1)
-            } else s
-          } else if (s.contains("scala-steward:off")) {
-            if (!ignoreLines && regexIgnoreMultiLinesBegins.findFirstIn(s).isDefined) {
-              ignoreLines = true
-            }
-            s
-          } else regex.replaceSomeIn(s, replacer1)
-        })
-        .mkString("")
-    }
+    val result = splitter(targetString)
+      .map {
+        case (str, canReplace) =>
+          if (canReplace) {
+            regex.replaceSomeIn(str, replacer1)
+          } else {
+            str
+          }
+      }
+      .mkString_("")
     if (changed) Some(result) else None
   }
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/util/string.scala
```diff
@@ -70,7 +72,29 @@ object string {
...
+    val result = if (!targetString.contains("scala-steward:off")) {
+      regex.replaceSomeIn(target, replacer1)
+    } else {
+      var ignoreLines = false
+      targetString.linesIterator
+        .map(s => {
+          if (ignoreLines) {
+            if (s.contains("scala-steward:on")) {
+              ignoreLines = false
+              regex.replaceSomeIn(s, replacer1)
```
fthomas: One concern I have is that the regular expressions now operate on single lines only. Before this change they operated on the whole content of files.

modules/core/src/main/scala/org/scalasteward/core/util/string.scala
```diff
@@ -70,7 +72,29 @@ object string {
...
+          } else {
+            if (s.contains("scala-steward:off")) {
+              if (ignoreLines || regexIgnoreMultiLinesBegins.findFirstIn(s).isDefined) {
+                ignoreLines = true
+              }
+              s
+            } else regex.replaceSomeIn(s, replacer1)
+          }
+        })
+        .mkString(System.lineSeparator())
```
fthomas: I think this could lead to huge diffs if the original file didn't use `System.lineSeparator()` as line separator.

modules/core/src/main/scala/org/scalasteward/core/util/string.scala
exoego: What is your worry?
I don't think operating on every single lines separately could be a problem
Also there is a fast-path (for most repositories), where the whole content is replaced one time if no `scala-steward:off` found.

modules/core/src/main/scala/org/scalasteward/core/util/string.scala
exoego: Changed from `str.lineIterator` to `str.linesWithSeparators` to preserve original end of lines in https://github.com/fthomas/scala-steward/pull/671/commits/d964a63744825d758baa22349c16af87ed2180b7

modules/core/src/main/scala/org/scalasteward/core/util/string.scala
fthomas: You're right that our current regex is line based so this is not a breaking change. But what if we want to use a regex that works on multiple lines in the future. Those wouldn't work in files that contain `scala-steward:off`.

What do you think about this: We split the content of a file into multiple parts where each part is associated with a `Boolean` that specifies if that part can be changed by Scala Steward. This would have a signature similar to this: `String => List[(String, Boolean)]`. We would then only apply our replace function on those parts where the `Boolean` is `true` (where we are allowed to do changes).

I have a feeling that this approach would also be more compositional.

modules/core/src/main/scala/org/scalasteward/core/util/string.scala
fthomas: I didn't know about `linesWithSeparators`. Nice!

modules/core/src/main/scala/org/scalasteward/core/util/string.scala
exoego: Refactored so in https://github.com/fthomas/scala-steward/pull/671/commits/4f28b79fbb89d1deaeb0333a1a983575e182c9bd

PR #676, opened 2019-07-09
Change groupId when dependencies published under a different groupId

### LINKED ISSUE #613: Change groupId when dependencies publish under a different groupId
Sometimes dependencies start publishing under a different `groupId` when maintainers change or the maintainers want to extend the pool of people able to publish new releases. One recent example is the [kind-projector](https://github.com/typelevel/kind-projector) project which was moved to the Typelevel organization:
 ```
"org.spire-math" %% "kind-projector" % "0.9.10" ->
"org.typelevel"  %% "kind-projector" % "0.10.0"
```
It would be nice if Scala Steward would also change the `groupId` in this instance. Currently this needs to be done manually.

@Philippus proposed this feature at Scala Days in Lausanne.

### PR DESCRIPTION
**Pull request by exoego**
Closes #613

## kind-projector

```sbt
"org.spire-math" %% "kind-projector" % "0.9.10" ->
"org.typelevel"  %% "kind-projector" % "0.10.0" 
```
Example PR  https://github.com/exoego/scala-steward-test-project/pull/91

##  sbt-scalafmt

```sbt
"com.geirsson"  %% "sbt-scalafmt" % "1.6.0-RC3" ->
"org.scalameta" %% "sbt-scalafmt" % "2.0.2"
```
Example PR https://github.com/exoego/scala-steward-test-project/pull/90

## Note
For 1st iteration, these old-to-newer relations are stored in `UpdateService.getNewerGroupId` and needs manually updated.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
@@ -23,6 +23,7 @@ import io.chrisdavenport.log4cats.Logger
 import org.scalasteward.core.application.Config
 import org.scalasteward.core.data.{Dependency, Update}
 import org.scalasteward.core.io.{FileAlg, FileData, ProcessAlg, WorkspaceAlg}
+import org.scalasteward.core.repocache.RepoCacheRepository
 import org.scalasteward.core.sbt.command._
 import org.scalasteward.core.sbt.data.{ArtificialProject, SbtVersion}
 import org.scalasteward.core.scalafix.Migration
@@ -58,6 +59,7 @@ object SbtAlg {
       processAlg: ProcessAlg[F],
       workspaceAlg: WorkspaceAlg[F],
       scalafmtAlg: ScalafmtAlg[F],
+      cacheRepository: RepoCacheRepository[F],
       F: Monad[F]
   ): SbtAlg[F] =
     new SbtAlg[F] {
@@ -136,7 +138,7 @@ object SbtAlg {
           updates <- withTemporarySbtDependency(repo) {
             exec(sbtCmd(commands), repoDir).map(parser.parseSingleUpdates)
           }
-          originalDependencies <- getOriginalDependencies(repo)
+          originalDependencies <- cacheRepository.getDependencies(List(repo))
           updatesUnderNewGroupId = originalDependencies.flatMap(
             UpdateService.findUpdateUnderNewGroup
           )
--- a/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
@@ -27,6 +27,7 @@ import org.scalasteward.core.sbt.command._
 import org.scalasteward.core.sbt.data.{ArtificialProject, SbtVersion}
 import org.scalasteward.core.scalafix.Migration
 import org.scalasteward.core.scalafmt.{scalafmtDependency, ScalafmtAlg}
+import org.scalasteward.core.update.UpdateService
 import org.scalasteward.core.util.Nel
 import org.scalasteward.core.vcs.data.Repo
 
@@ -91,9 +92,7 @@ object SbtAlg {
 
       override def getDependencies(repo: Repo): F[List[Dependency]] =
         for {
-          repoDir <- workspaceAlg.repoDir(repo)
-          cmd = sbtCmd(List(libraryDependenciesAsJson, reloadPlugins, libraryDependenciesAsJson))
-          lines <- exec(cmd, repoDir)
+          originalDependencies <- getOriginalDependencies(repo)
           maybeSbtVersion <- getSbtVersion(repo)
           maybeSbtDependency = maybeSbtVersion.flatMap(sbtDependency)
           maybeScalafmtVersion <- scalafmtAlg.getScalafmtVersion(repo)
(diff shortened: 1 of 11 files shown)


fthomas:
Wow, @exoego, you create pull requests faster than I can review them! :-)

modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
```diff
@@ -181,4 +181,20 @@ object UpdateService {
       case ("org.xerial.sbt", "sbt-pack")                    => false
       case _                                                 => true
     }
+
+  // (OldGroupId, ArtifactId) -> NewGroupId
+  final private val mapToNewGroupId = Map(
+    ("org.spire-math", "kind-projector") -> "org.typelevel",
```
kubukoz: I'd replace the second field in the key of this map with a predicate. In this case, `== "kind-projector"`, but if we were to add zio, we could do pretty much `("org.scalaz.zio", _ => true) -> "dev.zio"`.

Or just replace this with a partial function altogether and use a pattern match :)

exoego: Thanks !!
It helped me to understand that there could be needs for using new organization for multiple artifacts.

Replaced with pattern matching in https://github.com/fthomas/scala-steward/pull/676/commits/21fa1687b453401e53cf41c1d7e0dfbbaad77403 so one can add pattern easily like
```scala
case ("org.scalaz.zio", _) => "dev.zio"
case ("com.example", artifactId) if Set("foo", "bar")(artifactId) => "com.example.new"
```

kubukoz: Oops, I played myself. Zio was never part of the organization before. But a pattern match for modules starting with a specific prefix is still a thing!

exoego:
Ah, I tested this but it does not work.
Because current implementation try to find updates for 

* original `"org.spire-math" %% "kind-projector" % "0.9.10"`  and
* modified `"org.typelevel" %% "kind-projector" % "0.9.10"` 

but there is no `"org.typelevel" %% "kind-projector" % "0.9.10"`, so `"org.typelevel" %% "kind-projector" % "0.10.0"` is not found by `sbt-update`.

I will consider fix for this.

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
```diff
@@ -91,17 +92,22 @@ object SbtAlg {
...
             scalafmtDependency(defaultScalaBinaryVersion)
           )
         } yield (maybeSbtDependency.toList ++ maybeScalafmtDependency.toList ++
-          parser.parseDependencies(lines)).distinct
+          originalDependencies).distinct
+
+      def getOriginalDependencies(repo: Repo): F[List[Dependency]] =
+        for {
+          repoDir <- workspaceAlg.repoDir(repo)
+          cmd = sbtCmd(List(libraryDependenciesAsJson, reloadPlugins, libraryDependenciesAsJson))
```
exoego: Need investigation) It seems that `libraryDependenciesAsJson` command does not include compiler plugin.
Therefore `Update` for `kind-projector` is not found, since it is supposed to be used as `addCompilerPlugin("org.spire-math" %% "kind-projector" % "0.9.10")`.

exoego: `show libraryDependencies` includes compiler plugins,
```
sbt> show libraryDependencies
[info] project1 / libraryDependencies
[info]  List(org.scala-lang:scala-library:2.12.7, org.spire-math:kind-projector:0.9.10:plugin->default(compile))
```

but `show libraryDependenciesAsJson` do not.

```
sbt> show libraryDependenciesAsJson
[info] project1 / libraryDependenciesAsJson
[info]  [ { "groupId": "org.scala-lang", "artifactId": "scala-library", "artifactIdCross": "scala-library", "version": "2.12.7" } ]
```

exoego: `libraryDependenciesAsJson` is `StewardPlugin`'s feature.

This issue is fixed in https://github.com/fthomas/scala-steward/pull/676/commits/a02f3fffa638fe122e84e87b74dbfc5d663d6e02 and now compiler plugins get updated.
Example https://github.com/exoego/scala-steward-test-project/pull/91

fthomas:
This is awesome!

modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
```diff
@@ -130,7 +136,11 @@ object SbtAlg {
           updates <- withTemporarySbtDependency(repo) {
             exec(sbtCmd(commands), repoDir).map(parser.parseSingleUpdates)
           }
-        } yield updates
+          originalDependencies <- getOriginalDependencies(repo)
```
fthomas: This would be the second sbt call in `getUpdatesForRepo`. Since sbt takes a long time to load, this will nearly double the time needed to execute this method. And this is already one of the most expensive calls when working on a repository. I guess my public instance of Scala Steward would take over one hour more for a full run just for this additional sbt call.

I see two options to speed this up:
1. Use `RepoCacheAlg` here to get the dependencies of a repo.
1. Only call sbt once to get updates and dependencies.

I hadn't the time to thought about the full implications of either of these approaches.

exoego: I picked first option in https://github.com/fthomas/scala-steward/pull/676/commits/b51a779a506cffdd4c9350100231db40c92a4128 and it seems to work well.
Before test run, I cleared out `workspace`.

fthomas: Thanks! In the future we might want to have a new method that checks if the cache is up-to-date and refreshes it if is outdated before returning dependencies. Because right now, the caches are only up-to-date if Scala Steward is run with `--prune-repos=true`. But this OK for now.

Philippus:
It's a little unclear to me what happens if there are still updates available under the old group id. Does sbt-steward first do incremental updates until the latest version under the old group id? And only then an update under the new group id is created?

Philippus:
Probably doesn't happen very often, but besides a groupId changing the artifactId could change as well. For example in lz4-java both groupId and artifactId changed (https://github.com/lz4/lz4-java/blob/master/CHANGES.md#140)

PR #692, opened 2019-07-11
Rename DependencyService to RepoCacheAlg and store sbt versions

fthomas:
Storing the sbt versions will be useful later to determine which repos
use an outdated sbt version.


fthomas:
The second commit also caches `RepoConfig`s so that we can exclude ignored updates in the update check in `UpdateService`.

PR #694, opened 2019-07-11
List applied Scalafix rules in the PR body.

### LINKED ISSUE #593: List applied Scalafix rules in the PR body
If scala-steward applies Scalafix migrations for an update, those migrations should be mentioned in the body of the pull request. And maybe adding a label like `scalafix-migrations` would be nice too.

exoego:
Closes #593

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/vcs/data/NewPullRequestData.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/vcs/data/NewPullRequestData.scala
@@ -23,7 +23,7 @@ import org.scalasteward.core.git.Branch
 import org.scalasteward.core.model.{SemVer, Update}
 import org.scalasteward.core.nurture.UpdateData
 import org.scalasteward.core.repoconfig.RepoConfigAlg
-import org.scalasteward.core.git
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/org/scalasteward/core/vcs/data/NewPullRequestData.scala
```diff
@@ -61,11 +64,29 @@ object NewPullRequestData {
         |${RepoConfigAlg.configToIgnoreFurtherUpdates(update)}
         |```
         |</details>
-        |
-        |${semVerLabel(update).fold("")("labels: " + _)}
+        |${appliedMigrations.getOrElse("")}
+        |labels: ${labels.mkString(", ")}
```
fthomas: @exoego I think we shouldn't add the string `labels:` if `labels` is empty.

exoego: Updated in https://github.com/fthomas/scala-steward/pull/694/commits/d7f9771140614c5ebd9302c98668d6236ef5c53c

PR #719, opened 2019-07-16
Add ability to update version in .scalafmt.conf.

### LINKED ISSUE #625: Update scalafmt version
scalafmt gets its version from the `.scalafmt.conf` file, like here:

https://github.com/akka/alpakka/blob/06f323f9b2940a18fc6dd63d5127715847341e01/.scalafmt.conf#L1

It would be awesome if scala-steward could upgrade that when a new scalafmt version comes out.

exoego:
Closes #625

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
@@ -24,7 +24,7 @@ import io.chrisdavenport.log4cats.Logger
 import org.scalasteward.core.data.Update
 import org.scalasteward.core.io.{isFileSpecificTo, isSourceFile, FileAlg, WorkspaceAlg}
 import org.scalasteward.core.sbt.SbtAlg
-import org.scalasteward.core.scalafix
+import org.scalasteward.core.{scalafix, scalafmt}
 import org.scalasteward.core.scalafmt.ScalafmtAlg
 import org.scalasteward.core.util._
 import org.scalasteward.core.vcs.data.Repo
@@ -41,7 +41,7 @@ final class EditAlg[F[_]](
   def applyUpdate(repo: Repo, update: Update): F[Unit] =
     for {
       _ <- applyScalafixMigrations(repo, update)
-      _ <- scalafmtAlg.editScalafmtConf(repo)
+      _ <- applyScalafmtConfUpdate(repo, update)
       repoDir <- workspaceAlg.repoDir(repo)
       files <- fileAlg.findSourceFilesContaining(
         repoDir,
@@ -67,4 +67,10 @@ final class EditAlg[F[_]](
       case None =>
         F.unit
     }
+
+  def applyScalafmtConfUpdate(repo: Repo, update: Update): F[Unit] =
+    if (scalafmt.isScalafmtUpdate(update))
+      scalafmtAlg.editScalafmtConf(repo, update.nextVersion)
+    else
+      F.unit
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
@@ -25,6 +25,7 @@ import org.scalasteward.core.data.Update
 import org.scalasteward.core.io.{isFileSpecificTo, isSourceFile, FileAlg, WorkspaceAlg}
 import org.scalasteward.core.sbt.SbtAlg
 import org.scalasteward.core.scalafix
+import org.scalasteward.core.scalafmt.ScalafmtAlg
 import org.scalasteward.core.util._
 import org.scalasteward.core.vcs.data.Repo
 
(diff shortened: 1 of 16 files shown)


fthomas:
This looks good to me @exoego. There is only one thing that bothers me and that is that we're modifying/using `UpdateHeurisitc` to change `.scalafmt.conf`. We don't need any heuristic to change the scalafmt version number since it will be the same in all projects because we know exactly where and what we need to change. It is different from other version numbers which can appear anywhere in the build. (Btw, the same is true for the sbt version, but we didn't need to change `UpdateHeuristic` for that.)

exoego:
Done so.

fthomas:
Thanks for the update. I don't see `editScalafmtConf` being used anywhere except in tests. I think it is missing in `EditAlg`.

exoego:
Oops.
Added to `EditAlg` in https://github.com/fthomas/scala-steward/pull/719/commits/06c48c4c74a69051a3d3b32202addedc60570c8a

modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
```diff
@@ -33,12 +34,14 @@ final class EditAlg[F[_]](
...
     logger: Logger[F],
     sbtAlg: SbtAlg[F],
+    scalafmtAlg: ScalafmtAlg[F],
     workspaceAlg: WorkspaceAlg[F],
     F: Sync[F]
 ) {
   def applyUpdate(repo: Repo, update: Update): F[Unit] =
     for {
       _ <- applyScalafixMigrations(repo, update)
+      _ <- scalafmtAlg.editScalafmtConf(repo)
```
fthomas: Wouldn't this update `.scalafmt.conf` for any `update`? I guess it should only be called if `update` is an `Update` for `scalafmt`, right?

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -80,6 +84,11 @@ final class NurtureAlg[F[_]](
       }
     } yield ()
 
+  def getNonSbtUpdates(repo: Repo): F[List[Update.Single]] =
+    for {
+      maybeScalafmt <- scalafmtAlg.getScalafmtUpdate(repo)
+    } yield List(maybeScalafmt).flatten
```
jhnsmth:
```suggestion
    } yield maybeScalafmt.toList
```

exoego: This is intended so non-sbt update can be added like this.
```scala
for {
  maybeScalafmt <- scalafmtAlg.getScalafmtUpdate(repo)
  maybeFoo <- fooAlg.getUpdate()
} yield List(maybeScalafmt, maybeFoo).flatten
```

modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
exoego: True.
Fixed in https://github.com/fthomas/scala-steward/pull/719/commits/10212f80ceccd522052d2690ec197b2d5f8f1dbb

exoego:
Example of PR to update `.scalafmt.conf` https://github.com/exoego/scalafmt/pull/21

fthomas:
I just tried this branch (at dbd95c712934bfd788c40e6bbe1e0492aae3dc1f) on one of my repos: https://github.com/fthomas/fs2-cron/pull/42. The complete content of the `.scalafmt.conf` is in the PR body and all but the `version` line has been removed from the `.scalafmt.conf` file.

modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
```diff
@@ -33,12 +34,14 @@ final class EditAlg[F[_]](
...
     logger: Logger[F],
     sbtAlg: SbtAlg[F],
+    scalafmtAlg: ScalafmtAlg[F],
     workspaceAlg: WorkspaceAlg[F],
     F: Sync[F]
 ) {
   def applyUpdate(repo: Repo, update: Update): F[Unit] =
     for {
       _ <- applyScalafixMigrations(repo, update)
+      _ <- applyScalafmtConfUpdate(repo, update)
```
fthomas: I'm wondering since there is no `org.scalameta:scalafmt` artifact if we should just skip `applyScalafixMigrations` and `applyUpdateTo` for scalafmt updates. By doing this, we'd ensure that only `.scalafmt.conf` is edited. What do you think?

exoego: Yeah, that sounds reasonable.
Fixed in https://github.com/fthomas/scala-steward/pull/719/commits/8274417bac6e194c07d4126b7aa5a61b7203138f
We no longer see `No files found that contain the current version` messages.

exoego:
> 
> 
> I just tried this branch (at [dbd95c7](https://github.com/fthomas/scala-steward/commit/dbd95c712934bfd788c40e6bbe1e0492aae3dc1f)) on one of my repos: [fthomas/fs2-cron#42](https://github.com/fthomas/fs2-cron/pull/42). The complete content of the `.scalafmt.conf` is in the PR body and all but the `version` line has been removed from the `.scalafmt.conf` file.

Sorry for broken commit
I thinks it is now fixed. See this example https://github.com/exoego/fs2-cron/pull/1/files

fthomas:
LGTM now, many thanks @exoego!

Btw, if you force-push your changes to your PR branches it makes it harder for me to test them locally. It would be great if you could push your changes without using `--force` in the future.

exoego:
> Btw, if you force-push your changes to your PR branches it makes it harder for me to test them locally. It would be great if you could push your changes without using `--force` in the future.

OK, I will try to avoid force-push.
The reason why I use it is to cleanup commit history.
I squashed WIP commits and then force-push.

Issue #689, opened 2019-07-11
Errors when using --do-not-fork with repos owned by groups on GitLab


yaroot:
There is no `owner` field for repos owned by a group, I [made](https://github.com/yaroot/scala-steward/commit/b88e18a1793ca6609659c00cfc7ea71b7a0630ec) a quick & dirty fix to read _namespace/name_ if _name_ field failed to decode, but I don't think this is a good solution.

ping @daddykotex

daddykotex:
I've never tested `--do-not-fork` but it seems to me that this would be the correct fix

tyrcho:
I had the same issue, would be glad to see this fixed !

tyrcho:
I've tested your version @yaroot and it seems to work. I have other issues but they don't seem related to the owner property.

daddykotex:
My pull requests addresses the 404 you are seeing: https://github.com/fthomas/scala-steward/pull/738

tyrcho:
@yaroot I think this has been fixed with the latest changes in master. At least on my side I don't have the issue anymore.

PR #722, opened 2019-07-17
Fix errors when running with gitlab repos owned by groups with --do-not-fork flag

yaroot:
as discussed in #689, I also removed `parentRepoOutDecoder` as it's mostly the same with `repoOutDecoder`

@daddykotex


daddykotex:
There is also a second issue I had with this option, now that I've tried it a bit. When comes the time to create a MR, the code automatically change the "group/repo" to "owner/repo". See L127 of `Http4sGitlabApiAlg`.

daddykotex:
I'll submit another PR tomorrow.

fthomas:
Many thanks, @yaroot!

Issue #384, opened 2019-04-05
Automatic branch update and forced pushes can potentially flood CI


renatocaval:
As discussed here: https://github.com/playframework/playframework/pull/9159#issuecomment-479859679

We are facing an issue in which Travis is being flooded with new builds over and over. That happens because we have had a few PRs opened by @scala-steward and on each new merge on `master`, @scala-steward updates the branch and sends a new push. 

I personally don't think that this should be the default behaviour because it can potentially hijack CI for long periods. 

Maintainers have a few options to enforce this, for instance, they can configure branch protection which will force them to update the branch via GitHub or they can use something like Mergify that does it automatically and sequenced without triggering many builds at the same time.

So, ideally, either @scala-steward simply leave the responsibility to the project maintainers or over a flag in `.scala-steward.conf` to disable it.

fthomas:
Btw, this is mainly done to resolve merge conflicts which easily happen with scala-steward's PR since version numbers tend to be defined relatively adjacent in build files.

If scala-steward could figure out if a PR has conflicts with the base branch we could update PRs only if there are conflicts. That sounds like a good compromise between keeping PRs automatically mergeable and not uneccessarily stressing CI systems.

dwijnand:
https://developer.github.com/v3/git/#checking-mergeability-of-pull-requests

But I'm not sure if "mergeability" means pure git (i.e. has merge conflicts or not) or PR mergeability (requires review, etc).

fthomas:
Thanks for the hint. Since this is just another GitHub call it would be straightforward to implement.

Another option might be just using Git as shown in https://stackoverflow.com/a/501461/460387. The benefit of just using Git is that it would also work if scala-steward is ported to Bitbucket or GitLab

PR #724, opened 2019-07-17
Update PRs only if there are conflicts with the base branch

fthomas:
This changes when Scala Steward updates pull requests. Prior to this
change PRs were updated when they were behind the base branch. This
resolved potential merge conflicts but could also flood CI systems if
there were multiple open PRs by Scala Steward. With this change, only
PRs which are actually in conflict with the base branch are updated.
This should dramatically reduce the number of times Scala Steward
triggers CI builds by updating PRs.

closes: #384

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/git/GitAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/git/GitAlg.scala
@@ -22,7 +22,7 @@ import cats.implicits._
 import org.http4s.Uri
 import org.scalasteward.core.application.Config
 import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
-import org.scalasteward.core.util.{MonadThrowable, Nel}
+import org.scalasteward.core.util.{BracketThrowable, Nel}
 import org.scalasteward.core.vcs.data.Repo
 
 trait GitAlg[F[_]] {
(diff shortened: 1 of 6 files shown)


kluen:
Hi Thomas,

As far as I can see there is no setting to get the old behaviour (Please correct me if I'm wrong). Any chance this will be configurable in the future?
The current behaviour assumes that if there are no merge conflicts, the build should be fine. However, if in the meantime another library was updated, there might be incompatibilities not covered by tests.
I agree that the current behaviour should be the default, but an option for the safer version would be nice.

fthomas:
@kluen I've changed my mind about this topic again and I am now open to add a repository-specific configuration for updating PRs even if there are no merge conflicts. I acknowledge that there are circumstances in which updating PRs is very convenient and flooding the CI system is not a concern.

kluen:
@fthomas that's great news. Is there an issue for this yet or is it rather some possibility for the future?
Btw. I just realized your name is Frank, not Thomas. Sorry about that ;)

fthomas:
@kluen No there is no issue for this yet. Would you like to open one? I think the required changes are straightforward. We need a new flag in `RepoConfig` and use that to short-circuit this [`hasConflicts`](https://github.com/fthomas/scala-steward/blob/7afc0d2e23f712991798c58e48b9b7e108df0db5/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala#L204-L207) check to always return `true`. The `RepoConfig` can be accessed in `shouldBeUpdated` with `data.repoConfig`.

PR #732, opened 2019-07-18
document sbt 0.13.x not working out of the box

robstoll:


docs/faq.md
```diff
@@ -30,3 +30,7 @@ number adheres to [Semantic Versioning](https://semver.org/)) as demonstrated
 
 Other examples of Mergify rules for Scala Steward can be found via
 [GitHub's code search](https://github.com/search?p=6&q=%22author%3Dscala-steward%22+filename%3A.mergify.yml&type=Code).
+
+### Does it work with sbt 0.13.x
+
+not out of the box as it depends on dependencyUpdates task. However, adding the following sbt plugin https://github.com/rtimush/sbt-updates (requires sbt 0.13.9) does the job.
```
fthomas: Scala Steward installs sbt-updates as global plugin for the 0.13 and 1.0 sbt series on start-up. So using it with sbt 1 and sbt 0.13 projects should work out of the box. At least I'm not doing anything special for 0.13 projects and there are a lot of 0.13 projects in `repos.md`. I'm wondering what you mean with "adding sbt-updates does the job"? Do you need to add it to the projects build or as a global plugin? What happens if you don't do it?

robstoll: Hm... I had to install the plugin I mentioned in the project which scala-steward should take care of. Without it I got the error as shown in the [error.log](https://github.com/fthomas/scala-steward/files/3407292/error.log), with the plugin it worked.

fthomas: It seems to me that sbt is not picking up the global plugin that is installed by `SbtAlg.addGlobalPlugins`. It should have been installed into `$HOME/.sbt/0.13/plugins/scala-steward.sbt`. Can you find this file and check if it is available in the environment you're running Scala Steward?

robstoll: I think I have time to check that in the second part of next week. As additional information, I am running it from gitlab, more will follow.

robstoll: I cannot reproduce the error anymore. I am not sure why it happened but I changed a few things:
- I am using a docker image which has sbt 1.2.8 installed, before it was something along the line of 1.3.8
- I was executing it within firejail (now I am using --disable-sandbox)

Anyway could also something different. Must have been a coincidence that it worked after I installed the plugin. In this sense you can close this pull request IMO

PR #747, opened 2019-07-19
Colisweb is using Scala Steward

tyrcho:


fthomas:
Many thanks, @tyrcho!

PR #750, opened 2019-07-21
Filter Gitlab MRs on target branch

tyrcho:
I noticed that for Gitlab, Steward will consider an outdated MR with the wrong target branch. 

For Github it was already taken into account.

I could not test this patch properly yet (different setup home/work) but I can probably test it during the week.


tyrcho:
tested successfully on our Gitlab projects ! 
ready to merge from my side.

PR #757, opened 2019-07-23
Merge instead of reset/force-push to resolve conflicts

### LINKED ISSUE #486: Configurable use of merge rather than rebase/force-push
It's not at all uncommon for a dependency update to require further work before merge. Generally we (at slamdata) would want to do this by basing on scala-steward's PR branch, making some changes, and pushing them. For larger updates, multiple people may be doing this independently and/or collaborating. Force pushing the branch breaks all of this.

I would love it if we could configure (presumably on a per-repo basis) the force push behavior, causing scala-steward to use merge rather than rebase.

fthomas:
Closes #486


fthomas:
The changes in this PR can be seen in action here: https://github.com/fthomas/refined-sjs-example/pull/14

fthomas:
This change can be seen in action here: https://github.com/fthomas/refined-sjs-example/pull/15/commits

PR #760, opened 2019-07-25
exit with error if nurture fails

robstoll:
Currently failures like 404, 401 or 403 all end with success. Which means a build (I am using gitlab) is green instead of red.
I am very new to cats, so maybe I interpreted it wrong that methods ending with `_` return F[Unit]. Due to this I renamed attemptLog_


robstoll:
I have another change which assures that an error in creating a pull request does not stop the process + the process fails in the end. Shall I open a separate pull request or change this one?

fthomas:
Since I've already merged this PR, the other change you mentioned needs to be in another PR. :-) As a general rule I prefer smaller but more PRs since this makes reviewing easier.

robstoll:
That's perfectly fine, I prefer smaller PRs as well (writing and reviewing).
I'll create the PR for fail-safe pull request updating next week

PR #762, opened 2019-07-25
Include a link to the upgraded project in the PR 

### LINKED ISSUE #82: Include a link to the upgraded project in the PR
... in a perfect world to its release notes.

exoego:
Closes #82

TODO:
- [x] Invoke `CoursierAlg` somewhere in main algorithm.
- [x] <del>Give `Scm` info high priority over `homePage`, once https://github.com/coursier/coursier/pull/1291 is merged to `coursier` and pulished.  It is benefical even if only`homePage`. I will add `Scm` in Subsequent PR.</del> Done in this PR.
- [x] <del>Consider better interfaction between `coursier` and  `cats`, in terms of code simplification and runtime efficiency. It should be refactored later, once `coursier 2.0.0-RC2-7` or later published that includes https://github.com/coursier/coursier/pull/1294</del>Done in this PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -54,6 +54,17 @@ lazy val core = myCrossProject("core")
       Dependencies.scalaTest % Test
     ),
     assembly / test := {},
+    assemblyMergeStrategy in assembly := {
+      val nativeSuffix = "\\.(?:dll|jnilib|so)$".r
+
+      {
+        case PathList(ps @ _*) if nativeSuffix.findFirstMatchIn(ps.last).isDefined =>
+          MergeStrategy.first
+        case otherwise =>
+          val defaultStrategy = (assemblyMergeStrategy in assembly).value
+          defaultStrategy(otherwise)
+      }
+    },
     buildInfoKeys := Seq[BuildInfoKey](scalaVersion, sbtVersion),
     buildInfoPackage := moduleRootPkg.value,
     initialCommands += s"""
--- a/build.sbt
+++ b/build.sbt
@@ -36,6 +36,7 @@ lazy val core = myCrossProject("core")
       Dependencies.circeRefined,
       Dependencies.circeExtras,
       Dependencies.commonsIo,
+      Dependencies.coursierCore,
       Dependencies.fs2Core,
       Dependencies.http4sBlazeClient,
       Dependencies.http4sCirce,
--- a/build.sbt
+++ b/build.sbt
@@ -37,6 +37,7 @@ lazy val core = myCrossProject("core")
       Dependencies.circeExtras,
       Dependencies.commonsIo,
       Dependencies.coursierCore,
+      Dependencies.coursierCatsInterop,
       Dependencies.fs2Core,
       Dependencies.http4sBlazeClient,
       Dependencies.http4sCirce,
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -16,11 +16,13 @@
 
 package org.scalasteward.core.coursier
 
-import java.util.concurrent.ExecutorService
-
 import cats.effect._
 import cats.implicits._
+import cats.Parallel
+import coursier.interop.cats._
 import org.scalasteward.core.data.Dependency
+
+import scala.concurrent.ExecutionContext
 
 trait CoursierAlg[F[_]] {
   def getArtifactUrl(dependency: Dependency): F[Option[String]]
@@ -28,29 +30,15 @@ trait CoursierAlg[F[_]] {
 }
 
 object CoursierAlg {
-  // TODO: Use coursier.interop.cats._ in https://github.com/coursier/coursier/pull/1294 when published
-  implicit def createCoursierSync[F[_]](
-      implicit
-      F: Sync[F]
-  ): coursier.util.Sync[F] = new coursier.util.Sync[F] {
-    override def point[A](a: A): F[A] = F.point(a)
-    override def bind[A, B](elem: F[A])(f: A => F[B]): F[B] = F.flatMap(elem)(f)
-    override def delay[A](a: => A): F[A] = F.delay(a)
-    override def fromAttempt[A](a: Either[Throwable, A]): F[A] = F.fromEither(a)
-    override def handle[A](a: F[A])(f: PartialFunction[Throwable, A]): F[A] =
-      F.handleError(a)(f.apply)
-    override def gather[A](elems: Seq[F[A]]): F[Seq[A]] =
-      elems.foldLeft(F.delay(Seq.empty[A])) {
-        case (fseq, f) =>
-          fseq.flatMap(seq => f.map(seq :+ _))
-      }
-    override def schedule[A](unused: ExecutorService)(f: => A): F[A] = F.delay(f)
-  }
-
   def create[F[_]](
       implicit
       F: Sync[F]
   ): CoursierAlg[F] = {
+    implicit val P = Parallel.identity[F]
+    implicit val cs: ContextShift[F] = new ContextShift[F] {
+      override def shift: F[Unit] = F.unit
+      override def evalOn[A](ec: ExecutionContext)(fa: F[A]): F[A] = F.defer(fa)
+    }
     val cache = coursier.cache.FileCache[F]()
     val fetch = coursier.Fetch[F](cache)
     new CoursierAlg[F] {
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -24,6 +24,7 @@ import org.scalasteward.core.data.Dependency
 
 trait CoursierAlg[F[_]] {
   def getProjectHomepage(dependency: Dependency): F[Option[String]]
+  def getProjectHomepages(dependencies: List[Dependency]): F[Map[String, String]]
 }
 
 object CoursierAlg {
@@ -55,20 +56,32 @@ object CoursierAlg {
         )
 
         for {
-          fetchResult <- fetch
+          maybeFetchResult <- fetch
             .addDependencies(
               coursier.Dependency.of(module, dependency.version).withTransitive(false)
             )
             .addArtifactTypes(coursier.Type.pom)
             .ioResult
+            .map(Option.apply)
+            .recover {
+              case _: coursier.error.ResolutionError => None
+            }
         } yield {
-          fetchResult.resolution.projectCache.get((module, dependency.version)).flatMap {
-            case (_, project) =>
-              // TODO: use Scm info if published https://github.com/coursier/coursier/pull/1291
-              Option(project.info.homePage)
-          }
+          maybeFetchResult.flatMap(
+            _.resolution.projectCache.get((module, dependency.version)).flatMap {
+              case (_, project) =>
+                // TODO: use Scm info if published https://github.com/coursier/coursier/pull/1291
+                Option(project.info.homePage)
+            }
+          )
         }
-      }
+
+      override def getProjectHomepages(dependencies: List[Dependency]): F[Map[String, String]] =
+        for {
+          entries <- dependencies.traverse(dep => {
+            getProjectHomepage(dep).map(dep.artifactId -> _.getOrElse(""))
+          })
+        } yield Map(entries.filter { case (_, url) => url =!= "" }: _*)
     }
   }
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -62,8 +62,8 @@ object CoursierAlg {
           maybeFetchResult.flatMap(
             _.resolution.projectCache.get((module, dependency.version)).flatMap {
               case (_, project) =>
-                // TODO: use Scm info if published https://github.com/coursier/coursier/pull/1291
-                Option(project.info.homePage)
+                val info = project.info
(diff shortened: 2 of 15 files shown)


modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -0,0 +1,56 @@
...
+trait CoursierAlg[F[_]] {
+  def getProjectHomepage(dependency: Dependency): F[Option[String]]
+}
+
+object CoursierAlg {
+  def create[F[_]](
+      implicit
+      F: Sync[F]
+  ): CoursierAlg[F] = {
+    implicit val coursierSync: coursier.util.Sync[F] = new coursier.util.Sync[F] {
```
exoego: This anonymous sub class of `coursier.util.Sync[F]` works but may be not efficient...
I could not figure out a good way to inter-operate with `cats` and `coursier`.

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -0,0 +1,72 @@
...
+    new CoursierAlg[F] {
+      override def getProjectHomepage(dependency: Dependency): F[Option[String]] = {
+        val module = coursier.Module(
+          coursier.Organization(dependency.groupId),
+          coursier.ModuleName(dependency.artifactIdCross)
+        )
+
+        for {
+          fetchResult <- fetch
+            .addDependencies(coursier.Dependency.of(module, dependency.version))
```
alexarchambault: You can call `.withTransitive(false)` on `Dependency`, as it seems you don't need its transitive dependencies here.

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
alexarchambault: With a `cats.effect.ContextShift[F]`, `import coursier.interop.cats._` should automatically provide a `coursier.util.Sync[F]`.

exoego: Hmm, I tried below 
```scala
    import coursier.interop.cats._
    implicit val cs = cats.effect.ContextShift[F]
    implicit val coursierSync: coursier.util.Sync[F] = implicitly
    val cache = coursier.cache.FileCache[F]()
    val fetch = coursier.Fetch[F](cache)(coursierSync)
```
but
```
[error] ./IdeaProjects/Scala-Steward/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala:33:47: Cannot find an implicit value for ContextShift[F]:
[error] * import ContextShift[F] from your effects library
[error] * if using IO, use cats.effect.IOApp or build one with cats.effect.IO.contextShift
[error]     implicit val cs = cats.effect.ContextShift[F]
[error]                                               ^
```

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
exoego: Thanks !!
Applied change.

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
exoego: IIUC, `coursier.interop.cats.catsIOSync` accepts only `ContextShift[IO]`, not `ContextShift[F]`, so I think there is no automatic way right now.

exoego:
Example of PRs with links to the upgraded projects
* https://github.com/exoego/scalafmt/pull/3
* https://github.com/exoego/scalafmt/pull/4

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
alexarchambault: Indeed, https://github.com/coursier/coursier/pull/1294 should help.

modules/core/src/test/scala/org/scalasteward/core/coursier/CoursierAlgTest.scala
```diff
@@ -0,0 +1,22 @@
...
+import org.scalasteward.core.mock.MockState
+import org.scalatest.{FunSuite, Matchers}
+
+class CoursierAlgTest extends FunSuite with Matchers {
+
+  test("getArtifactUrl") {
+    val dep = Dependency("org.typelevel", "cats-effect", "cats-effect_2.12", "1.0.0")
+    coursierAlg
+      .getArtifactUrl(dep)
+      .runS(MockState.empty)
```
cb372: I'm a bit confused by this test. It doesn't seem to be testing anything. Don't you want to call `runA(MockState.empty)` and then assert that the artifact URL is what you expect?

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -0,0 +1,80 @@
...
+            .map(Option.apply)
+            .recover {
+              case _: coursier.error.ResolutionError => None
+            }
+        } yield {
+          maybeFetchResult.flatMap(
+            _.resolution.projectCache.get((module, dependency.version)).flatMap {
+              case (_, project) =>
+                val info = project.info
+                info.scm.map(_.url).orElse(Option(info.homePage)).filter(_.nonEmpty)
```
cb372: Is it ever possible for `scm` to be present but `scm.url` to be null? Might be worth wrapping it in `Option(...)` just in case, like you do with `info.homePage`.

exoego: Thanks !
Added defensive `Option` and `.filter(_.nonEmpty)` in https://github.com/fthomas/scala-steward/pull/762/commits/41f2629331e5a17b93506c3d07627437f64faaf7

modules/core/src/test/scala/org/scalasteward/core/coursier/CoursierAlgTest.scala
exoego: Thanks !
I had no idea how to get result, not `state`, but now figured out.
Added assertion in `https://github.com/fthomas/scala-steward/pull/762/commits/98521746ee7aefbd8b55ddc3c88214a142616236

fthomas:
Thanks for working on this @exoego! This already looks quite good, I only have two change requests. I would prefer if we wouldn't change `Update` for this since I think homepage or SCM URLs are only peripherally related to the data in `Update`. In my mind `Update` is basically just a dependency with a `from` and `to` version. Secondly I would only fetch these URLs if we're actually composing the body of a pull request because they are only used then. We shouldn't unnecessarily call Coursier if we don't make use of its result.

exoego:
> we wouldn't change Update for this since I think homepage or SCM URLs are only peripherally related to the data in Update

Done in https://github.com/fthomas/scala-steward/pull/762/commits/299b695fc9d24dfd604330b3b724ef952ca4f59e 


> Secondly I would only fetch these URLs if we're actually composing the body of a pull request because they are only used then.

Done in https://github.com/fthomas/scala-steward/pull/762/commits/eb3828ff9432527ac047d995de1e7b036fef32a7

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -74,12 +84,24 @@ final class NurtureAlg[F[_]](
       nonSbtUpdates <- getNonSbtUpdates(repo)
       updates = sbtUpdates ::: nonSbtUpdates
       filtered <- filterAlg.localFilterMany(repoConfig, updates)
+      dependencies <- sbtAlg.getDependencies(repo)
+      filteredDependencies = dependenciesInUpdates(dependencies, filtered)
+      mapping <- coursierAlg.getArtifactIdUrlMapping(filteredDependencies)
```
fthomas: In https://github.com/fthomas/scala-steward/pull/762#issuecomment-520037653 I meant to call  `coursierAlg.getArtifactIdUrlMapping` only in `NurtureAlg.createPullRequest` because only there these URLs are needed. There is no need to call Coursier if a pull request already exists and Scala Steward only updates/ignores it.

exoego: I see.
I think it is addressed in https://github.com/fthomas/scala-steward/pull/762/commits/2b3d60f45ed5f07930f1ac71984c2714ce907fcc

fthomas: Thanks! What I didn't realized until now is that the `sbtAlg.getDependencies` call is needed to get artifactIds with the cross version suffix and that we don't get them from `Update` because that only contains artifactIds without that suffix. Unfortunately `sbtAlg.getDependencies` is a expensive call because it runs sbt. Calling it in `createPullRequest` now means that we potentially call `getDependencies` more than once for the same project although it would always return the same result.

There are a few options to handle this:
* Find a way to add artifactIds with cross version suffixes to `Update`. I've no idea if we can get this information from sbt-updates.
* Use `RepoCacheAlg` because dependencies are already available in `RepoCache`. Unfortunately this is only true if `Config.pruneRepos` is `true`. If we want to use `RepoCacheAlg` in `NurtureAlg`, we would have to ensure that the caches are fresh even if `Config.pruneRepos` is false.
* Memoize the result of  `sbtAlg.getDependencies` so that it is called at most once.
* Go back to 2b3d60f45ed5f07930f1ac71984c2714ce907fcc so that `sbtAlg.getDependencies` is called once for every repo.
* Leave it as is and handle this in another PR to prevent more back and forth in this PR.

I really would prefer the first option but don't know if it is feasible at all. It would also require its own PR. To not further delay this feature I'm also okay with the last option. What do you think, @exoego?

And sorry for not realizing this earlier!

exoego: I think 3rd option (memoize) is good for now since it is not difficult to implement and do not degrade runtime speed, and runs on any configuration.
Then let's find a way to realize 1st option. (maybe open PR to sbt-update?)

I will implement memoize in SbtAlg using `mutable.Map.getOrElseUpdate`, but suggestion is appreciated if you have a better design.
Ofcourse it is OK to merge as-is.

fthomas: You could use `cats.effect.Async.memoize` to memoize `getDependencies` in `NurtureAlg`. I think there is no need to modify `SbtAlg` for that.

For the 1st option I'm currently looking at sbt-updates' `dependencyUpdatesData` task which returns a `Map[ModuleID, SortedSet[Version]]`. Maybe the `ModuleID` there is enough to get the artifactIds with the cross version suffix.

exoego: Added memoize in https://github.com/fthomas/scala-steward/pull/762/commits/80ee906d3b00575ff7e814dfc142eee249db0a55

build.sbt
```diff
@@ -52,6 +54,17 @@ lazy val core = myCrossProject("core")
...
+      val nativeSuffix = "\\.(?:dll|jnilib|so)$".r
+
+      {
+        case PathList(ps @ _*) if nativeSuffix.findFirstMatchIn(ps.last).isDefined =>
+          MergeStrategy.first
+        case otherwise =>
+          val defaultStrategy = (assemblyMergeStrategy in assembly).value
+          defaultStrategy(otherwise)
+      }
+    },
```
fthomas: Just to document why this is necessary, here are the errors we get without this change:
```
sbt:scala-steward> core/assembly
[info] Updating core...
[info] Done updating.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[error] 5 errors were encountered during merge
[error] java.lang.RuntimeException: deduplicate: different file contents found in the following:
[error] /home/frank/.ivy2/cache/org.fusesource.jansi/jansi/jars/jansi-1.18.jar:META-INF/native/linux32/libjansi.so
[error] /home/frank/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.12.8.jar:META-INF/native/linux32/libjansi.so
[error] deduplicate: different file contents found in the following:
[error] /home/frank/.ivy2/cache/org.fusesource.jansi/jansi/jars/jansi-1.18.jar:META-INF/native/linux64/libjansi.so
[error] /home/frank/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.12.8.jar:META-INF/native/linux64/libjansi.so
[error] deduplicate: different file contents found in the following:
[error] /home/frank/.ivy2/cache/org.fusesource.jansi/jansi/jars/jansi-1.18.jar:META-INF/native/osx/libjansi.jnilib
[error] /home/frank/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.12.8.jar:META-INF/native/osx/libjansi.jnilib
[error] deduplicate: different file contents found in the following:
[error] /home/frank/.ivy2/cache/org.fusesource.jansi/jansi/jars/jansi-1.18.jar:META-INF/native/windows32/jansi.dll
[error] /home/frank/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.12.8.jar:META-INF/native/windows32/jansi.dll
[error] deduplicate: different file contents found in the following:
[error] /home/frank/.ivy2/cache/org.fusesource.jansi/jansi/jars/jansi-1.18.jar:META-INF/native/windows64/jansi.dll
[error] /home/frank/.ivy2/cache/org.scala-lang/scala-compiler/jars/scala-compiler-2.12.8.jar:META-INF/native/windows64/jansi.dll
```

SethTisue:
awesome

ennru:
Looking forward to see this in action! Thank you.

PR #765, opened 2019-07-26
introduce execFromRepo, unitUnless and rename push to forcePush

robstoll:


robstoll:
@fthomas any thoughts about this? Let me know if I should improve something

fthomas:
I *think* using `--force` after #757 isn't necessary anymore, so I would change the implementation instead of renaming `push`. But this change should be in its own PR.

`unitUnless` does not pull its weight IMO.

It is nice that `execFromRepo` saves a few lines but if I want to combine it with a second `execFromRepo` call, I would probably go back to the for comprehension and switch back to `exec` (just like the other methods that use two `exec` calls). IMO this diminishes its usefulness.

PR #772, opened 2019-07-30
make updating pull requests fail safe

robstoll:
fail-safe in the sense that the error is recorded (and the whole
process fails at the end) but it does not break the process from
processing other PRs.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
@@ -22,6 +22,7 @@ import org.scalasteward.core.application.Config
 import org.scalasteward.core.data.Update
 import org.scalasteward.core.edit.EditAlg
 import org.scalasteward.core.git.{Branch, GitAlg}
+import org.scalasteward.core.implicits._
 import org.scalasteward.core.repoconfig.RepoConfigAlg
 import org.scalasteward.core.sbt.SbtAlg
 import org.scalasteward.core.scalafmt.ScalafmtAlg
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -77,37 +80,43 @@ final class NurtureAlg[F[_]](
...
       pullRequests <- vcsApiAlg.listPullRequests(data.repo, head, data.baseBranch)
-      _ <- pullRequests.headOption match {
+      result <- pullRequests.headOption match {
         case Some(pr) if pr.isClosed =>
-          logger.info(s"PR ${pr.html_url} is closed")
+          logger.info(s"PR ${pr.html_url} is closed").asRight
         case Some(pr) if data.repoConfig.updatePullRequests =>
-          logger.info(s"Found PR ${pr.html_url}") >> updatePullRequest(data)
+          logger.info(s"Found PR ${pr.html_url}") >>
+            updatePullRequest(data).attempt
```
fthomas: One potential problem I see here is that `updatePullRequest` may leave the working copy of the repository in a dirty state which may affect and corrupt the next `processUpdate` call. Suppose `updatePullRequest` leaves changed files around which will be committed during the next `processUpdate` call. I guess the only way to prevent problems caused by a dirty state is to use a fresh clone after `updatePullRequest` failed. But I would rather try to fix `updatePullRequest` instead of making it really fail safe.

robstoll: A lighter approach would be to do a hard reset in case of a failure and let it throw in case it fails

fthomas: I would first like to understand why you want to make `updatePullRequest` fail safe at all. In which cases does `updatePullRequest` fail and why can't we fix these problems first?

robstoll: As far as I remember the problems were not inherent to scala-steward. Once there was an IOException because of a network-glitch or similar, another time a user deleted the origin/update/Xy branch and this caused a problem and there were other user related problems. My first priority is actually not the fail safe aspect but that the process as a whole fails if a problem occurs during the update process. For me it is fine if I change the PR in this direction. The fail-safe aspect is IMO nice as it works around the above mentioned problems

robstoll:
@fthomas seems like this one is stuck so I am going to close it and open a new one which only deals with failing the process in case there is a failure during the update process.

Issue #626, opened 2019-06-25
Consider using immutable references for Scalafix rules


fthomas:
See https://twitter.com/fst9000/status/1143454066229600257 for motivation.

PR #777, opened 2019-07-30
Add scalafix migration for scalatest 3.1.0 

### LINKED ISSUE #771: Scalafix migration for scalatest 3.1.0
`scalatest` 3.1.0 will have long list of rewrites (See https://github.com/scalatest/scalatest/issues/1657).
Fortunately it offers scalafix in https://github.com/scalatest/autofix
It would be great if Scala Steward apply these scalafix migration when updating `scalatest`.

exoego:
Closes #771 

Unfortunately, [`github:` scheme](https://scalacenter.github.io/scalafix/docs/developers/tutorial.html#using-github) is not available for `scalatest/autofix`, since the repository has their arbitrary project structure (multiple scala projects in sub directories), IIUC.

So used [`https:` schema](https://scalacenter.github.io/scalafix/docs/developers/tutorial.html#using-http) to support arbitrary project structure.
Also, used a SHA1-marked immutable reference as stated in https://github.com/fthomas/scala-steward/issues/626 for security consideration.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/scalafix/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/scalafix/package.scala
@@ -51,6 +51,15 @@ package object scalafix {
         Nel.of("cats-core".r),
         Version("1.0.0"),
         Nel.of("github:fthomas/cats/Cats_v1_0_0?sha=update/scalafix")
+      ),
+      Migration(
+        "org.scalatest",
+        Nel.of("scalatest".r),
+        Version("3.1.0"),
+        Nel.of(
+          "https://raw.githubusercontent.com/scalatest/autofix/6168da0e2bd113872b7dcd22cad7688d97ef9381/3.0.x/rules/src/main/scala/org/scalatest/autofix/v3_0_x/RenameDeprecatedPackage.scala",
+          "https://raw.githubusercontent.com/scalatest/autofix/6168da0e2bd113872b7dcd22cad7688d97ef9381/3.1.x/rules/src/main/scala/org/scalatest/autofix/v3_1_x/RewriteDeprecatedNames.scala"
+        )
       )
     )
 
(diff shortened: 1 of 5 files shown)


exoego:
Ah, I have no idea how to test before `scalatest` 3.1.0 is published.
Once published, it is too late to test...

fthomas:
To test this, I think you could set the version of the migration to 3.1.0-SNAP13 and try it on a project which uses ScalaTest 3.1.0-SNAP12 but with the old names.

modules/core/src/main/scala/org/scalasteward/core/scalafix/package.scala
```diff
@@ -51,6 +51,16 @@ package object scalafix {
         Nel.of("cats-core".r),
         Version("1.0.0"),
         Nel.of("github:fthomas/cats/Cats_v1_0_0?sha=update/scalafix")
+      ),
+      Migration(
+        "org.scalatest",
+        Nel.of("scalatest".r),
+        Version("3.1.0"),
+        Nel.of(
+          "https://raw.githubusercontent.com/scalatest/autofix/6168da0e2bd113872b7dcd22cad7688d97ef9381/3.0.x/rules/src/main/scala/org/scalatest/autofix/v3_0_x/RenameDeprecatedPackage.scala",
```
fthomas: If this is a rule for ScalaTest 3.0.x, it should be a separate migration with a 3.0.x `Version`.

fthomas: https://github.com/scalatest/scalatest/issues/1657 seems to suggest that this is a rewrite for ScalaTest 3.0.6.

exoego: Yes.
But most of existing Steward-enabled repos should have been updated to  scalatest 3.0.8, if they merge Steward's  PR.
So I think migration for 3.0.6 are not performed for many repos, unless maintainers manually migrate.

exoego:
It worked when testing updates from `3.1.0-SNAP12` to `3.1.0-SNAP13`: https://github.com/exoego/scala-steward-test-project/pull/2

And I believe migration for `3.0.6` should be performed same time, as mentioned in https://github.com/fthomas/scala-steward/pull/777#discussion_r309418536

modules/core/src/main/scala/org/scalasteward/core/scalafix/package.scala
fthomas: But what about new repos that are added to Scala Steward that use ScalaTest < 3.0.6? Those projects would benefit if this migration is applied in an update to the latest 3.0.x version. Maybe we should think about changing the `version` field of `Migration` to a range of versions. We should open a new ticket to discuss this.

PR #783, opened 2019-08-02
Allow using a custom base image for docker builds

kubukoz:
Reason: Custom resolvers that are usually added in our CI pipeline before `sbt` starts, as well as adding certificates. This change allows me to build a base image with all that myself and use it to build the scala-steward docker image that'll work in my gitlab org.

Let me know if the name needs changing or there's anything else wrong with this ;)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -127,7 +127,7 @@ lazy val metadataSettings = Def.settings(
 
 lazy val dockerSettings = Def.settings(
   dockerCommands := Seq(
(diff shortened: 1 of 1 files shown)


fthomas:
I find the name a tad too generic. Maybe `DOCKER_BASE_IMAGE` would be more descriptive? Apart from that this LGTM.

kubukoz:
Sure, that makes lotsa sense

PR #788, opened 2019-08-05
Update Christopher Davenport g8 Template

ChristopherDavenport:
Heres what the upgrade to try to allow the changes to proliferate looks like. 

My curiosity is whether it can find default.properties or if it only finds the sbt plugins.

https://github.com/ChristopherDavenport/library.g8/pull/5

### DIFF
--- a/repos.md
+++ b/repos.md
@@ -59,7 +59,6 @@
 - ChocPanda/scalacheck-magnolia
 - ChristopherDavenport/agitation
 - ChristopherDavenport/bank
-- ChristopherDavenport/base.g8
 - ChristopherDavenport/cats-par
 - ChristopherDavenport/cats-scalacheck
 - ChristopherDavenport/cats-time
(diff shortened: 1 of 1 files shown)


fthomas:
Scala Steward only modifies [`*.scala`, `*.sbt`, `build.properties`, and `.travis.yml` files](https://github.com/fthomas/scala-steward/blob/100beb81fd622f8110e38fa979bfa88a8067ac0d/modules/core/src/main/scala/org/scalasteward/core/io/package.scala#L24), so it will ignore the `default.properties` file. If there is no other way to make this work, I've no problem allowing Scala Steward to edit `default.properties` files too.

ChristopherDavenport:
Not sure how else to do it, I could hard-code the library versions but allow variables for the other components.

fthomas:
That may be easiest option if you're only using them [here](https://github.com/ChristopherDavenport/library.g8/blob/0eaa25f2ca2db8c33b4a22c712698ecff3906958/src/main/g8/build.sbt#L25-L46).

PR #793, opened 2019-08-06
Erroneous Cats update

fthomas:
Scala Steward currently fails to update Cats correctly in the provided example. I think we should fix this by adding a new `UpdateHeuristic` that only matches `$groupdId %+ $artifactId %+ $version`. It would only match verbatim definitions of the `ModuleID`.


fthomas:
https://github.com/http4s/http4s/pull/2849 is another instance of this bug.

PR #808, opened 2019-08-09
add kind-projector, scalacheck

SethTisue:
fyi @non @ashawley @rickynils

### DIFF
--- a/repos.md
+++ b/repos.md
@@ -518,11 +518,13 @@
 - typelevel/discipline-specs2
 - typelevel/frameless
 - typelevel/jawn
+- typelevel/kind-projector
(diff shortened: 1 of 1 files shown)


fthomas:
This requires approval by a kind-projector and ScalaCheck maintainer. I'd like to avoid flooding projects with pull requests without their consent.

SethTisue:
I'm actually a co-maintainer in both repos these days. but it wouldn't hurt to wait a bit and see if @non has any objection

fthomas:
Sorry Seth, I wasn't aware of that! I'll merge tomorrow if there are no objections until then.

Issue #810, opened 2019-08-11
scala-steward can't run on sbt projects that depends on other sbt project that is downloaded by git submodule


tanishiking:
If the sbt project depends on the other sbt project that is downloaded using git submodule like this:

```
[submodule "submodule"]
  path = submodule
  url = git@github.com:test/submodule.git
  # Assuming that test/submodule is also a sbt project.
```

```
lazy val submodule = (project in file("submodule"))
```

Scala Steward will crash when it runs on such repository with "Changes not staged for commit".

This is because
- scala-steward edits the dependencies **in submodule project**
- Try to commit it
- But it is impossible to commit the changes in the submodule work tree, so `git commit` end up with "Changes not staged for commit".

## Solution
We can avoid this problem by using the `--ignore-submodules` option when checking the changes. https://git-scm.com/docs/git-status

PR #811, opened 2019-08-11
Ignore changes for git submodules when looking for changes

tanishiking:
This PR fixes https://github.com/fthomas/scala-steward/issues/810


fthomas:
This LGTM, thanks @tanishiking!

With this change Scala Steward does not commit when there are only changes in submodules. But what if there are changes in the repo and a submodule. Scala Steward will then still run `git commit -a` which probably also produces the error reported in #810.

PR #814, opened 2019-08-12
Add a link to diff (tags comparison view)

### LINKED ISSUE #794: Link to diff of changes for GitHub projects
Once #762 has been merged, I think a nice enhancement would be to link to a GitHub URL showing the changes, e.g. `https://github.com/typelevel/cats/compare/v2.0.0-M4...v2.0.0-RC1`

The logic would be something like:

1. Grab a list of the tags from GitHub (if the dependency is a public repo)?
2. Try to match the old and new version numbers to tag names. This could be really naive, e.g. version is `2.0.0-M4` so assume the tag name is `v2.0.0-M4`, or something a bit more sophisticated
3. Somehow validate whether the generated URL is valid? For public repos, this could be done by hitting the URL and checking it's not a 404.

Obviously there's quite a lot that can go wrong here, so this would be very much a "best effort" thing. If anything went wrong, we wouldn't display the link.

exoego:
Closes #794 

This PR adds a link to diff (tags comparison view) in the form of markup in message `Updates [com.example:foo](SCM URL HERE) [from 0.1.0 to 0.2.0](TAGS-COMPARE URL HERE)`.
The link will be available only if artifact has SCM URL in `pom.xml` and the tags-compare URL for `from`/`to` versions exists in SCM.

Below example showcases URLs (and its differences) for the 3 SCMs supported in this PR.

* https://github.com/circe/circe/compare/v0.12.0-M4...v0.12.0-RC1 format: `from 3dot to`
* https://gitlab.com/gitlab-org/gitlab-ce/compare/v12.1.3...v12.1.4 format: `from 3dot to`
* https://bitbucket.org/gluon-oss/ignite/compare/1.0.1..1.0.0#diff format: `to 2dot from #diff`, no `v` prefix before tag


modules/core/src/main/scala/org/scalasteward/core/util/HttpExistenceClient.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import cats.effect.Sync
+import cats.implicits._
+import org.http4s.client.Client
+import org.http4s.{Method, Request, Uri}
+
+final class HttpExistenceClient[F[_]: Sync](
+    implicit client: Client[F]
+) {
+  def exists(uri: String): F[Boolean] = exists(Uri.unsafeFromString(uri))
```
fthomas: `unsafeFromString` can throw. To catch exceptions in `F` this should be something like `F.fromEither(Uri.fromString(uri)).flatMap(exists)`

modules/core/src/main/scala/org/scalasteward/core/vcs/data/NewPullRequestData.scala
```diff
@@ -63,6 +68,13 @@ object NewPullRequestData {
         |""".stripMargin.trim
   }
 
+  def fromTo(update: Update, branchCompareUrl: Option[String]): String =
+    branchCompareUrl match {
+      case None => s"from ${update.currentVersion} to ${update.nextVersion}"
+      case Some(compareUrl) =>
+        s"[from ${update.currentVersion} to ${update.nextVersion}](${compareUrl})"
```
fthomas: `from ${update.currentVersion} to ${update.nextVersion}` is duplicated here.

modules/core/src/main/scala/org/scalasteward/core/vcs/VCSExtraAlg.scala
```diff
@@ -0,0 +1,48 @@
...
+
+trait VCSExtraAlg[F[_]] {
+  def getBranchCompareUrl(maybeRepoUrl: Option[String], update: Update): F[Option[String]]
+}
+
+object VCSExtraAlg {
+  def create[F[_]](
+      implicit
+      existenceClient: HttpExistenceClient[F],
+      F: Async[F]
```
fthomas: Why does this require `Async[F]`? It seems to me that `Monad[F]` would be enough.

modules/core/src/main/scala/org/scalasteward/core/util/HttpExistenceClient.scala
exoego: Fixed in https://github.com/fthomas/scala-steward/pull/814/commits/936417d2b6c61a14e92e32619ca4bd6f5ed63950

modules/core/src/main/scala/org/scalasteward/core/vcs/VCSExtraAlg.scala
exoego: No intention, I just refered my memory from last PR 😄 
Changed to weaker type classes in https://github.com/fthomas/scala-steward/pull/814/commits/869d65beb23d86dc7024f41b9e1dcadac3698b15

modules/core/src/main/scala/org/scalasteward/core/vcs/data/NewPullRequestData.scala
exoego: Extracted the duplicates in https://github.com/fthomas/scala-steward/pull/814/commits/06e5f33d68e25ceb25f57101dc057c8b40f79cc1

PR #820, opened 2019-08-12
Fix missing project URLs for sbt and scalafmt 

### LINKED ISSUE #819: Project URLs are missing for sbt and scalafmt
PR #762 and #814 introduced links to the updated artifact project.

However, they lacks support of `sbt` and `scalafmt`, since those dependencies are not actually included in `sbtAlg.getDependencies` therefore their URLs are not fetched by `coursierAlg`.

Examples of PR with no URL
* sbt https://github.com/pathikrit/dijon/pull/37
* scalafmt https://github.com/playframework/playframework/pull/9554

exoego:
Closes #819

Example PRs:

* sbt https://github.com/exoego/scala-steward-test-project/pull/12
* fscalafmt https://github.com/exoego/scala-steward-test-project/pull/13

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/scalafmt/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/scalafmt/package.scala
@@ -23,7 +23,7 @@ import org.scalasteward.core.util.Nel
 package object scalafmt {
   val latestScalafmtVersion: Version = Version("2.0.1")
   val scalafmtGroupId = "org.scalameta"
-  val scalafmtArtifactId = "scalafmt"
+  val scalafmtArtifactId = "scalafmt-core"
 
   def findNewerScalafmtVersion(currentVersion: Version): Option[Version] =
     if (currentVersion < latestScalafmtVersion) Some(latestScalafmtVersion) else None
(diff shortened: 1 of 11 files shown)


modules/core/src/main/scala/org/scalasteward/core/scalafmt/package.scala
```diff
@@ -23,7 +23,7 @@ import org.scalasteward.core.util.Nel
 package object scalafmt {
   val latestScalafmtVersion: Version = Version("2.0.1")
   val scalafmtGroupId = "org.scalameta"
-  val scalafmtArtifactId = "scalafmt"
+  val scalafmtArtifactId = "scalafmt-core"
```
exoego: Note)
`"org.scalameta" %% "scalafmt"` is a dummy artifact and not exist.
To fetch the project URL for scalfmt, it should be real artifact `scalafmt-core`.

fthomas: What happens if a project actually depends on `scalafmt-core`? I think Scala Steward will only update the version number in `.scalafmt.conf` then.

exoego: Correct.
I think Steward updates both `.scalafmt.conf` and dependencies at same time.
Fixed in https://github.com/fthomas/scala-steward/pull/820/commits/c2341c7d034dea563bc0f6f638ddd7773ad1f4f5
And example of PR is https://github.com/exoego/scala-steward-test-project/pull/34/files

fthomas:
@exoego What do you think about using the same mechanism as in #827 also for scalafmt? I assume the scalafmt version in `.scalafmt.conf` is the version of the `scalafmt-core` artifact which is downloaded by `sbt-scalafmt`. It would be safe then to bump the version in `.scalafmt.conf` if a new version of `scalafmt-core` is available.

exoego:
I think #827 is better design.
Open a new PR based on that.

PR #827, opened 2019-08-14
Add sbt temporarily as dependency

fthomas:
This adds sbt temporarily as dependency during `getDependencies` and
`getUpdatesForRepo`. This has two advantages:

1. We don't need to manually bump sbt versions when a new sbt version
   comes out.
2. Project and diff URLs for sbt are determined by the same mechanisms
   as for all other dependencies and don't need special handling
   (see https://github.com/fthomas/scala-steward/pull/820).

Disadvantages of this approach might be:

1. We're modifying the build which could mean that it might not load
   after we've added the dependency.
2. Dependencies determined by Scala Steward contain "fake" dependecies.
   It is more than what `libraryDependencies` contains.

Here is a PR created by this: https://github.com/fthomas/refined-sjs-example/pull/20


fthomas:
I've merged this now to get a sense of how this works out with the 560+ repositories Scala Steward is looking after.

PR #836, opened 2019-08-15
Add update state UpdateRejectedByConfig

fthomas:
When pruning repos, Scala Steward now takes into account if updates are
rejected by repo configuration. This will prevent Scala Steward from
unnecessarily working on repos where it determined that a dependency is
outdated but it was actually ignored by the repo config.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
@@ -19,9 +19,8 @@ package org.scalasteward.core.update
 import cats.implicits._
 import io.chrisdavenport.log4cats.Logger
 import org.scalasteward.core.data.{Dependency, Update}
-import org.scalasteward.core.git.Sha1
 import org.scalasteward.core.nurture.PullRequestRepository
-import org.scalasteward.core.repocache.RepoCacheRepository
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/org/scalasteward/core/update/UpdateService.scala
```diff
@@ -127,29 +126,34 @@ final class UpdateService[F[_]](
...
-        pullRequestRepo.findPullRequest(repo, dependency, update.nextVersion).map {
-          case None =>
-            DependencyOutdated(dependency, update)
-          case Some((uri, _, state)) if state === Closed =>
-            PullRequestClosed(dependency, update, uri)
-          case Some((uri, baseSha1, _)) if baseSha1 === sha1 =>
-            PullRequestUpToDate(dependency, update, uri)
-          case Some((uri, _, _)) =>
-            PullRequestOutdated(dependency, update, uri)
+        repoCache.maybeRepoConfig.map(_.updates.keep(update)) match {
```
exoego: Can Steward update a dependency again if the dependency is removed from ignore list in repo config?

fthomas: Yes. If the repo config does not ignore it any longer `_.updates.keep` will return a `Right(_)` there.

PR #842, opened 2019-08-16
Improve scalafmt update detection and .scalafmt.conf edit

exoego:
Closes #819. 
Suepecedes PR #820 based on same strategy inPR #827 

This PR also simplify how to edit `.scalafmt.conf` in 2nd comit https://github.com/fthomas/scala-steward/commit/1cd67f8ab22449381128a7539e89431af13bd3a1 and 3rd commit https://github.com/fthomas/scala-steward/pull/842/commits/3ea2de6f7167798d8b8c7d33c098f2ee7eb5b2d9
It introduces `Update`-specific `UpdateHeuristic` so `ScalafmtAlg.editScalafmtConf` and special handling in `EditAlg` are no longer required.

PR #845, opened 2019-08-17
Remove scalaBinaryVersion from temporary scalafmt dependency.

exoego:
Follow-up of #842 

https://github.com/fthomas/scala-steward/pull/842#issuecomment-522218941
>[info]   org.scalameta:scalafmt-core_2.12 : 2.0.0 -> 2.0.1

Temporary `scalafmt-core` dependency's artifact id is `scalafmt-core_2.12` (including scalaBinaryVersion) due to https://github.com/fthomas/scala-steward/blob/1718dc26c08df42caf8376bc761c92965218a25f/modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala#L178-L180
Therefore `UpdateHeuristic` for `scalafmt-core` does not work.

This PR makes the artifact id scalaBinaryVersion-agnostic for `scalafmt-core`.
So future updates for any Scala binary version (2.13, 2.14 or 3.0) should work.

Example PR
https://github.com/exoego/scala-steward-test-project/pull/77

### DIFF
--- a/modules/core/src/test/scala/org/scalasteward/core/edit/UpdateHeuristicTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/edit/UpdateHeuristicTest.scala
@@ -243,6 +243,26 @@ class UpdateHeuristicTest extends FunSuite with Matchers {
       .replaceVersionIn(original) shouldBe (Some(expected) -> UpdateHeuristic.groupId.name)
   }
 
+  test("specific to scalafmt: should be Scala version agnostic") {
+    Seq(
+      ("""version = "2.0.0" """, """version = "2.0.1" """),
+      ("""version="2.0.0"""", """version="2.0.1""""),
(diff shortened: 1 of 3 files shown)


modules/core/src/test/scala/org/scalasteward/core/edit/UpdateHeuristicTest.scala
```diff
@@ -243,6 +243,26 @@ class UpdateHeuristicTest extends FunSuite with Matchers {
       .replaceVersionIn(original) shouldBe (Some(expected) -> UpdateHeuristic.groupId.name)
   }
 
+  test("specific to scalafmt: should be Scala version agnostic") {
```
exoego: Actually, those tests are needed for this fix.

Added just to ensure that whitespace does not matter https://github.com/fthomas/scala-steward/pull/842#issuecomment-522227235

PR #849, opened 2019-08-17
Encapsulate moduleId for Scala binary version

exoego:
By this refactoring, we can simply use just `formatAsModuleId` for any new dependencies in `withTemporarySbtDependency`.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Dependency.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Dependency.scala
@@ -30,10 +30,10 @@ final case class Dependency(
     configurations: Option[String] = None
 ) {
   def formatAsModuleId: String =
-    s""""$groupId" % "$artifactIdCross" % "$version""""
-
-  def formatAsModuleIdScalaVersionAgnostic: String =
-    s""""$groupId" %% "$artifactId" % "$version""""
+    artifactId match {
+      case `artifactIdCross` => s""""$groupId" % "$artifactIdCross" % "$version""""
+      case _                 => s""""$groupId" %% "$artifactId" % "$version""""
+    }
 
   def toUpdate: Update.Single =
     Update.Single(groupId, artifactId, version, Nel.of(version), configurations)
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/sbt/SbtAlg.scala
```diff
@@ -168,18 +168,15 @@ object SbtAlg {
...
-              _.map(scalafmtDependency(defaultScalaBinaryVersion))
-                .map(_.formatAsModuleIdScalaVersionAgnostic)
-            )
+            .map(_.map(scalafmtDependency(defaultScalaBinaryVersion)))
           fakeDeps = Option(maybeSbtDep.toList ++ maybeScalafmtDep.toList).filter(_.nonEmpty)
           a <- fakeDeps.fold(fa) { dependencies =>
             workspaceAlg.repoDir(repo).flatMap { repoDir =>
               val content = dependencies
-                .map(dep => s"libraryDependencies += ${dep}")
+                .map(dep => s"libraryDependencies += ${dep.formatAsModuleId}")
```
exoego: Note) By this refactoring, we can simpl use `formatAsModuleId` for any new dependencies in `withTemporarySbtDependency`.

modules/core/src/main/scala/org/scalasteward/core/data/Dependency.scala
```diff
@@ -30,10 +30,10 @@ final case class Dependency(
     configurations: Option[String] = None
 ) {
   def formatAsModuleId: String =
-    s""""$groupId" % "$artifactIdCross" % "$version""""
-
-  def formatAsModuleIdScalaVersionAgnostic: String =
-    s""""$groupId" %% "$artifactId" % "$version""""
+    artifactId match {
+      case `artifactIdCross` => s""""$groupId" % "$artifactIdCross" % "$version""""
+      case _                 => s""""$groupId" %% "$artifactId" % "$version""""
```
fthomas: I see two problems here. One is that we assume that the dependency exists for the `scalaVersion` used in `ArtificialProject` which might not be true. And the other is that `%%` might cause problems with Scala.js artifacts.

exoego:
Closing since this aproach seems problematic https://github.com/fthomas/scala-steward/pull/849#discussion_r329756764

PR #877, opened 2019-09-03
fail process if updating a pull request fails

robstoll:
Similar to #722 but this time we do not try to recover from a failure during the update process but kind of fail fast. However, what we keep is that the process as a whole fails if there is a failure during the update process


exoego:
> this time we do not try to recover from a failure during the update process but kind of fail fast

What is a motivation for fail-fast?
Personally, I prefer fail-**last**, in which process keep active even if updates on some repos fail, and report each errors.
Because I think, even if some repos may have timeout or some failure, the rest of repos may not and can be updated.

robstoll:
I see the confusion. The current behaviour is fail-last per repository and fail-fast per PR within a repo. This PR does not change that but makes sure a failure in the update proccess of a PR fails the whole proccess at the end. Meaning other repos are still processed.

892 on the other hand introduced fail-last on a PR level

PR #878, opened 2019-09-03
do not force push as it might overwrite other changes

robstoll:


fthomas:
Thanks @robstoll! I'll merge when you update the failing test.

Using `--force` should not have been necessary since https://github.com/fthomas/scala-steward/pull/757.

PR #893, opened 2019-09-05
Use http4s-async-http-client

### LINKED ISSUE #888: Use HTTP client that support proxy settings
I'm currently experiencing a problem with Scala Steward behind a corporate web proxy. It tries to create a PR and fails with a `java.net.ConnectException: Connection refused` when it tries to check if the URL for the version diff exists via the `HttpExistenceClient`. Using standard [Java options for proxy settings](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html) have no effect and I believe this due https://github.com/http4s/http4s/issues/251. So instead of using http4s-blaze-client, I think we should switch to a http4s-client that supports the Java proxy settings.

fthomas:
This needs more manual tests. I just wanted to know what code changes would be required to replace the Blaze client.

Closes #888

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -39,7 +39,7 @@ lazy val core = myCrossProject("core")
       Dependencies.coursierCore,
       Dependencies.coursierCatsInterop,
       Dependencies.fs2Core,
-      Dependencies.http4sBlazeClient,
+      Dependencies.http4sAsyncHttpClient,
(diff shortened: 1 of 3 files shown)


fthomas:
Manual tests were looking good. I'm going ahead and merge this now.

PR #903, opened 2019-09-07
projectId in repo for gitlab

cesarcolle:
## Description
gitlab community edition, for some version, can't recognise a project through the encoded path, there is a way only for gitlab stewarding  to use directly the project ID.

## Note
I also remove the 'final' mark of some classes because I cannot mock them with mockito so it's a ... @ VisibleForTesting stuff I guess. If it's too dangerous I can remove the test itself or work in a way to mock all the Alg of the project.

## behaviour
now for a gitlab steward : 
a little change on the model : 

```
final case class Repo(
    owner: String,
    repo: String,
    pid : Option[Long] = None
) 
```
and so the new parsing provide : 

```  test("the StewardAlg read the repos.md file like ") {
    val f = File.temp / "repos_test.md"
    val repos = algTest
      .readRepos(f)
      .runA(MockState.empty.add(f, "- HCScorps/baton\n- HCScorps/SOA/pid:344"))
      .unsafeRunSync()
   
    repos shouldEqual List(Repo("HCScorps", "baton"), Repo("HCScorps", "SOA", Some(344)))
  }
```

Thanks

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,7 +51,8 @@ lazy val core = myCrossProject("core")
       Dependencies.http4sDsl % Test,
       Dependencies.refinedScalacheck % Test,
       Dependencies.scalacheck % Test,
-      Dependencies.scalaTest % Test
+      Dependencies.scalaTest % Test,
+      Dependencies.mockito % Test
     ),
     assembly / test := {},
     assemblyMergeStrategy in assembly := {
--- a/modules/core/src/main/scala/org/scalasteward/core/application/StewardAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/StewardAlg.scala
@@ -51,7 +51,7 @@ final class StewardAlg[F[_]](
   def readRepos(reposFile: File): F[List[Repo]] =
     fileAlg.readFile(reposFile).map { maybeContent =>
       val regex = """-\s+(.+)/([^/]+)""".r
-      val regexWithProjectId = """-\s+(.+)/([^/]+)/pid:([0-9]+)""".r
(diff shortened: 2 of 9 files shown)


fthomas:
Sorry @cesarcolle for not responding earlier. I'm a bit wary of adding workarounds like this because of the additional maintenance burden. At $WORK we're also using Scala Steward with GitLab CE and it works fine for us. Do you know which GitLab versions are affected and if it is already reported in their issue tracker?

cesarcolle:
Hey, thanks for your answer.

Didn't fine any issue related to this on the gitlab's issue tracker. 
But, we have our own gitlab CE instance, I create a public project to test. 

Gitlab version :  1.12.6
https://gitlab.linkfluence.com/api/projects/linkfluence%2Ftest
which is : https://gitlab.linkfluence.com/linkfluence/test (a public test repo). 

And return 404. It works fine for : https://gitlab.linkfluence.com/api/v4/projects/919 with the direct project's id. 

I understand that's very peculiar functionality but we add more feature to scala-steward at my company and I test all the other features with pid, so I'm kind of stuck to publish other PR (I don't want to publish corrupt stuff).  

Thanks for your comment ! 

The project rocks !

fthomas:
What I find peculiar is the difference between the responses of
https://gitlab.linkfluence.com/api/v4/projects/linkfluence%2Ftest and 
https://gitlab.linkfluence.com/api/v4/projects/linkfluence/test

The first one is a 404 page from Apache while the second is 404 JSON response from nginx. Could this be some kind of configuration error?

mzuehlke:
@cesarcolle has this issue been fixed with #1315 ?
Than this PR could be closed.

Thanks

Issue #839, opened 2019-08-16
Improve "link to diff" detection for uncommon git tags


exoego:
Commonly, releases are tagged as `v1.0.0` or just `1.0.0`.
But some repositories tags releases uncommon prefix.
Example is `util-19.8.1` in https://github.com/twitter/util/releases/tag/util-19.8.1

Therefore, Scala Steward fail to link to diff URL in
https://github.com/travisbrown/catbird/pull/141

Instead of current naive implementation #814, we may improve detection by using [Releases API](https://developer.github.com/v3/repos/releases/) to search releases by version number.

PR #907, opened 2019-09-09
Add link to release note or changelog.

### LINKED ISSUE #905: Include a link to the release notes of the updated depedency
#762 includes a link to the SCM or homepage of the updated dependency and #814 includes a link to the diff between the current and updated version of the dependency. A further improvement would be to link to the GitHub release page of the updated version if the dependency is hosted on GitHub and the page exists. Here is an example for circe: https://github.com/circe/circe/releases/tag/v0.12.0-RC4. As in #814 we would only include this link if the URL exists. Maybe similar release pages also exist for projects hosted on other platforms.

This is in response to @dwijnand, who [noted on Twitter](https://twitter.com/dwijnand/status/1170604859902562304) that:
>one "downside" to @ScalaSteward is that release notes aren't read and applied

exoego:
Closes #905

Example PR: https://github.com/exoego/scala-steward-test-project/pull/94

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/vcs/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/vcs/package.scala
@@ -79,6 +79,6 @@ package object vcs {
       name <- List("CHANGELOG", "Changelog", "changelog", "RELEASES", "Releases", "releases")
       ext <- List("md", "markdown", "rst")
     } yield s"${canonicalized}/${name}.${ext}"
-    vcsSpecific ++ files
+    files ++ vcsSpecific
   }
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/vcs/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/vcs/package.scala
@@ -64,4 +64,21 @@ package object vcs {
     else
       List.empty
   }
+
+  def possibleReleaseNoteFiles(repoUrl: String, update: Update): List[String] = {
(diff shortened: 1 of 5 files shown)


modules/core/src/main/scala/org/scalasteward/core/vcs/package.scala
```diff
@@ -64,4 +64,21 @@ package object vcs {
     else
       List.empty
   }
+
+  def possibleReleaseNoteFiles(repoUrl: String, update: Update): List[String] = {
+    val canonicalized = repoUrl.replaceAll("/$", "")
+    val vcsSpecific =
+      if (repoUrl.startsWith("https://github.com/"))
```
exoego: Gitlab offers "Releases" page.
But unlike GitHub, it have no permalink for specific version, and hard to detect since it returns `200` even Releases are empty (not used).

Example
* Releases https://gitlab.com/gitlab-org/gitlab-runner/-/releases
* Empty releases https://gitlab.com/gitlab-org/gitlab-ce/-/releases

Maybe use GitLab's Releases API to detect release (Ref. https://github.com/fthomas/scala-steward/issues/839)

fthomas: I agree that what is in this PR is good enough. Using the GitLab API to find out if a project uses GitLab's releases feature seems to be overkill to me.

fthomas:
It seems that some PRs include links to non-existent pages:
* https://github.com/balhoff/arachne/pull/17
* https://github.com/balhoff/arachne/pull/19

PR #916, opened 2019-09-09
Do not include non-existent release note/changelog

### LINKED ISSUE #915: Non-existent release note is shown in PR description
Link to release note/change log was introduced in PR #907.

It needs some refinment since the below false-positive links are shown.

This one returns `404` so easy to address.
https://scalaz.github.io/CHANGELOG.md in https://github.com/balhoff/arachne/pull/19
image

This one returns `200`, so a bit trickier 😦 
image
http://groovy-lang.org/CHANGELOG.md in https://github.com/balhoff/arachne/pull/17

exoego:
Closes #915 

`HttpExistenceClient` should return `true` only if http response status is 200.
This is necessary since some servers return 302 and then 200 for 404.html.

Example PR:
* scalaz (404) have no link https://github.com/exoego/scala-steward-test-project/pull/98
* groovy (302->200) have no link https://github.com/exoego/scala-steward-test-project/pull/97
* minitest (200) have link to GitHub release https://github.com/exoego/scala-steward-test-project/pull/99

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/util/HttpExistenceClient.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/util/HttpExistenceClient.scala
@@ -19,7 +19,7 @@ package org.scalasteward.core.util
 import cats.implicits._
 import io.chrisdavenport.log4cats.Logger
 import org.http4s.client.Client
-import org.http4s.{Method, Request, Uri}
+import org.http4s.{Method, Request, Status, Uri}
 
 final class HttpExistenceClient[F[_]](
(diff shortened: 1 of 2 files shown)


eugeniyk:
After merging #1563 found out that release notes are still not added in PR description
Reason - github on-prem returns "302" (redirects to github.com internally I assume?) even when the file exists (in case when file doesn't exist it returns 302 as well)

@fthomas I'd say `HttpExistenceClient` should supports redirects? Otherwise on-prem users won't get release notes feature

PR #922, opened 2019-09-10
Circe 0.12.0 was a bad release

travisbrown:


modules/core/src/main/scala/org/scalasteward/core/update/FilterAlg.scala
```diff
@@ -133,6 +133,11 @@ object FilterAlg {
           // https://github.com/lightbend/migration-manager/pull/260
           "3.2.0-SNAP10"
         )
+      case ("io.circe", _, _, _) =>
```
fthomas: This will mark all circe 0.12.0 modules as bad. Even those that don't live in the `circe/circe` repo.

travisbrown: Hmm, yeah, I guess that's not ideal. I'll scope it down.

fthomas:
Let's just close this. Scala Steward won't ever propose an update to 0.12.0 again because 0.12.1 already exists and it will always prefer the latter.

travisbrown:
Oh, yes, sorry I forgot about this. Thanks.

PR #975, opened 2019-09-19
Update repos.md

jisantuc:
👋


fthomas:
Hi @jisantuc, there is no sbt build in the root directory of that repo and unfortunately Scala Steward currently only works if the sbt build is in the root directory. :-(

jisantuc:
Ah yeah I worried that might be the case. :man_shrugging: Thanks for confirming

PR #1031, opened 2019-10-10
Delete update branch when syncing a fork

### LINKED ISSUE #741: Branch named "update" breaks pull requests
Git doesn't allow branches named `update` and `update/something` in the same repository. If a repository has an `update` branch when `scala-steward` forks the repo, the same `update` branch is present in the fork and it blocks `update/dependency-version` branch creation.

The simplest workaround could be to change the branch prefix to something less common than just `update`. Alternatively, `scala-steward` can clean up all branches from the fork to avoid such conflicts.

fthomas:
Closes #741

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/vcs/VCSRepoAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/vcs/VCSRepoAlg.scala
@@ -19,7 +19,7 @@ package org.scalasteward.core.vcs
 import cats.implicits._
 import org.http4s.Uri
 import org.scalasteward.core.application.Config
-import org.scalasteward.core.git.GitAlg
+import org.scalasteward.core.git._
 import org.scalasteward.core.util
 import org.scalasteward.core.util.MonadThrowable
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/org/scalasteward/core/vcs/VCSRepoAlg.scala
```diff
@@ -45,10 +45,19 @@ object VCSRepoAlg {
...
         }
 
       val withLogin: Uri => Uri =
         util.uri.withUserInfo.set(config.vcsLogin)
+
+      def maybeDeleteUpdateBranch(repo: Repo): F[Unit] = {
+        val branch = Branch(updateBranchPrefix)
+        for {
+          _ <- gitAlg.deleteLocalBranch(repo, branch).attempt
+          _ <- gitAlg.deleteRemoteBranch(repo, branch).attempt
```
fthomas: Instead of `attempt`ing to delete these branches, it would be better to check if theses branches exists before deleting them and then not use `attempt` at all.

fthomas:
Closing since I currently don't have time to finish this PR.

PR #1044, opened 2019-10-16
Resolve POMs of sbt plugins

fthomas:
This adds `sbtVersion` and `scalaVersion` attributes to
`coursier.Dependency` to correctly resolve POMs of sbt plugins. With
this change, Scala Steward shows the SCM URL or homepage in the PR body
of sbt plugins that are on Maven Central.

Here is a PR of sbt-sonatype before this change:
https://github.com/fthomas/refined-sjs-example/pull/57

and here with this change:
https://github.com/fthomas/refined-sjs-example/pull/63

### DIFF
--- a/modules/core/src/main/resources/StewardPlugin.scala
+++ b/modules/core/src/main/resources/StewardPlugin.scala
@@ -48,6 +48,7 @@ object StewardPlugin extends AutoPlugin {
             "version" -> moduleId.revision
           ) ++
             moduleId.extraAttributes.get("e:sbtVersion").map("sbtVersion" -> _).toList ++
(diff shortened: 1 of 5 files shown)


fthomas:
I should note that this won't work for plugins in the https://bintray.com/sbt/sbt-plugin-releases/ repository because that is an Ivy repository and not a Maven repository and we currently only extract URLs from `pom.xml` files and not from `ivy.xml` files.

fthomas:
https://github.com/fthomas/scala-steward/issues/1090 discusses support for sbt plugins in the sbt-plugin-releases repository.

PR #1049, opened 2019-10-18
add heuristic taking into account org type for short group ids

manuelcueto:
As discussed in Gitter. 
because GroupId heuristic filters search terms containing 3 or less characters, organisations like Sky, BBc, etc. get dropped and not updated (unless using variables for their versions).


fthomas:
LGTM, many thanks @manuelcueto!

PR #1050, opened 2019-10-18
add support for limiting updates through config

### LINKED ISSUE #1046: Limit number of updates
Is there any way to limit the number of updates that happen?
In my use case, for each update we have 2 builds (push and PR). For N updates, that would result in 2N builds. I would like to be able to configure a limit so that I can have 1-2 updates for a run. Reasons are:
- This prevents flooding of our CI
- We run Steward daily, so by limiting the updates we can amortize the effort that goes into maintaining the repos.
- When build.sbt gets updated (following a merge to master), we need to resolve conflict and then merge the resolution. So that's another 2 builds. And we have to repeat this for each PR. Therefore I conclude that it's not worthwhile to have N updates in a day, I'd rather spread this over several days.

manuelcueto:
as @exoego pointed out, 
closes #1046 
Adds support in repoConfig for limiting the amount of Updates.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/ProcessResult.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/ProcessResult.scala
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2018-2019 Scala Steward contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.scalasteward.core.data
+
+sealed trait ProcessResult extends Product with Serializable
+
+object ProcessResult {
+  case object Ignored extends ProcessResult
+  case object Updated extends ProcessResult
+}
--- a/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
@@ -21,7 +21,8 @@ import cats.implicits._
 import io.chrisdavenport.log4cats.Logger
 import org.scalasteward.core.application.Config
 import org.scalasteward.core.coursier.CoursierAlg
-import org.scalasteward.core.data.{Dependency, Update}
+import org.scalasteward.core.data.ProcessResult.{Ignored, Updated}
+import org.scalasteward.core.data.{Dependency, ProcessResult, Update}
 import org.scalasteward.core.edit.EditAlg
 import org.scalasteward.core.git.{Branch, GitAlg}
 import org.scalasteward.core.repoconfig.RepoConfigAlg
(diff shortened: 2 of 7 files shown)


exoego:
Closes #1046

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -76,12 +76,13 @@ final class NurtureAlg[F[_]](
       updates <- sbtAlg.getUpdatesForRepo(repo)
       filtered <- filterAlg.localFilterMany(repoConfig, updates)
       grouped = Update.group(filtered)
-      _ <- logger.info(util.logger.showUpdates(grouped))
+      limited = repoConfig.updatesLimit.fold(grouped)(grouped.take)
```
fthomas: @manuelcueto I think the limit is applied too early here. `limited` could contain an update for which a PR exists that was closed without merging. IMO the limit should only take updates into account that create or update a PR.

manuelcueto: Agree. I thought create or update were the only choices so it seemed valid here. I'll take a look.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
```diff
@@ -22,7 +22,8 @@ import io.circe.{Decoder, Encoder}
 
 final case class RepoConfig(
     updates: UpdatesConfig = UpdatesConfig(),
-    updatePullRequests: Boolean = true
+    updatePullRequests: Boolean = true,
+    updatesLimit: Option[Int] = None
```
fthomas: Can this be moved to `UpdatesConfig` as `limit` so that it reads as `updates.limit` in `.scala-steward.conf`?

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
fthomas: If `processUpdate` would return an enum that describes what it has done (created a PR, updated a PR, nothing) instead of `Unit`, it would be easier to apply the limit here, I guess.

manuelcueto: Got to a similar solution before reading this 😅

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -81,40 +82,38 @@ final class NurtureAlg[F[_]](
...
-          logger.info(s"Found PR ${pr.html_url}") >> updatePullRequest(data)
+          logger.info(s"Found PR ${pr.html_url}") >> updatePullRequest(data) >> F
+            .pure[ProcessResult](Updated)
         case Some(pr) =>
-          logger.info(s"Found PR ${pr.html_url}, but updates are disabled by flag")
+          logger.info(s"Found PR ${pr.html_url}, but updates are disabled by flag") >> F
+            .pure[ProcessResult](Ignored)
         case None =>
-          applyNewUpdate(data, getDependencies)
+          applyNewUpdate(data, getDependencies) >> F.pure[ProcessResult](Updated)
```
fthomas: `applyNewUpdate` and `updatePullRequest` do not always result in creating or updating a PR. For example if Scala Steward does not find the version number to bump, `applyNewUpdate` will do nothing. And `updatePullRequest` will only push to the PR branch if there are conflicts with the base branch and no one else committed to it. To take this into account, both methods should return a `ProcessResult` themself.

modules/core/src/main/scala/org/scalasteward/core/data/ProcessResult.scala
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.scalasteward.core.data
+
+sealed trait ProcessResult extends Product with Serializable
+
+case object Ignored extends ProcessResult
+case object Updated extends ProcessResult
```
fthomas: Minor nitpick: Could you put these two objects into a new `ProcessResult` companion? I prefer if these names do not pollute the package namespace.

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -210,14 +209,35 @@ final class NurtureAlg[F[_]](
...
+          previous >>= {
+            case None                => updateF(update).map(_ => None)
+            case prev @ Some((0, _)) => F.pure(prev)
+            case Some((remaining, Ignored)) =>
+              updateF(update).map(result => (remaining -> result).some)
+            case Some((remaining, Updated)) =>
+              updateF(update).map(result => (remaining - 1 -> result).some)
+          }
+      } >> F.unit
+    }
```
fthomas: I imagine this could be easier implemented with a `fs2.Stream`. Here is my idea:
 - `Stream.emits(updates)`
 - `evalMap` the stream with `updateF`
 - map `Updated` to 1 and `Ignored` to 0
 - transform the stream to the running sum of the 1s and 0s (with `scan(0)(_ + _)` or `scanMonoid`)
 - `takeWhile` the running sum is smaller than `updatesLimit`

What do you think?

manuelcueto: Sounds good.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
```diff
@@ -40,9 +40,14 @@ final class RepoConfigAlg[F[_]](
...
-        case Right(config)  => logger.info(s"Parsed $config") >> F.pure(config.some)
-        case Left(errorMsg) => logger.info(errorMsg).as(none[RepoConfig])
+        case Right(config) =>
+          println(config)
+          logger.info(s"Parsed $config") >> F.pure(config.some)
+        case Left(errorMsg) =>
+          println(errorMsg)
+          logger.info(errorMsg).as(none[RepoConfig])
       }
+
```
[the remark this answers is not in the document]
fthomas: This comment is actually about the `println`s a few lines above this blank line.

PR #1057, opened 2019-10-22
External scalafix-migrations file.

### LINKED ISSUE #1043: Flag to Disable Scalafix in Execution
Some of the scalafix migrations take quite a while in the pipeline when there are many repos. For some cases we don't need them to be ran, we just need to check dependencies and wait for the test suite to let us know about the behavior, e.g. when testing locally.

Perhaps adding a flag to when steward is to be run, e.g. `--scalafix-disable`, would be helpful.

manuelcueto:
closes #1043 .

This approach considers loading a list of migrations per repo. Which would mean after this change every repo that was previously relying on existing migration rules would have to make changes on their repositories.


fthomas:
Thanks @manuelcueto for working on this!

One benefit of the current Scalafix integration in my opinion is that maintainers benefit from migrations added to Scala Steward without being aware that they exist. I don't want to lose this because I think this feature has a lot more potential than we are currently using and I like to extend it in the future.

I think the available Scalafix migrations should be configurable on a per-instance and not a per-repo basis. Scala Steward could ship a default list of migrations. If you want to change this list of migrations for your own instance, you could supply it a file that lets you append your own Scalafix migrations to the default migrations or completely overwrite the default migrations with your own list of migrations (which could be empty to disable Scalafix). I imagine this append/overwrite behaviour would be similar to what sbt allows with the `++=` and `:=` operators but I've no idea how this could be implemented. I guess the simplest solution would be a boolean flag like `excludeDefaultMigrations = true`.

WDYT?

manuelcueto:
Yeah that's the other approach i thought of (per instance file.) Hopefully most of the work i did can be reused for that approach :)
the file to be read could be something like: 
```scala
final case class ScalafixMigrations(disableDefaults: Boolean, extraMigrations: List[Migration])
``` 

If you're happy with something like this i'll start working on it

modules/core/src/main/scala/org/scalasteward/core/scalafix/Migration.scala
```diff
@@ -17,12 +17,54 @@
...
 
 final case class Migration(
     groupId: GroupId,
     artifactIds: Nel[Regex],
     newVersion: Version,
     rewriteRules: Nel[String]
-)
+) {
+
+  override def hashCode: Int = Hash[Migration].hash(this)
```
fthomas: @manuelcueto It isn't obvious to me why you override `hashCode`, `equals`, and why there is a `Hash` instance. Where is this used?

modules/core/src/main/scala/org/scalasteward/core/scalafix/MigrationAlg.scala
```diff
@@ -0,0 +1,50 @@
...
+
+import better.files.File
+import org.scalasteward.core.io.FileAlg
+import cats.implicits._
+import cats.Monad
+import io.circe.parser._
+import io.chrisdavenport.log4cats.Logger
+
+trait MigrationAlg[F[_]] {
+  def loadMigrations(migrationsFile: Option[File]): F[List[Migration]]
```
fthomas: @manuelcueto This function exposes to the rest of the program that there can be an optional file that contains a list of migrations. I think this is a detail which could be hidden inside the implementation of `MigrationAlg`. The rest of the program is more interested about available migrations for an `Update`, so a method like `def findMigrations(update: Update): F[List[Migration]]` would be more suitable here. WDYT?

manuelcueto: Makes sense.

modules/core/src/main/scala/org/scalasteward/core/scalafix/Migration.scala
manuelcueto: migration's default equals was failing for 2 migrations with the same stuff. This is due to 2 regex instances being not equal even if they were formed with the same string. I brought this in order to use a custom equals for Regex. But ideally i wouldn't want to do this at all. Do you have any suggestions?

fthomas: Would it make things simpler if we change the type of `artifactIds: Nel[Regex]` to `Nel[String]` and convert them to regexes later when we're trying to find suitable migrations for an update?

manuelcueto: Seems sensible

PR #1067, opened 2019-10-24
add sorting of updates so migration ones run last

### LINKED ISSUE #1065: Apply updates with Scalafix migration last
Updates that involve Scalafix migrations sometimes fail. If there are multiple updates for one repo and the first one Scala Steward applies fails because a Scalafix migration fails, other updates are not applied. To apply as much updates as possible, updates that involve Scalafix migrations should be applied last.

manuelcueto:
closes #1065


fthomas:
LGTM. Many thanks, @manuelcueto!

fthomas:
This is a small change but it already resulted in over a dozen new PRs by @scala-steward that were previously blocked by failing Scalafix migrations. So thanks again @manuelcueto!

PR #1082, opened 2019-11-07
default to hompeage on coursierAlg if url starts with git@

### LINKED ISSUE #1071: Do not use non-browsable URLs as links in the PR body
As happened in https://github.com/mfirry/scala-http-clients/pull/76 with SCM URL `git@github.com/playframework/play-ws.git`. We're extracting this URL from the [POM](https://repo1.maven.org/maven2/com/typesafe/play/play-ws-standalone-json_2.12/2.1.0-M7/play-ws-standalone-json_2.12-2.1.0-M7.pom). I'd suggest to fall back to the homepage if the scm.url is not suitable as link.

manuelcueto:
closes #1071


fthomas:
LGTM, many thanks @manuelcueto!

PR #1087, opened 2019-11-11
Add Hellosoda to the list of users

vaslabs:
Thank you for this project, it's a breeze and very simple to setup.
Saves a lot of time and effort.


fthomas:
Many thanks, @vaslabs!

Issue #678, opened 2019-07-09
Support monorepos with scala subfolders


pvillega:
We currently have a project in which the scala project is located in a subfolder in the repo. Currently Steward can't identify that project as a Scala project, and it ignores it.

We would need a way to indicate a subfolder in a git repository where the scala project is located.

We are aware that there is a closed issue regarding projects following the Play/Akka structure, but this is a different issue.

exoego:
Similar to https://github.com/fthomas/scala-steward/issues/46

pvillega:
Yes, but not exactly the same as #46 if I understand correctly. #46 seems to be focussed on Play projects (based on examples), this is more 'any subfolder that can be a full project by itself'. Then again I may be misreading :)

exoego:
Yes, I mean this issue seems to supersede #46, since this seems able to cover #46

exoego:
I am working on this.

ashawley:
Will this also make it possible to support repos that contain [Giter8](http://www.foundweekends.org/giter8/) templates?

Here's an example: https://github.com/scala/scala-seed.g8

ashawley:
> Will this also make it possible to support repos that contain [Giter8](http://www.foundweekends.org/giter8/) templates?

I saw that Scala Steward is at least updating the sbt version of a g8 tempmlate, see https://github.com/scala/scalatest-example.g8/pull/14

kluen:
Hi @exoego, just curious, is this still being worked on?

exoego:
No, I am no longer working on this feature (I closed my WIP PR #721).
Anyone can take over the PR.

FYI, #1093 is another approach for this feature.

kluen:
Thanks for the info, I haven't found those PRs :)

PR #1093, opened 2019-11-15
scan for projects if --project-dirs is defined

### LINKED ISSUE #46: Support nonstandard project layout.

robstoll:
- introduce --project-dirs with glob patterns which allows to search for projects 
- resolves #46 and #678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@
 - Add update state UpdateRejectedByConfig [#836](https://github.com/fthomas/scala-steward/pull/836) by [@fthomas](https://github.com/fthomas)
 - Better handling scm url  [#834](https://github.com/fthomas/scala-steward/pull/834) by [@exoego](https://github.com/exoego)
 - Added --process-timeout option [#832](https://github.com/fthomas/scala-steward/pull/832) by [@tkawachi](https://github.com/tkawachi)
+- Added --project-dirs option [#1093](https://github.com/fthomas/scala-steward/pull/1093) by [Tegonal GmbH](https://github.com/tegonal)
 
 #### Bug Fixes
 
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ Thanks goes to these wonderful people:
 * [Robert Stoll](https://github.com/robstoll)
 * [sullis](https://github.com/sullis)
 * [TATSUNO Yasuhiro](https://github.com/exoego)
+* [Tegonal GmbH](https://github.com/tegonal)
 * [Thomas Heslin](https://github.com/tjheslin1)
 * [Thomas Kaliakos](https://github.com/thomaska)
 * [Yan](https://github.com/yaroot)
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -24,6 +24,7 @@ import org.http4s.Uri
 import org.http4s.syntax.literals._
 import org.scalasteward.core.application.Cli._
 import org.scalasteward.core.util.ApplicativeThrowable
+
 import scala.concurrent.duration._
 
 final class Cli[F[_]](implicit F: ApplicativeThrowable[F]) {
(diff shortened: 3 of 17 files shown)


robstoll:
@fthomas personally I don't like many `case p if ...` statements, but I can switch from 
[if/else](https://github.com/fthomas/scala-steward/pull/1093/files#diff-37e80d344c3c0e98731d5b411d4e050cR78) to case if you prefer it.

fthomas:
Hi @robstoll, many thanks for the PR. Can you provide an overview what this PR does and what the overall goal is? This PR touches a lot of code and it should be easier to review if the intention behind the changes are known.

robstoll:
@fthomas the overall goal is to allow that one can use scala-steward in a project which 
- has the build.sbt not in the root (e.g. in app/build.sbt)
- has multiple sbt projects within the same repo

In this sense it should solve #46 and #678 IMO. The functionality is kind of documented in [Cli](https://github.com/fthomas/scala-steward/pull/1093/files#diff-71fd031647efcd52d63b16a9b7c49eebR39) and in [WorkspaceAlgTest](https://github.com/fthomas/scala-steward/pull/1093/files#diff-3ecdb3d39edb057274aeebd206285b0eR27)

robstoll:
@fthomas any comments/suggestion for improvements on this one?

mzuehlke:
Hi @robstoll
I see the need to support projects which are not following the default layout and as you mention this would solve multiple open issues.

But in my opinion a global setting is not very practical, unless you set it to `**` or you would have to update it often (especially fro the public bot instance). But if you introduce a local flag in `.scala-steward.conf` you don't need all the fancy globbing logic:

```properties
# specify the root(s) of the projects. In this directory a "build.sbt" file is expected
# Default: './' (The root of the repository)
projects.roots = ['/app']
```

robstoll:
Would be fine with me as well but I am not changing things unless I get green lights for idea as such from a collaborator including feedback in case something else needs to be done.

jan-pieter:
A previous PR suggests using `.scala-steward.conf` for this type of configuration: #688. For me, both ways would work because we are self-hosting scala steward. My personal opinion is that this is best done using `.scala-steward.conf` because that would make it configurable for open source repo's without changing the flags in _this_ repository.

robstoll:
Not changing the configuration of this repository makes a lot of sense IMO. So let's wait what @fthomas  thinks about the general idea and feedback

robstoll:
@fthomas would you mind to give feedback about the general idea of supporting non-default project layouts? #1382 deals with one aspect of it as well, so there is definitely some need for it. I would still update this PR to move the config to .scala-steward.conf

alexklibisz:
Hi, are there any updates on this? We're getting a lot of value out of scala-steward and would really like it for a few repos which have sbt projects in sub-directories. Is it a matter of getting the conflicts resolved? If so I can take a pass at it. Or is it a matter of more substantial changes which would take longer to resolve?

albertpastrana:
Hi @robstoll and @fthomas, thanks your work on this issue.

I see this PR has been sitting here for a while, given that we are very interested in this feature, is there anything we can do to get it done?

Happy to help if needed!

robstoll:
@albertpastrana thomas or another contributor with merge rights needs to get active. See my last two posts.
I am using this PR in production for almost a year and it works great.

albertpastrana:
Thanks a lot for the info @robstoll!!

fthomas:
I took a look at this PR and #1382 again now. A lot has changed since November last year - for example, we now support multiple build tools; sbt and scalafmt versions are not explicitly in `RepoCache` anymore; etc. - but I think the goal is now clear to me and there is a solution that would go well with the current implementation.

Until now there is 1:1 relation between a `Repo` and the root directory of sbt, Maven, or Mill builds. But some repos do not contain the build files in the root directory of the repository or contain multiple builds in different directories. So we want to allow a 1:n relation between a `Repo` and build roots.

I would start with adding a new data type for build roots (lets call it `BuildRoot`) which contains a `Repo` and the relative path to a build root in the repository. To keep things as simple as possible we should only need to go from a `Repo` to a `BuildRoot` when we're interacting with the build system so that we can query multiple builds for dependencies but in all other places we ignore from which `BuildRoot` a dependency is coming from.

The entry point for interactions with build systems is `BuildToolDispatcher`. I think most changes need to happen here. `BuildToolDispatcher` could use `RepoConfigAlg` to get the `RepoConfig`, which need to have a new config for build roots, and construct the user-specified `BuildRoot`s. For each `BuildRoot` it then does what it is currently doing for the root directory and collects the results.

`BuildToolDispatcher` and the build tool traits like `SbtAlg` inherit from `BuildToolAlg` which uses `Repo` for its methods. Since `SbtAlg`, `MavenAlg`, and `MillAlg` should now operate on `BuildRoot`s instead of `Repo` but `BuildToolDispatcher` should still use `Repo` we need to add a new type parameter to the `BuildToolAlg`, i.e. `BuildToolAlg[F, R]` where `R` would be either a `Repo` or a `BuildRoot`.

`SbtAlg`, `MavenAlg`, and `MillAlg` then need to be changed to use `BuildRoot` instead of `Repo`. A new `buildRootDir` in `WorkspaceAlg` is probably handy for that. These changes are rather mechanical.

I think these are all necessary changes to support repos with the build in a subdirectory or multiple builds in different subdirectories.

robstoll:
I don't have time to work on this right now, so if some else would like to take over (or start over again), go on. Maybe it makes sense that @bytecodeguru addresses the multiple files aspect in his PR as this one is at least rebased on master. I'll take a look in October again, if there was no action going on here or in #1382 then I'll might consider to do it myself.

PR #1131, opened 2019-12-06
add checkForUpdate Test and use coursier for getting updates

### LINKED ISSUE #1126: Replace SbtAlg.getUpdatesForProject with coursier.Versions()
Coursier recently added an API to list versions of a module, see https://github.com/coursier/coursier/pull/1439. It seems to me that this could be used to replace `SbtAlg.getUpdatesForProject` which we're currently using to find updates of all dependencies Scala Steward knows about in order to decide on which repos it should work on. In `getUpdatesForProject` we're writing modules to an artificial sbt project that we then load with sbt and where we call sbt-update's task to find new versions. It seems to me that the new Coursier API allows us to find new versions without executing sbt. And since executing sbt is really slow, this change could make Scala Steward a lot faster (at least if `--prune-repos=true`).

manuelcueto:
closes #1126 
Would like to hear some input in case i missed something!

WDYT @fthomas

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -21,14 +21,16 @@ import cats.effect._
 import cats.implicits._
 import coursier.interop.cats._
 import coursier.util.StringInterpolators.SafeIvyRepository
-import coursier.{Info, Module, ModuleName, Organization}
+import coursier.{Info, Module, ModuleName, Organization, Versions}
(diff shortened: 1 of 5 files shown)


fthomas:
This is cool, many thanks @manuelcueto! I'm going to merge this into a new topic branch and will try it out locally.

manuelcueto:
@fthomas would be nice if you share your results when you do it :)

fthomas:
Of course!

Feel free to open further PRs against the topic branch if you'd like to modify your code.

fthomas:
I just tested this branch with my repos. `checkForUpdates` now takes 10 seconds instead of 3 minutes! This is great!

Issue #707, opened 2019-07-13
Add config option for pruning repos


fthomas:
Scala Steward can prune the list of repos it will work on to reduce its overall workload (see https://github.com/fthomas/scala-steward/issues/123 and https://github.com/fthomas/scala-steward/issues/319#issuecomment-504208183 for explanations why we're doing this). This pruning can be disabled by commenting [this line in `StewardAlg`](https://github.com/fthomas/scala-steward/blob/7fa884a7ad29bad982456ac97716982ab2194cc7/modules/core/src/main/scala/org/scalasteward/core/application/StewardAlg.scala#L79) and uncommenting the line below. Pruning makes sense when running Scala Steward very frequently and with many repos but not if it is only running once a day / or week with only a few repos. And I guess most private instances of Scala Steward are used this way. The default behaviour can also lead to errors when checking dependencies for updates like https://github.com/fthomas/scala-steward/issues/319. Therefore we should add a new config option to enable pruning repos and disable it by default.

fthomas:
Implemented in #711.

PR #1133, opened 2019-12-06
Use coursier.Versions in checkForUpdate

fthomas:
This contains @manuelcueto's changes from #1131 targeted to master.

We can use this PR for further discussion and changes.


manuelcueto:
It looks good, removes a  lot of code that was a bit hard to follow and wasn't tested. my first commit was just to use it but i thought that it was missing some logic that sbtAlg was doing so that's why i only created a draft. but it seems you got it to a much better shape. And the bump in performance is very exciting. 
My only question is why is this only used for `pruneRepos = true` only?

fthomas:
`checkForUpdates` is only used in the implementation of the repos pruning scheme as described in #707. It is not (yet) a full replacement for using `sbt-updates` since it doesn't use any custom resolvers and credentials that might be defined in the sbt build and it also ignores if an update is available for all `crossScalaVersions`. If we could implement these features and replace sbt-updates, Scala Steward could be blazingly fast.

fthomas:
I'm merging this now because I'm curious how this reduces the runtime of my public Scala Steward instance. I'll report back once the numbers are in. :timer_clock:

Thanks again @manuelcueto!

fthomas:
The numbers are in. Instead of 80 minutes, `checkForUpdates` now takes only 8 minutes! :confetti_ball:

manuelcueto:
😍

Issue #1141, opened 2019-12-08
Allow repo specific resolvers in CoursierAlg


fthomas:
`CoursierAlg` currently looks for `pom.xml`s, `ivy.xml`s, and new versions in Coursier's default package repositories and in sbt-plugin-releases. It won't find any of those information for packages that are in repo specific repositories. As a prerequisite to replacing sbt-updates with `CoursierAlg` and to make it more useful in general, we need to change `CoursierAlg` so that it can take custom package repositories (and accompanying credentials) into account.

Here are references to other issues and PRs which make the case for this feature:
* https://github.com/fthomas/scala-steward/issues/1088#issuecomment-552585756
* https://github.com/fthomas/scala-steward/pull/1092
* https://github.com/fthomas/scala-steward/pull/1133#issuecomment-562848761

manuelcueto:
By custom package repositories you mean private nexus and resolvers?

fthomas:
Yes, exactly. Basically we need the ability to pass the information from `sbt resolvers` to `CoursierAlg`. sbt calls these [`Resolver`](https://www.scala-sbt.org/1.x/api/sbt/librarymanagement/Resolver.html) / [`Repository`](https://www.scala-sbt.org/1.x/api/sbt/librarymanagement/MavenRepository.html) and in Coursier these are just called [Repository](https://get-coursier.io/docs/api.html#repositories).

I think the first step would be a new task in `StewardPlugin` that outputs the resolvers as JSON. We then would need to store them in the `RepoCacheRepository` which we then could query in different places to pass the additional repositories to `CoursierAlg`.

manuelcueto:
Sounds good. I'll take a look later. that would mean we still need to run Sbt though, right?

fthomas:
Yes, but only to keep the cache of dependencies and resolvers fresh. We wouldn't need to run it to find dependency updates.

PR #1144, opened 2019-12-09
Use coursier to get dependencies

### LINKED ISSUE #1139: Find updates of sbt and scalafmt with CoursierAlg
Since #1131 and #1133 we can use `CoursierAlg` to find updates of dependencies. We should also use it to find updates of sbt and scalafmt instead of doing it in `SbtAlg` via `withTemporarySbtDependency`.

manuelcueto:
closes #1139 

What do you think @fthomas ?

Would this `(UpdateAlg.findUpdateUnderNewGroup)` in line 115 have to be reimplemented?

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
@@ -77,8 +77,8 @@ object Context {
       implicit val updateRepository: UpdateRepository[F] =
         new UpdateRepository[F](new JsonKeyValueStore("updates", "3"))
       implicit val coursierAlg: CoursierAlg[F] = CoursierAlg.create
-      implicit val nurtureAlg: NurtureAlg[F] = new NurtureAlg[F]
       implicit val updateAlg: UpdateAlg[F] = new UpdateAlg[F]
+      implicit val nurtureAlg: NurtureAlg[F] = new NurtureAlg[F]
       new StewardAlg[F]
     }
 }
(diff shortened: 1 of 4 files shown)


fthomas:
If we want to get rid of `getUpdatesForRepo`, we need to have #1141 first so that we also find updates of dependencies that are only in private repositories. My intention with #1139 was to only replace `withTemporarySbtDependency` with `UpdateAlg.findUpdate`. `withTemporarySbtDependency` allows us to also get updates of `org.scala-sbt:sbt` and `org.scalameta:scalafmt-core` which are then used to update the sbt version in `project/build.properties` and the scalafmt version in `.scalafmt.conf`. To find updates of these two dependencies, support for custom repositories in `CoursierAlg` is not needed.

manuelcueto:
Ohh i see. I first tried calling findUpdate from SbtAlg but i couldn't create it because it was creating a Circular Dependency.

fthomas:
Oh, I didn't noticed that there would be circular dependency if `findUpdate` is used in `SbtAlg`. Which class is causing it?

manuelcueto:
RepoCacheAlg needs SbtAlg And UpdateAlg needs RepoCacheAlg 
so SbtAlg cannot need UpdateAlg

fthomas:
I think we should split `UpdateAlg` into two classes. One contains `findUpdate`, `findUpdateUnderNewGroup`, and `getNewerGroupId` and the other one the rest which just exists for the "pruning repos" implementation.

fthomas:
Superseded by #1209.

PR #1165, opened 2019-12-13
Add wehkamp to companies using scala-steward

Philippus:


fthomas:
Many thanks @Philippus!

Issue #1156, opened 2019-12-12
Check that updates are available for all crossScalaVersions


fthomas:
`PruningAlg` does not take `crossScalaVersions` into account when deciding on which repos Scala Steward should work on. Currently it only checks for updates of dependencies with a repo's default `scalaVersion`. So if in a repo there is an update of a dependency for `scalaVersion` but not for all other `crossScalaVersions`, `PruningAlg.needsAttention` will return true for this repo. But Scala Steward will not create a PR for this dependency, because the update is not available for all `crossScalaVersions` and thus will not be reported by sbt-updates. I think Scala Steward should copy sbt-updates' behaviour here.

Here is an example. [`fthomas/crjdt`](https://github.com/fthomas/crjdt) currently depends on circe-core 0.11.2, has `scalaVersion` set to `2.12.8` and `crossScalaVersions` set to  `List(2.11.12, 2.12.8)`. The (shortened) output of `stewardDependencies` for `scalaVersion` and for all `crossScalaVersions` in this project is:
```
sbt:crjdt> show circeJVM/stewardDependencies
[info] { "groupId": "io.circe", "artifactId": "circe-core", "artifactIdCross": "circe-core_2.12", "version": "0.11.2", "configurations": null, "sbtVersion": null, "scalaVersion": null }

sbt:crjdt> ++show circeJVM/stewardDependencies
[info] { "groupId": "io.circe", "artifactId": "circe-core", "artifactIdCross": "circe-core_2.11", "version": "0.11.2", "configurations": null, "sbtVersion": null, "scalaVersion": null }
[info] { "groupId": "io.circe", "artifactId": "circe-core", "artifactIdCross": "circe-core_2.12", "version": "0.11.2", "configurations": null, "sbtVersion": null, "scalaVersion": null }
```

A newer version of circe-core exists (currently 0.12.3) but it is only available for Scala 2.12 and not for Scala 2.11. Scala Steward nonetheless does not prune this repo and selects it for further processing:
```
[info] 2019-12-12 09:20:51,685 INFO  Update states for fthomas/crjdt:
[info]   DependencyOutdated(Dependency(io.circe,circe-core,circe-core_2.12,0.11.2,None,None,None),Single(io.circe,circe-core,0.11.2,NonEmptyList(0.12.3),None,None))
```

[...]

PR #1174, opened 2019-12-17
Check updates for all crossScalaVersions

fthomas:
This PR

* adds a new ArtifactId type that contains the name of an artifact as
  used in build files and names with cross version suffixes.
* changes Dependency and Update to use ArtifactId.
* changes SbtAlg to use "+ show stewardDependencies" and
  "+ show stewardUpdates" so we get dependencies and updates for all
  crossScalaVersions
* changes PruningAlg such that it marks a dependency only outdated if
  there are updates for all cross artifact names.

closes: #1156


fthomas:
As expected, PRs like https://github.com/circe/circe-derivation/pull/150 and https://github.com/circe/circe-derivation/pull/151 are a consequence of this change. These update dependencies that are only used for Scala versions that are not `scalaVersion` but which are in `crossScalaVersions`.

PR #1193, opened 2019-12-23
Maven support

stanislav-chetvertkov:


fthomas:
This is a very interesting PR! I'm generally interested in supporting different build systems.

I haven't had time yet to look at it in detail but we're currently trying to replace `SbtAlg.getUpdates` with `UpdateAlg.findUpdates` so that there is no need for the build system to provide updates to Scala Steward. This change will also affect this PR and will make some of the changes herein superfluous and will require that Scala Steward can query custom package repositories from the build system. This feature is currently tracked in #1141 and I think it makes sense for this PR to wait until #1141 is resolved to prevent wasting efforts. WDYT @stanislav-chetvertkov?

stanislav-chetvertkov:
@fthomas Hi, thanks for looking into it
Replacing `SbtAlg.getUpdates` with `UpdateAlg.findUpdates` definitely makes sense because it will produce the same output irrelevant of a build system, unfortunately, the output of `mvn versions:display-dependency-updates `is different from sbt-update because it can only output a simple minor and a major possible version bump but not the whole list.
If I understood correctly `CoursierAlg` will be responsible for parsing both dependencies and updates from .pom so there isn't much left for MavenAlg to do except running code migration but then again this could be achieved by directly calling `scalafix` commands from CLI.
I agree that it would make sense to wait until #1141 is finished, it could make the approach in this PR somewhat obsolete

fthomas:
`CoursierAlg` will be responsible for finding metadata and newer versions of dependencies. `BuildSystemAlg`, `SbtAlg`, and `MavenAlg` will be used get dependencies and resolvers from the project. But we won't need sbt-updates or `mvn versions:display-dependency-updates` to find updates for a project which means adding support for other build systems should become easier. #1209 shows how Scala Steward could look like if it takes care of finding updates itself.

fthomas:
@stanislav-chetvertkov #1209 has been merged in the meanwhile so that Scala Steward itself is now responsible for finding updates. As far as I'm concerned, there is no other change that blocks this PR any more and I hope that it is now easier to add support for Maven. :-)

stanislav-chetvertkov:
👍  @fthomas will get back to it soon

modules/core/src/main/scala/org/scalasteward/core/maven/MavenAlg.scala
```diff
@@ -0,0 +1,209 @@
...
+
+  private val version: Parser[Version] = {
+    val version3args: Parser[Version] =
+      for {
+        a <- int <~ dot
+        b <- int <~ dot
+        c <- int
+      } yield Version(a, b, Some(c))
+
+    val version2args: Parser[Version] =
```
tyrcho: what is the difference with version3args ?

stanislav-chetvertkov: @tyrcho Hi, this bit is not actually used and I just removed it, currently only `override def getDependencies(repo: Repo): F[List[Scope.Dependencies]]` needs to be implemented and `def parseDependencies(lines: List[String]): List[Dependency]` method rewritten with atto (and the output type changed to `List[Scope.Dependencies]`) -  will probably require an output of some other maven commands to additionally get `resolvers: List[Resolver]` 
I've been quite busy at work lately and kind of neglected the PR but I'll try to finally finish the work

fthomas:
Oops, I accidentally closed this by merging #1429. That was not my intention since #1429 doesn't contain the changes for supporting Maven. My intention was to rebase this PR after merging #1429. I probably should have squashed the commits of #1429 instead of merging it. Sorry @stanislav-chetvertkov!

I'll open a new PR that is based on current master with your changes for supporting Maven.

PR #1209, opened 2020-01-02
Scala Steward without sbt-updates

fthomas:
This shows how Scala Steward would work without sbt-updates.


fthomas:
My public Scala Steward instance is now using this branch and I love it. Instead of 2 to 3 hours to check all 900 repos it now only takes 30 to 40 minutes.

mwz:
This is an amazing performance improvement @fthomas!

fthomas:
@mwz Yeah, turns out that not calling sbt repeatedly really helps with performance. ;-)

The other nice consequence of this change is that it will be much easier to add support for other build systems because the only things we're now querying the build system for are dependencies and resolvers.

PR #1226, opened 2020-01-06
1141 custom resolver

### LINKED ISSUE #1141: Allow repo specific resolvers in CoursierAlg

manuelcueto:
closes #1141 
@fthomas  Let me know what you think about it

Sorry for taking so much time to work on it


fthomas:
Thanks @manuelcueto, this already looks quite good. I would tweak it a little bit mostly for performance reasons:

* Resolvers should be cached by `RepoCacheAlg` so that we don't need to depend on `SbtAlg` in `PruningAlg`. Executing sbt takes a lot of time and the less we need to call it, the faster Scala Steward will be.

* Since executing sbt takes a lot of time, we should query dependencies and resolvers at the same time. So `SbtAlg.getDependencies` should become `SbtAlg.getDependenciesAndResolvers` which returns `F[(List[Dependency], List[Resolver])]`. I know this is ugly but calling sbt twice when we can call it only once would make Scala Steward unnecessarily slow.

* I would not pass extra resolvers in `findAdditionalUpdates` since sbt and scalafmt are on Maven Central and #1209 will remove that function anyway.

I think these changes doesn't require too much code changes.

manuelcueto:
Makes sense, will take a look at it later.

fthomas:
FYI, I merged master to resolve merge conflicts due to the merge of #1235.

manuelcueto:
Ohh , thought it was up to date :) no worries! let me know what you think

fthomas:
No worries, it was up to date until I merged the other PR. :-)

Issue #1162, opened 2019-12-13
Make PR update behaviour configurable


kluen:
As of #724, Scala Steward updates its pull requests only when they are in conflict with the base branch. This behaviour helps reducing the number of builds. However, it doesn't allow you to check for semantic conflicts which a successful git merge might have. This use case had been covered before the change.

This issue aims to make Scala Steward configurable so that the user can choose the new or the old (pre-#724) behaviour.

@fthomas also [mentioned some ideas](https://github.com/fthomas/scala-steward/pull/724#issuecomment-561362046) on how this might be implemented.

fthomas:
Fixed by #1227.

PR #1227, opened 2020-01-06
 Change updatePullRequests config to be an enum(always, on-conflicts, never)

kiranbayram:
Addresses #1162

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -22,6 +22,11 @@ updates.limit = 5
 # long as you don't change it yourself.
 # Default: true
 updatePullRequests = true
+
+# If true, Scala Steward will perform a check to see if the PR has conflicts in
+# order to decide to update the PR (assuming that all other conditions are satisfied).
+# Default: true
+checkConflictsWhenUpdatingPRs = true
 ```
 
 ## Ignore lines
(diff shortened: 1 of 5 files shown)


kiranbayram:
I wanted to add some tests for `shouldBeUpdated` function but initializing `NurtureAlg` looked too difficult 😅 . Do you suggest if we extract `shouldBeUpdated` to companion object ?

fthomas:
Thanks @kiranbayram, this looks good! I only find the name of the flag a little bit unwieldy. While thinking about another name, I had the idea of reusing `updatePullRequests` for this. Instead of being a boolean value, it could be an enum with the values `always`, `never`, and `on-conflicts`:
```
always       would be same as updatePullRequests = true and checkConflictsWhenUpdatingPRs = false
on-conflicts would be same as updatePullRequests = true and checkConflictsWhenUpdatingPRs = true
never        would be same as updatePullRequests = false
```
I feel that this is simpler for users than two interacting boolean flags. The implementation would get a little bit more complex, since we should still allow `true` and `false` as values and map them to `on-conflicts` and `never` to keep current `.scala-steward.conf` files valid. WDYT? Does this makes sense and would you be interested in changing this PR in this direction? If not, we can also merge this now and I'll make the changes later.

kiranbayram:
Makes sense. I'll try to change the PR :+1:

fthomas:
Awesome, thanks!

kiranbayram:
Just finished the implementation, hopefully this is not far from what you had in mind @fthomas

fthomas:
This looks great! Many thanks again, @kiranbayram! I'll merge it tomorrow.

mzuehlke:
This closes #1162

fthomas:
Thanks for the reminder, @mzuehlke. I'm closing that now.

PR #1230, opened 2020-01-08
Add new scio 0.8.0 scalafix rule

regadas:


regadas:
Hi @fthomas lmk if I missed smth here 😄

fthomas:
@regadas I think we should add a new migration for scio 0.8.0 instead of changing the one for 0.7.0. Scala Steward will run both if it is updating from a version < 0.7.0 and only run the new one when the current version is >= 0.7.0. Or is there a specific reason for running the old migrations when updating from >= 0.7.0?

regadas:
Hi @fthomas no specific reason ... I just misunderstood things 😄

regadas:
@fthomas thanks!

PR #1237, opened 2020-01-12
Support Ivy repositories

fthomas:
This adds support for Ivy repositories so that we can find new versions of sbt plugins without using any default repositories in `CoursierAlg`. `Resolver` is now an enum with two cases, `MavenRepository` and `IvyRepository`.

It works but results in unnecessary lookups of library dependencies in Ivy repositories and plugin dependencies in Maven repositories. I think we need to bundle libraries and resolvers and add scopes so that library dependencies and plugin dependencies have their own set of resolvers. I'll work on this in this branch.


fthomas:
This branch groups dependencies and resolvers now into `Scope`s. Each `Scope` has its own list of resolvers and `SbtAlg.getDependencies` returns now a list of `Scope`s where library and plugin dependencies are in their own scope since they use different resolvers.

Issue #1088, opened 2019-11-11
Compare url & release notes/changelog are restricted to {github,gitlab}.com


anilkumarmyla:
Current logic of generating compare url and release notes/changelog are restricted to either github.com or gitlab.com
```
if (repoUrl.startsWith("https://github.com/") || repoUrl.startsWith("https://gitlab.com/"))
```

Is it possible to configure/extend this, so that git enterprise urls (such as git.enterprise.company.com) can also be included in this?

fthomas:
Sure. But `repoUrl` needs to determined first and it is currently coming from `CoursierAlg`. If the depedency is private, `CoursierAlg` needs to be made aware of the private artifact repository where it is hosted.

anilkumarmyla:
Yes, coursier reads the repoUrl from the artifact, but if it’s other than github or gitlab, we don’t generate the compare link or point to release.

exoego:
FYI:
I implemented this logic so because URLs other than github/gitlab are unlikely to have release notes/changelog, so I restrict to github/gitlab to avoid costy URL existence checking.

But recent PR https://github.com/fthomas/scala-steward/pull/1081 introduced cache for URL existence checking.
So I think now is reasonable to remove github/gitlab restriction, and add more URL patterns.

anilkumarmyla:
> URLs other than github/gitlab are unlikely to have release notes/changelog

What I meant is that the url regex strictly checks for github.com/gitlab.com while in git enterprise setup these can be git.enterprise.company.com urls

fthomas:
> But repoUrl needs to determined first and it is currently coming from CoursierAlg. If the depedency is private, CoursierAlg needs to be made aware of the private artifact repository where it is hosted.

This is not an issue anymore because `CoursierAlg` uses since https://github.com/scala-steward-org/scala-steward/pull/1226 the same artifact repositories as the build. I think the only functions that need to be changed to support this are the ones in the `vcs` package object.

PR #1238, opened 2020-01-12
Don't use default resolvers for finding artifact URL

fthomas:
The same as #1236 but for `CoursierAlg.getArtifactUrl`.

Prerequisite for #1088.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -34,18 +34,15 @@ import org.scalasteward.core.data.{Dependency, Resolver, Version}
   * fetching dependency versions and metadata.
   */
 trait CoursierAlg[F[_]] {
-  def getArtifactUrl(
-      dependency: Dependency,
-      extraResolvers: List[Resolver] = List.empty
-  ): F[Option[Uri]]
+  def getArtifactUrl(dependency: Dependency, resolvers: List[Resolver]): F[Option[Uri]]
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -90,20 +87,12 @@ object CoursierAlg {
             result <- maybeFetchResult.toOptionT[F]
             moduleVersion = (coursierDependency.module, coursierDependency.version)
             (_, project) <- result.resolution.projectCache.get(moduleVersion).toOptionT[F]
-            url <- getScmUrlOrHomePage(project.info)
-              .toOptionT[F]
-              .orElse(OptionT(getParentArtifactUrl(project)))
+            parentUrl = getParentDependency(project)
```
mzuehlke: Will this trigger the fetching of the parent project metadata also in case where it is not needed ?

fthomas: No, since `parentUrl` is not `flatMap`ped via `<-` in the for-comprehension. And since the code here is referentially transparent `x.orElse(y)` is the same as `val a = y; x.orElse(a)`.

fthomas: One marginal "downside" is that `getParentDependency(project).traverseFilter(getArtifactUrlImpl(_, resolvers))` is now always evaluated (but not executed) even if the parent POM is not needed. I think I'll change that again to make it clearer. Thanks for the nudge, @mzuehlke!

PR #1245, opened 2020-01-17
1173 host prefix

### LINKED ISSUE #1173: Support bitbucket & gitlab on public instance
How about enhancing the `repos.md` syntax to include an optional prefix for the git host on the public `scala-steward` instance?
For example:
```
- mslinn/awslib_scala  # defaults to github
- github:mslinn/dottyTemplate
- bitbucket:mslinn/myProject
- gitlab:mslinn/MyProject
```

manuelcueto:
closes #1173 
Runs all 4 vcs and grabs the vcs type from the repo (defaulting to github) instead of grabbing it from the config.
Couldn't run it locally as i don't have bitbucket set up. If anyone could try it it would be awesome


manuelcueto:
Although i've just realised that now we need to pass vcs login and api host for all 4 vcs types (if i'm not mistaken.) 
@fthomas unless you have any better approach for it?

mzuehlke:
Hello @manuelcueto, 
as you just mentioned yourself we would need to pass 4 different API host and maybe even some different users. One idea that comes to my head is to invoke `scala-steward` multiple times in `scripts/run.sh` and each time pass the information for a different VCS.

The caching build in by @fthomas should help to not have repeated maven central calls.
So you would _only_ need the improved `repos.md` parsing to use only these repos from the currently invoked VCS.
What do you think ?

fthomas:
@mzuehlke raises a good point about the benefits of having one process that works with all different Git repository manager that Scala Steward supports. Before we used Coursier for finding updates, there was an in-memory process for finding updates of all distinct dependencies of all projects Scala Steward worked on to reduce the required sbt and sbt-updates calls. This process would have been slower if it would have been repeated for each repository manager. So at this time there was a benefit of supporting all repository manager within the same process.

With Coursier we're now using its persistent cache for finding updates and this cache can be shared (without effort) with other Scala Steward instances that use a different Git repository manager. So the biggest benefit of supporting all repository manager in one process is gone now.

I'm trying to think of other benefits this change might have, but currently don't see any. And having only one `repos.md` file instead of four is not a big enough benefit that warrant this change IMO.

I'm sorry I haven't articulated this before but I have been busy the last one and a half months working on the Coursier integration.

manuelcueto:
@fthomas @mzuehlke I agree with you guys. Also, seeing the amount of code that needs to be changed plus requiring potentially all vcs login credentials on each run seems like not worthy compared to the benefits.

mzuehlke:
I think this PR can then be closed, as the issue it is referring to (#1173) has been closed.

fthomas:
Right. I've been running another instance for GitLab repositories for some time now and it works like a charm. I currently don't see any advantages we would gain if one process can work with multiple Git repository managers.

PR #1250, opened 2020-01-20
Add the ability to use default reviewers to bitbucket server

andreami:
I added the ability, for bitbucket server only, to use the default reviewers configured in the repository settings.

I just call the default-reviewers API of Bitbucket and add the list of the reviewer to the PR input.

Here the documentation:
https://docs.atlassian.com/bitbucket-server/rest/6.10.0/bitbucket-default-reviewers-rest.html

The feature can be enabled with the flag --use-default-reviewers that by default is set to false.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -53,7 +53,8 @@ object Cli {
       processTimeout: FiniteDuration = 10.minutes,
       scalafixMigrations: Option[String] = None,
       cacheTtl: FiniteDuration = 2.hours,
-      cacheMissDelay: FiniteDuration = 0.milliseconds
+      cacheMissDelay: FiniteDuration = 0.milliseconds,
+      useDefaultReviewers: Boolean = false
   )
 
   final case class EnvVar(name: String, value: String)
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -24,6 +24,7 @@ import org.http4s.Uri
 import org.http4s.syntax.literals._
 import org.scalasteward.core.application.Cli._
 import org.scalasteward.core.util.ApplicativeThrowable
+
 import scala.concurrent.duration._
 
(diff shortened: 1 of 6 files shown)


andreami:
Codacy can't pass the check for a NonEmptyList .head that I just move into a for comprehension.
In my opinion, this should be correct because if the PR has no link references this should raise an error.

fthomas:
@andreami I'm fine with adding this, but it would be nice if the `--use-default-reviewers` flag would indicate that this is a "Bitbucket Server"-only flag. Something along the lines of `--bitbucket-server:use-default-reviewers`. Is it possible with case-app to group flags like `--X:A --X:B` in a `case class X` with fields `A` and `B`?

andreami:
Hi, @fthomas I like your suggestion, I'll try to implement it and I'll check if it is suitable.
Thanks, Andrea

DStranger:
Hi @andreami, do you plan on finishing this? Otherwise I can pick this up. Great feature we are currently missing in our workflow!

andreami:
Hi @DStranger, 
sorry if I left this feature unfinished, I have been very busy in recent months. 
Now for the coronavirus situation, I have some free time after work and I could work on it.

I was working on this feature only on the bitbucket server "branch", therefore if you want we could find a definitive solution for all VCSs.

I have explored @fthomas 's solution and I think it is good, but case-app does not offer that by default, so I thought about opening a PR to implement it.

Do you have a possible solution in mind?

Thanks,
AM

DStranger:
Hi @andreami 
I'm interested in this working for bitbucket server (don't know how (or if) default reviewers work in other VCSs).
As for the flag name, I think something like `--bitbucket-server-use-default-reviewers` would also work, but that's probably not my call :)

fthomas:
I'm also fine with `--bitbucket-server-use-default-reviewers` if case-app does not support grouping options.

Is anyone aware of other command-line parsers that support this?

andreami:
Hi @DStranger @fthomas,

I am not aware of other command-line parsers, so I change the flag in 
`--bitbucket-server-use-default-reviewers` and then rebased my branch with your master and test it on some project.

This PR contains all the modify.

Thanks,
AM

DStranger:
@andreami thanks a lot for this! :)

PR #1254, opened 2020-01-21
should "allow" behave like white list?

jcouyang:
The behavior of "allow" really confuse me, in document:
```properties
# The only dependencies which match the given pattern are updated.
# Each pattern must have `groupId`, and may have `artifactId` and `version`.
# Defaults to empty `[]` which mean Scala Steward will update all dependencies.
updates.allow  = [ { groupId = "com.example" } ]
```
> The only dependencies which match the given pattern are updated.

I thought it should filler out everything else not in the "allow" list

But the current behavior is
When there are 3 artifacts
```scala
    val update1 = Single("org.http4s" % "http4s-dsl" % "0.17.0", Nel.of("0.18.0"))
    val update2 = Single("eu.timepit" % "refined" % "0.8.0", Nel.of("0.8.1"))
    val shouldNotAllow = Single("abc.def" % "ghi" % "0.8.0", Nel.of("0.8.1"))
```
And the rules is
```
        allow = List(
          UpdatePattern(update1.groupId, None, Some("0.17")),
          UpdatePattern(update2.groupId, Some("refined"), Some("0.8"))
        )
```
the artifacts after filter is `update1` and `shouldNotAllow`

but the group id of `sholdNotAllow` is not even listed in the allow list.

while my expectation is aligned with the doc, only `update2` should get update

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -6,13 +6,13 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 # The only dependencies which match the given pattern are updated.
 # Each pattern must have `groupId`, and may have `artifactId` and `version`.
 # Defaults to empty `[]` which mean Scala Steward will update all dependencies.
-updates.include  = [ { groupId = "com.example" } ]
+updates.allow  = [ { groupId = "com.example" } ]
 
 # The dependencies which match the given version pattern are updated.
 # Each pattern must have `groupId`, `version` and optional `artifactId`.
 # Defaults to empty `[]` which mean Scala Steward will update all dependencies.
 # the following example will allow to update foo when version is 1.1.x
-updates.allow  = [ { groupId = "com.example", artifactId="foo", version = "1.1" } ]
+updates.pin  = [ { groupId = "com.example", artifactId="foo", version = "1.1" } ]
 
 # The dependencies which match the given pattern are NOT updated.
 # Each pattern must have `groupId`, and may have `artifactId` and `version`.
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -6,7 +6,13 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 # The only dependencies which match the given pattern are updated.
 # Each pattern must have `groupId`, and may have `artifactId` and `version`.
 # Defaults to empty `[]` which mean Scala Steward will update all dependencies.
-updates.allow  = [ { groupId = "com.example" } ]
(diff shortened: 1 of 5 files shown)


fthomas:
I agree that the documentation for `updates.allow` could be improved but it wasn't designed as whitelist (see https://github.com/fthomas/scala-steward/issues/91). The idea was to restrict updates that match a groupId and artifactId in `updates.allow` to the given version range. In hindsight, it would have probably been better if it would have been called `pin` (as in "version pinning") or something other than `allow`.

jcouyang:
Oh, I get it now, or maybe don't change `allow` for backward compat
but can I request new feature like adding a new config that works like whitelist i.e. `updates.include` or `updates.includeOnly`?
because we have usecase like only need to bump our own organisation packages and whitelist will make our config much easier.

fthomas:
I'm fine with adding a new whitelist feature using `updates.include`.

mzuehlke:
The current PR can lead to confusion by the user, because dependencies that are in `updates.allow`, but not in `updates.included` get ignored.
An alternative idea to having 3 lists with dependencies in the future would be do add a `Boolean` option `updates.allowAsWhitelist`  or `updates.onlyAllowed` (default being `false`) which would switch `updates.allow` into a whitelist.

fthomas:
Good point, @mzuehlke. Another idea would be to append `updates.allow` to `updates.included` if the latter is not empty because I think `updates.included` subsumes the functionality of `updates.allow`. WDYT @jcouyang?

jcouyang:
yeah, if `include` is not empty, it cover the version pin feature too. @fthomas so, are you suggesting skip `allow` list if `include` is provided?

I think it is confused because the name of `allow` and `include` sound similar. @mzuehlke 
I'm not a fan of using true false toggle
since `include` provide whitelist, `ignore` provide blacklist, `allow` provide version pin list, I think simply rename `allow` will avoid such confusing. apart from naming, if the artifact is not even in whiltelist, it make sense to me that the pin version list should not affect anything

fthomas:
My idea is to append the values from the `allow` list to the `include` list if it is not empty. By doing this, entries in the `allow` list are not ignored if they are not also in the `include` list.

jcouyang:
isn’t that more confusing? If artifact is not listed in whitelist but in version allow list, still update the artifact will against the law of whitelist (ignore everything not in the list) then.
Also from users point of view, when they provide a whitelist that also can pin version, no reason to split part of the config into allow.

fthomas:
Okay, here is another proposal. I've just checked how many repos (that my public instance works on) actually make use of `updates.allow` and it turns out nobody is using it. Therefore I'd be comfortable with renaming `updates.allow` to `updates.pin` and `updates.include` to `updates.allow` and leaving the implementation as-is.

PR #1216, opened 2020-01-04
Revert "Revert "Parallelize UpdateAlg.findUpdate""

fthomas:
Reverts fthomas/scala-steward#1215

It seems that checking for updates sequentially didn't help with not being blacklisted by Maven Central. Other services are also seeing 403s from Maven Central (see https://twitter.com/dwijnand/status/1213406224450048001), so maybe it is not entirely Scala Steward's fault.

I'll now trying out the [Maven Central mirror on Google Cloud Storage](https://cloudplatform.googleblog.com/2015/11/faster-builds-for-Java-developers-with-Maven-Central-mirror.html) via [Coursier's mirrors mechanism](https://get-coursier.io/blog/2019/07/05/1.1.0-M14.html#mirrors):
```
~$ cat .config/coursier/mirror.properties 
gcs-maven-central.from=https://repo1.maven.org/maven2
gcs-maven-central.to=https://maven-central.storage-download.googleapis.com/repos/central/data/
```

PR #1259, opened 2020-01-24
Explicitly set Coursier's cache policies

fthomas:
By not using the default cache policies we ensure that `cache` does not
download files if local ones are younger than the TTL.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -19,7 +19,7 @@ package org.scalasteward.core.coursier
 import cats.effect._
 import cats.implicits._
 import cats.{Applicative, Parallel}
-import coursier.cache.FileCache
+import coursier.cache.{CachePolicy, FileCache}
 import coursier.core.Project
 import coursier.interop.cats._
 import coursier.{Fetch, Info, Module, ModuleName, Organization, Versions}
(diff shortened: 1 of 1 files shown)


fthomas:
It seems that this change fixed my recent problems with Maven Central (#1215, #1216, #1218, and #1253). My expectation that `getVersions` only hits the internet if the cached versions are older than the `cacheTtl` didn't hold because of the default cache policies. By using `LocalUpdateChanging` only, `getVersions` now behaves as I expected.

fthomas:
I dig a little deeper and found that the default cache policies only caused repeated requests to Maven Central for sbt plugins that are not hosted on Central. For sbt plugins Coursier gets the versions from a directory listing (like https://repo1.maven.org/maven2/org/scalameta/sbt-scalafmt_2.12_1.0/) since there is no `maven-metadata.xml`. The result of this listing is cached if the plugin exists, but if it doesn't exist the listing is requested every time the `Versions` API is called, irregardless of the cache's TTL. And since the majority of sbt plugins is not hosted on Central, Scala Steward repeatedly requested (non-existing) directory listings of the sbt plugins of every project for which it then got blocked.

mzuehlke:
Good catch, this is true for artefacts coming from other repos than Maven Central, isn't it? Is this a missing feature / bug in Coursier ? To cache no only the existence but also the absence of an artefact ?

fthomas:
It might be a bug in the case of sbt plugins. I think the absence of normal artifacts is cached.

PR #1274, opened 2020-02-01
Support version suffixes in `UpdatePattern`

### LINKED ISSUE #1038: Target JRE as part of artifact name
In one of my projects I can see update of `"com.microsoft.sqlserver" % "mssql-jdbc"`
from `7.2.2.jre8` to `7.2.2.jre11`, where it should stay at `jre8`.
https://mvnrepository.com/artifact/com.microsoft.sqlserver/mssql-jdbc

I also know such JRE versioning exist for other projects like Guava.
https://mvnrepository.com/artifact/com.google.guava/guava

Maybe this case could be handled by Steward?

mzuehlke:
This PR implements the idea proposed by @fthomas in a [comment](https://github.com/fthomas/scala-steward/issues/1038#issuecomment-543178540) to issue #1038:

> `UpdatePattern`, which is used for `.scala-steward.conf`, allows to match versions but only via `startsWith`. In this case we want to match a version if it ends with `jre8`. I would propose to change the version matching code to `endsWith` when `UpdatePattern.version` starts with `.*`.

After finishing this I realised one shortcoming: it is not possible to restrict  for the same artefact the prefix and the suffix. E.g pin a major version and enforce `jre8`. One solution would be to use `.*` as a separator between prefix and suffix, e.g. `7..*jre8` (there are 2 dots so it matches only against version 7 and not against version 70)

fixes #1038

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -5,11 +5,15 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 ```properties
 # The only dependencies which match the given pattern are updated.
 # Each pattern must have `groupId`, and may have `artifactId` and `version`.
+# The version is treated as a prefix of the new version, unless it starts with
+# ".*" then it is treated as a suffix.
 # Defaults to empty `[]` which mean Scala Steward will update all dependencies.
 updates.allow  = [ { groupId = "com.example" } ]
 
 # The dependencies which match the given pattern are NOT updated.
 # Each pattern must have `groupId`, and may have `artifactId` and `version`.
+# The version is treated as a prefix of the new version, unless it starts with
+# ".*" then it is treated as a suffix.
 # Defaults to empty `[]` which mean Scala Steward will not ignore dependencies.
 updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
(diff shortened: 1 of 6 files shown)


fthomas:
Thanks for working on this @mzuehlke! Unfortunately there are now merge conflicts due to the merge of #1254.

I agree that it would be nice to specify both a prefix and a suffix. In addition to your proposal we could also allow regular expressions or a complex object like `{ prefix="7.", suffix="jre8" }` for the `version` field. I dislike regular expressions but I'm not yet decided if I like `.*` as separator or a complex object better. One is short but a little bit obscure while the other is self-explaining but longer.

mzuehlke:
I will  resolve the merge conflicts. I expected  that ;-) 

I agree with you that regular expression are too powerful and like your idea with the dedicated `prefix` and `suffix`:

```properties
1: { groupId = "foo", artifactId="bar", version="x.y."}

2a: { groupId = "foo", artifactId="bar", suffix="jre8"}
2b: { groupId = "foo", artifactId="bar", prefix="x.y." , suffix="jre8"}

3a: { groupId = "foo", artifactId="bar", version={suffix="jre8"}}
3b: { groupId = "foo", artifactId="bar", version={prefix="x.y." , suffix="jre8"}}
```  

While **1** is the simple case. Do you mean / prefer **2** or **3** for specifying a `suffix`?

fthomas:
I meant **3** since it makes clear that one specifies a version prefix and/or suffix. That is not really obvious for **2**.

mzuehlke:
I updated the PR to have version (optionally) as a complex object. This has less _magic_ than the ".*" version.

PR #1309, opened 2020-02-10
Load Scalafix migrations once at start-up

fthomas:
This changes a few things about the loading of Scalafix migrations:

* Default Scalafix migrations are now defined in a new `scalafix-migrations.conf` file that are packaged in the JAR as resource. @kubukoz proposed this idea in https://github.com/fthomas/scala-steward/pull/1298#issuecomment-583498840 for groupId changes. A big benefit is that default Scalafix migrations are defined in the same format as user-supplied migrations.

* All Scalafix migrations are now loaded at start-up and a decoding error will lead to Scala Steward exiting immediately. This way we provide rapid feedback if the migrations file contains syntax errors and we avoid repeatedly loading the same migrations again and again.

* [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md) is used as file format for Scalafix migrations. I think this is better suited for files that are written by humans since it allows comments for example. If someone feels strongly about this, I can be persuaded to change it back to JSON.


kubukoz:
This means if anyone used to pass JSON files with scalafix config, their files will no longer work, right? Or is JSON a valid subset of HOCON?

modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
```diff
@@ -58,9 +58,7 @@ final class EditAlg[F[_]](
   }
 
   def applyScalafixMigrations(repo: Repo, update: Update): F[Unit] =
-    migrationAlg.findMigrations(update) >>= { migrationsList =>
-      Nel.fromList(migrationsList).fold(F.unit) { migrations =>
-        logger.info(s"Applying migrations: $migrations") >> sbtAlg.runMigrations(repo, migrations)
-      }
+    Nel.fromList(migrationAlg.findMigrations(update)).fold(F.unit) { migrations =>
```
kubukoz: it's traverse!

modules/core/src/main/scala/org/scalasteward/core/io/package.scala
```diff
@@ -48,4 +50,9 @@ package object io {
   private def isScalafmtCoreUpdate(update: Update): Boolean =
     update.groupId === GroupId("org.scalameta") &&
       update.artifactIds.exists(_.name === "scalafmt-core")
+
+  def readResource[F[_]](resource: String)(implicit F: Sync[F]): F[String] =
+    Resource
+      .fromAutoCloseable(F.delay(Source.fromResource(resource)))
```
kubukoz: Nice, I didn't know about `Source.fromResource`!

modules/core/src/main/scala/org/scalasteward/core/sbt/package.scala
```diff
@@ -49,9 +48,8 @@ package object sbt {
     val name = "StewardPlugin.scala"
     // I don't consider reading a resource as side-effect,
     // so it is OK to call `unsafeRunSync` here.
-    Resource
-      .fromAutoCloseable(IO(Source.fromResource(s"org/scalasteward/plugin/$name")))
-      .use(src => IO(FileData(name, src.mkString)))
+    io.readResource[IO](s"org/scalasteward/plugin/$name")
```
kubukoz: `show`?

modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
fthomas: It always is. :-) Thanks, I'll change it.

modules/core/src/main/scala/org/scalasteward/core/sbt/package.scala
fthomas: I'm not a big fan of the `Show` type class and I don't think we use `show` anywhere else in the code base. If we would have an interpolator that ensures the interpolated value is a `String`, I'd use it here and probably everywhere else.

fthomas:
@kubukoz Yes, JSON is a subset of HOCON, see https://github.com/lightbend/config/blob/master/HOCON.md#goals--background:

> a JSON superset, that is, all valid JSON should be valid and should result in the same in-memory data that a JSON parser would have produced.

But I changed the name of the key `extraMigrations` to just `migrations`.  So at least this needs to updated.

modules/core/src/main/scala/org/scalasteward/core/sbt/package.scala
kubukoz: Alright, let's keep it with `s` then :)

kubukoz:
:shipit:

kubukoz:
I'm waiting for this with #1298 because I had to rebase to borrow `readResource` :D

modules/core/src/main/scala/org/scalasteward/core/io/package.scala
```diff
@@ -48,4 +50,9 @@ package object io {
...
+      .use(src => F.delay(src.mkString))
```
kubukoz: I just noticed this will block (maybe `fromResource` too)... shall we add Blocker here?

PR #1313, opened 2020-02-10
Add FileAlg.readResource

fthomas:
So that #1298 does not need to wait for #1309. /cc @kubukoz

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/sbt/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/sbt/package.scala
@@ -16,12 +16,11 @@
 
 package org.scalasteward.core
 
-import cats.effect.{IO, Resource}
+import cats.Functor
 import cats.implicits._
 import org.scalasteward.core.data.{ArtifactId, Dependency, GroupId, Version}
-import org.scalasteward.core.io.FileData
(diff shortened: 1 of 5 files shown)


kubukoz:
Nice, thanks!

modules/core/src/main/scala/org/scalasteward/core/sbt/package.scala
```diff
@@ -45,13 +44,8 @@ package object sbt {
       """addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.11")"""
     )
 
-  val stewardPlugin: FileData = {
+  def stewardPlugin[F[_]](implicit fileAlg: FileAlg[F], F: Functor[F]): F[FileData] = {
     val name = "StewardPlugin.scala"
-    // I don't consider reading a resource as side-effect,
```
kubukoz: Changed your mind? :D

fthomas: It seems I did. :-) I guess it was convenient to call `unsafeRunSync` here when this code was added. Now that `FileAlg` has `readResource` it seemed more convenient to properly use it.

PR #1321, opened 2020-02-13
add Bitrock to the companies using Scala Steward

massimosiani:


fthomas:
Thanks @massimosiani!

PR #602, opened 2019-06-20
Install sbt-scalafix plugin only during migrations

fthomas:
This fixes two issues. The first is a conflict between the globally
installed sbt-scalafix and sbt-scalafix comming from a repo that uses a
different version. Here is an example where scala-steward can't load a
build because of a conflict:
```
[info] Loading global plugins from /home/frank/.sbt/1.0/plugins
[info] Updating ProjectRef(uri("file:/home/frank/.sbt/1.0/plugins/"), "global-plugins")...
[info] Done updating.
[info] Compiling 1 Scala source to /home/frank/.sbt/1.0/plugins/target/scala-2.12/sbt-1.0/classes ...
[info] Done compiling.
[info] Loading settings for project recursionz-build-build from hack.sbt ...
[info] Loading project definition from /home/frank/scala-steward/workspace/repos/enzief/recursionz/project/project
[info] Updating ProjectRef(uri("file:/home/frank/scala-steward/workspace/repos/enzief/recursionz/project/project/"), "recursionz-build-build")...
[info] Done updating.
[info] Compiling 2 Scala sources to /home/frank/scala-steward/workspace/repos/enzief/recursionz/project/project/target/scala-2.12/sbt-1.0/classes ...
[info] Done compiling.
/home/frank/scala-steward/workspace/repos/enzief/recursionz/project/plugins.sbt:9: error: reference to scalafix is ambiguous;
it is imported twice in the same scope by
import enzief.Dependencies.SbtPlugin._
and import _root_.scalafix.sbt.ScalafixPlugin.autoImport._
addSbtPlugin(scalafix)
             ^
[error] Type error in expression
Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore?
```

The second issue is an OOME I've seen with http4s:
```
[info] Loading global plugins from /home/frank/.sbt/1.0/plugins
[info] Loading settings for project http4s-build from plugins.sbt,build.sbt ...
[info] Loading project definition from /home/frank/scala-steward/workspace/repos/http4s/http4s/project
[info] Updating ProjectRef(uri("file:/home/frank/scala-steward/workspace/repos/http4s/http4s/project/"), "http4s-build")...
[info] Done updating.
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
[info] Compiling 4 Scala sources to /home/frank/scala-steward/workspace/repos/http4s/http4s/project/target/scala-2.12/sbt-1.0/classes ...
[info] Done compiling.
[...]

PR #1326, opened 2020-02-15
Update dependencies for scalafix given via `scalafixDependencies`

### LINKED ISSUE #1032: Suggest updates for Scalafix dependencies
It would be fantastic if Scala Steward could also suggest updates for `scalafixDependencies` the same way it does for `libraryDependencies`.

Is that something that would need to be implemented in the sbt-updates plugin or could it be done directly in Scala Steward?

mzuehlke:
Scalafix rules given via `scalafixDependencies` are updated, too.
Scalafix rule updates are getting the label `"scalafix-rule-update"`.

Fixes #1032

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -107,6 +107,7 @@ lazy val core = myCrossProject("core")
 lazy val plugin = myCrossProject("plugin")
   .settings(noPublishSettings)
   .settings(
+    addSbtPlugin(Dependencies.scalafixPlugin),
     scalaVersion := "2.12.10",
     sbtPlugin := true,
     Compile / compile / wartremoverErrors -= Wart.Equals
--- a/build.sbt
+++ b/build.sbt
@@ -39,6 +39,7 @@ lazy val core = myCrossProject("core")
       Dependencies.commonsIo,
       Dependencies.coursierCore,
       Dependencies.coursierCatsInterop,
+      Dependencies.cron4sCore,
       Dependencies.fs2Core,
       Dependencies.http4sAsyncHttpClient,
       Dependencies.http4sCirce,
--- a/build.sbt
+++ b/build.sbt
@@ -107,7 +107,6 @@ lazy val core = myCrossProject("core")
 lazy val plugin = myCrossProject("plugin")
   .settings(noPublishSettings)
   .settings(
-    addSbtPlugin(Dependencies.scalafixPlugin),
     scalaVersion := "2.12.10",
     sbtPlugin := true,
     Compile / compile / wartremoverErrors -= Wart.Equals
--- a/modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
+++ b/modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
@@ -18,8 +18,8 @@ package org.scalasteward.plugin
 
 import sbt.Keys._
 import sbt._
+
 import scala.util.matching.Regex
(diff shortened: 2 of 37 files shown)


build.sbt
```diff
@@ -107,6 +107,7 @@ lazy val core = myCrossProject("core")
 lazy val plugin = myCrossProject("plugin")
   .settings(noPublishSettings)
   .settings(
+    addSbtPlugin(Dependencies.scalafixPlugin),
```
fthomas: Adding sbt-scalafix to the build here does not mean that its settings or tasks can be used in `StewardPlugin.scala`. The reason is that this file is added as resource to the `core` project and while Scala Steward is running this source file is added as global project to `$HOME/.sbt/1.0/plugins/`. As far as I know, these global source-based plugins can't declare dependencies on other plugins it can only use settings and keys of ther global plugins (that was the reason why it could use sb-updates' settings before, because it was also installed as global plugin).

It would be great if we could support `scalafixDependencies` without adding sbt-scalafix as global plugin for all projects. The reason I'm wary of adding `sbt-scalafix` as global plugin is that I've seen it breaking projects that already used that plugin locally. I'm wondering if sbt settings/tasks can be called by name as `String` if they are defined in the build. That way we could get `scalafixDependencies` only in projects that already depend on sbt-scalafix

fthomas: https://github.com/fthomas/scala-steward/pull/602 documents why I'd like to avoid adding sbt-scalafix as global plugin.

mzuehlke: This is a very good remark. I overlooked that problem.

I will find a way to not install `sbt-scalafix` as a global plugin.

mzuehlke: was more complicated, that I first thought.

modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
```diff
@@ -96,6 +102,13 @@ object StewardPlugin extends AutoPlugin {
   private def isCompilerPlugin(moduleId: ModuleID): Boolean =
     moduleId.configurations.exists(_ == "plugin->default(compile)")
 
+  lazy val findScalafixDependencies: Def.Initialize[Seq[ModuleID]] = Def.settingDyn {
+    val scalafixDependencies = SettingKey[Seq[ModuleID]]("scalafixDependencies")
+    Def.setting {
+      scalafixDependencies.value
```
fthomas: That looks interesting. What happens if a project does not depend on sbt-scalafix and this is evaluated? Will it just return an empty `Seq` or does it throw?

mzuehlke: I realized that I had `sbt-scalafix` in my global plugins 😞 
So I went back to the code from `mdoc` I had used as an inspiration and fixed my implementation.
I tested it with a  test project with and  with another without `sbt-scalafix`

fthomas: It is cool that this works!

fthomas:
I just tried this out: https://github.com/fthomas/scalafix-test/pull/29

Very cool! Thanks @mzuehlke!

fthomas:
Btw, we can watch which `scalafixDependencies` are updated by Scala Steward here:
https://github.com/pulls?q=author%3Ascala-steward+is%3Apr+%22scalafix-rule-update%22

mzuehlke:
Interesting link

Issue #994, opened 2019-09-24
Add formatting options to commit messages


chuwy:
We have a convention to name all our dependency-bump commits as `Bump some-dependency to A.B.C (close #XYZ)` (where `XYZ` can be a PR number).

Scala Steward creates `Update some-dependency to A.B.C` commits.

Wondering if this is something that can be configured via `.scala-steward.conf`.

exoego:
Formatting is good to have.
But how do Scala Steward know issue number to close?

dilyand:
Steward could be in charge of opening the issue as well. Then it will know to close the issue it opened.

fthomas:
Customizing the commit message sounds like a small and uncontroversial change to me and I'd happily accept a PR for that. The other stuff about creating or finding issues sounds like a lot of work that benefit only a small number of people. I don't think I'd want to maintain such code here.

tyoras:
I also have the need for commit message customization, and I think I'll open a PR for this soon.
My idea is to keep the current commit message but allow to prefix and/or to suffix it with an optionnal custom message part that would be configurable in `.scala-steward.conf`

mzuehlke:
I saw your PR, but I would suggest a little more general attempt.
Instead of only adding a prefix and/or a suffix to the existing message, why not having a simple template that can contains some know variables:
```
# known variables: ${artifactName},  ${currentVersion},  ${nextVersion}
commit.message.template = "Update ${artifactName} to ${nextVersion}"
```

tyoras:
@mzuehlke good suggestion, it would allow to have a completely customizable message.
I will update the PR with this change.

PR #1330, opened 2020-02-16
Add formatting options to commit messages

tyoras:
Closes https://github.com/fthomas/scala-steward/issues/994


fthomas:
This looks great, thanks @tyoras! Running `sbt fmt` should make the CI green. I'm wondering if a `${default}` variable would also make sense. So for example, if you just want to add a prefix to the message, one could write `commits.message = "my_prefix ${default}"`. WDYT @tyoras?

tyoras:
Thanks, no problem I'll format and add support for `${default}` variable today.

tyoras:
@fthomas done, and I've also rebased the PR

PR #1333, opened 2020-02-17
Add Besedo to the community list

Koroeskohr:
Installing Scala Steward on our Gitlab instance wasn't that hard, in the end :)


fthomas:
> Installing Scala Steward on our Gitlab instance wasn't that hard, in the end :)

That would make a perfect "customer review". :smile: Thanks @Koroeskohr!

PR #1336, opened 2020-02-18
Add config to specify how frequent PRs are created

### LINKED ISSUE #1316: Allow users to specify how frequent PRs can be created
Scala Steward currently creates pull requests as soon as it detects that a new version is available. This leads to a lot of PRs for projects that have many dependencies or dependecies that are very frequently updated. To reduce the noise this creates for maintainers of such projects, we should add the ability control how frequent Scala Steward is allowed to create PRs in a repo via an option in `.scala-steward.conf`. This option could be a simple enum with the cases `asap | daily | weekly | monthly`.

See https://twitter.com/olafurpg/status/1227392299904258048 for more context.

fthomas:
This PR adds a new `pullRequests.frequency` option to `.scala-steward.conf` files which can be used to control how often or when Scala Steward is allowed to create pull requests. Possible values of this option are `@asap` (default), `@daily`, `@weekly`, `@monthly`, or a [cron4s](https://www.alonsodomin.me/cron4s/) compatible CRON expression. The effects of these values are explained in [repo-specific-configuration.md](https://github.com/fthomas/scala-steward/blob/aaa4284fb1fcaef6f083c8f8b412c1e4cd2b8470/docs/repo-specific-configuration.md).

Some example configs:
```hocon
# only create PRs on weekdays
pullRequests.frequency = "* * ? * mon-fri"

# wait 7 days after the last PR before creating new PRs
pullRequests.frequency = "@weekly"

# create PRs only once on mondays
pullRequests.frequency = "0 8 ? * MON"
```

Note that this option does not affect the number of pull requests Scala Steward creates when it is allowed to. So if the option is set to `@weekly` Scala Steward might create n PRs at a given time and then a week later it might create another m pull requests.

Closes #1316.

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -16,8 +16,8 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 #   <CRON expression>
 #     PRs are created roughly according to the given CRON expression.
 #
-#     CRON expressions consist of six fields: seconds, minutes, hour of day,
-#     day of month, month, and day of week.
+#     CRON expressions consist of five fields:
+#     minutes, hour of day, day of month, month, and day of week.
 #
 #     See https://www.alonsodomin.me/cron4s/userguide/index.html#parsing for
 #     more information about the CRON expressions that are supported.
@@ -28,7 +28,7 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 #
 # Default: @asap
 #
-#pullRequests.frequency = "0 0 0 ? * 3" # every thursday on midnight
+#pullRequests.frequency = "0 0 ? * 3" # every thursday on midnight
 pullRequests.frequency = "@weekly"
 
 # Only these dependencies which match the given patterns are updated.
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -28,7 +28,7 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 #
 # Default: @asap
 #
-#pullRequests.frequency = "0 0 * ? * 3" # every thursday on midnight
+#pullRequests.frequency = "0 0 0 ? * 3" # every thursday on midnight
 pullRequests.frequency = "@weekly"
 
 # Only these dependencies which match the given patterns are updated.
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -3,6 +3,34 @@
 You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward updates your repository.
 
 ```properties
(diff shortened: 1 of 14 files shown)


fthomas:
I think this is ready now. This PR adds a new `pullRequests.frequency` config option to `.scala-steward.conf` files which can be used to control how often or when Scala Steward is allowed to create pull requests. Possible values of this option are `@asap` (the default), `@daily`, `@weekly`, `@monthly`, or a [cron4s](https://www.alonsodomin.me/cron4s/) compatible CRON expression. The effects of these values are explained in [repo-specific-configuration.md](https://github.com/fthomas/scala-steward/blob/e851c85ab59bfc56971b8481c7151a007a7c4525/docs/repo-specific-configuration.md).

docs/repo-specific-configuration.md
```diff
@@ -3,6 +3,34 @@
...
+#     PRs are created at least 1 day | 7 days | 30 days after the last PR.
+#
+#   <CRON expression>
+#     PRs are created roughly according to the given CRON expression.
+#
+#     CRON expressions consist of six fields: seconds, minutes, hour of day,
+#     day of month, month, and day of week.
+#
+#     See https://www.alonsodomin.me/cron4s/userguide/index.html#parsing for
+#     more information about the CRON expressions that are supported.
```
mzuehlke: maybe add a link to web based cron tan generator like https://crontab.guru/

fthomas: I'd love to but cron4s requires 6 fields and the generator has only 5.

mzuehlke: you are right. I missed that.

mwz:
That's nice 👍 A few somewhat related thoughts:
Perhaps it's worth adding a note to be careful when using this setting in combination with `updates.limit`?
A colleague of mine (@JackTreble) described Scala Steward to me as a great way to track your tech debt (I didn't really look at it that way before), so even if it takes you a while to merge away those PRs, it's probably still better to be aware of all of your outstanding updates rather than pretend that they don't exist and live in ignorance (oh boy, it's hard to imagine how we managed before Scala Steward existed!). I agree with that, but I can also see how some people would want to reduce the amount of noise, especially if they maintain a large number of repos. A combination of both settings definitely gives you more control over that, but the downside can be hidden tech debt and potentially not getting important dependency updates. (Are PRs opened/updated for dependencies in alphabetical order? In such case would it make sense to have a mechanism allowing you to prioritise specific dependencies over others so that you will always get updates for the packages that are the most important to you?)

docs/repo-specific-configuration.md
fthomas: I'm thinking out loud: we'll never be able to take the "seconds" field into account, so maybe we should reduce the number of fields to 5 by always prepending `* ` to the CRON expressions. That way we'll only the support the more common format that uses only 5 fields.

mzuehlke: I like this idea. 👍 
As the crontab syntax is cryptic and we have no possibility to interact with the user. Making it simpler is a good idea.

PR #1354, opened 2020-03-03
Mark com.nequissimus:sort-imports:36845576 as an invalid version.

mwz:
It appears that the first version of [NeQuissimus/sort-imports](https://github.com/NeQuissimus/sort-imports) scalafix rule was published with an incorrect version back in June [36845576](https://search.maven.org/artifact/com.nequissimus/sort-imports_2.12), which is getting picked up by Scala Steward as the latest version. See the following PR https://github.com/beautiful-scala/sbt-scalastyle/pull/13.

Could this version be ignored globally by Scala Steward?

cc @NeQuissimus

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ workspace/
 
 .bloop/
 .metals/
+metals.sbt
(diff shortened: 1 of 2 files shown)


NeQuissimus:
:+1: from me. I was not even aware of the artifact :-)

PR #1361, opened 2020-03-15
Support sbt credentials for coursier

### LINKED ISSUE #1291: Extract credentials from sbt and pass them to Coursier
Some package repositories require credentials for checking which versions of an artifact are available but we currently ignore credentials when extracting information from sbt. To support finding updates in private package repositories that require credentials for read access, we need to include credentials in `StewardPlugin`'s `stewardDependencies` task, add them to [`Scope`](https://github.com/fthomas/scala-steward/blob/83b85f1edad64e8ad7a024e556f9f89cb0ffbb23/modules/core/src/main/scala/org/scalasteward/core/data/Scope.scala), and pass them to Coursier in `CoursierAlg`.

sbt has two types of credentials, [`DirectCredentials`](https://www.scala-sbt.org/1.x/api/sbt/librarymanagement/ivy/DirectCredentials.html) and [`FileCredentials`](https://www.scala-sbt.org/1.x/api/sbt/librarymanagement/ivy/FileCredentials.html) and I think we should also support both.

implmnt:
This PR fixes #1291

### DIFF
--- a/modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
+++ b/modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
@@ -20,6 +20,7 @@ import sbt.Keys._
 import sbt._
 
 import scala.util.matching.Regex
+import scala.util.Try
 
 object StewardPlugin extends AutoPlugin {
   override def trigger: PluginTrigger = allRequirements
@@ -52,8 +53,7 @@ object StewardPlugin extends AutoPlugin {
       val dependencies = libraryDeps ++ scalafixDeps
 
       def getCredentials(url: URL): Option[Resolver.Credentials] =
-        Credentials
-          .forHost(sbtCredentials, url.getHost)
+        Try(Credentials.forHost(sbtCredentials, url.getHost)).toOption.flatten
           .map(c => Resolver.Credentials(c.userName, c.passwd))
 
       val resolvers = fullResolvers.value.collect {
--- a/modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
+++ b/modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
@@ -18,6 +18,7 @@ package org.scalasteward.plugin
 
 import sbt.Keys._
 import sbt._
+
 import scala.util.matching.Regex
 
 object StewardPlugin extends AutoPlugin {
@@ -37,6 +38,7 @@ object StewardPlugin extends AutoPlugin {
       val buildRoot = baseDirectory.in(ThisBuild).value
       val scalaBinaryVersionValue = scalaBinaryVersion.value
       val scalaVersionValue = scalaVersion.value
+      val sbtCredentials = credentials.value
 
       val libraryDeps = libraryDependencies.value
         .filter(isDefinedInBuildFiles(_, sourcePositions, buildRoot))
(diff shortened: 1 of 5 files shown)


fthomas:
This looks great, many thanks @implmnt! 

I was wondering if there is a specific reason to load `FileCredentials` via `Credentials.forHost` instead of passing them along to Coursier?

implmnt:
Hi @fthomas, thank you for the great bot!
Actually FileCredentials in SBT and FileCredentials in Coursier are different and not compatible, so it cant be just passed from one to another.
I also looked into sbt-coursier that uses the same approach.

fthomas:
Thanks! I didn't noticed that the file credentials are incompatible.

[sbt's file credentials](https://www.scala-sbt.org/1.x/docs/Publishing.html#Credentials) look like this:
```
realm=My Nexus Repository Manager
host=my.artifact.repo.net
user=admin
password=admin123
```
while [Coursier's file credentials](https://get-coursier.io/docs/other-credentials.html#property-file) look like this:
```
simple.username=simple
simple.password=SiMpLe
simple.host=artifacts.simple.com
simple.realm=simple realm
```
So the difference is that Coursier requires a prefix and the key for user / username is different.

implmnt:
Yes, you're absolutely right! I may have misspoken about credentials' incompatiblity. 
The point was that coursier file credentials' parser isn't compatible with sbt credentials out of the box. That's why sbt-coursier doesn't pass sbt file credentials directly to coursier.

In my vision, to make it possible either the extra transformation stage before parsing or the custom credentials parser should be used.

Please, feel free to notify me if I'm wrong or if you have a different point of view.

fthomas:
@implmnt The approach you took in this PR seems reasonable to me now that I know of the differences between the two file credentials formats. I just wanted to understand why the file credentials can't be simply forwarded to Coursier.

> In my vision, to make it possible either the extra transformation stage before parsing or the custom credentials parser should be used.

Do you mean a parser that reads sbt's file credentials as `coursier.credentials.DirectCredentials`? Does such a thing already exist and could we use it in Scala Steward?

implmnt:
> Do you mean a parser that reads sbt's file credentials as `coursier.credentials.DirectCredentials`? Does such a thing already exist and could we use it in Scala Steward?

Yes, that's what I meant, but unfortunately it doesn't exist and that's why I used `Credentials.forHost`, it's straightforward and takes less effort.

modules/plugin/src/main/scala/org/scalasteward/plugin/StewardPlugin.scala
```diff
@@ -49,11 +51,19 @@ object StewardPlugin extends AutoPlugin {
         )
       val dependencies = libraryDeps ++ scalafixDeps
 
+      def getCredentials(url: URL): Option[Resolver.Credentials] =
+        Credentials
+          .forHost(sbtCredentials, url.getHost)
```
fthomas: [`forHost`](https://github.com/sbt/librarymanagement/blob/879bd5399c6d53a3c692513cd1e91d24c89eacc7/ivy/src/main/scala/sbt/librarymanagement/ivy/Credentials.scala#L36) throws if it can't load `FileCredentials`. If we let `stewardDependencies` return with an error in this case, Scala Steward won't be able to do anything in such a repo even if the credentials are only required for publishing. I think we should therefore return `None` here if `forHost` throws.

implmnt: True! Thanks for catching this!

PR #1370, opened 2020-03-22
Optimize docker image size

### LINKED ISSUE #347: Firejail doesn't work in Docker
Hi @fthomas, I've discovered that firejail doesn't work in Docker:
```bash
$ firejail echo hello
Warning: an existing sandbox was detected. echo will run without any additional sandboxing features
```

Running it with the `--force` flag, which is supposed to disable the PID namespace checking, results in the following error:
```bash
$ firejail --force echo hello
Error clone: main.c:2519 main: Operation not permitted
```

Additionally, the `--force` flag was removed in firejail 0.9.54 and apparently it is not possible to run firejail in Docker anymore (I asked about this recently https://github.com/netblue30/firejail/issues/2579).

I think we should remove firejail from scala-steward docker image and force it to run without the sandbox when the command isn't found or make it explicit in the docs that the sandboxing features don't work when running scala-steward in Docker.

implmnt:
This PR changes docker base image from debian`openjdk:8` to alpine `openjdk:8-jre-alpine`.

The motivation was to reduce the resulting image size. 
I replaced JDK with JRE since neither sbt nor scalac require JDK.
I also removed `firejail` from image and  put `--disable-sandbox` to docker entrypoint args to fix #347.

The resulting image size is reduced from 1.06GB to 208MB.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -170,36 +170,22 @@ lazy val metadataSettings = Def.settings(
 )
 
 lazy val dockerSettings = Def.settings(
-  dockerCommands := Seq(
-    Cmd("FROM", Option(System.getenv("DOCKER_BASE_IMAGE")).getOrElse("openjdk:8")),
-    Cmd("ARG", "DEBIAN_FRONTEND=noninteractive"),
-    ExecCmd("RUN", "apt-get", "update"),
-    ExecCmd("RUN", "apt-get", "install", "-y", "apt-transport-https", "firejail"),
-    ExecCmd(
-      "RUN",
-      "sh",
-      "-c",
-      """echo \"deb https://dl.bintray.com/sbt/debian /\" | tee -a /etc/apt/sources.list.d/sbt.list"""
-    ),
-    ExecCmd(
-      "RUN",
-      "apt-key",
-      "adv",
-      "--keyserver",
-      "hkp://keyserver.ubuntu.com:80",
-      "--recv",
-      "2EE0EA64E40A89B84B2DF73499E82A75642AC823"
-    ),
-    ExecCmd("RUN", "apt-get", "update"),
-    ExecCmd("RUN", "apt-get", "install", "-y", "sbt"),
-    Cmd("WORKDIR", "/opt/docker"),
-    Cmd("ADD", "opt", "/opt"),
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -170,36 +170,22 @@ lazy val metadataSettings = Def.settings(
...
-    ExecCmd("ENTRYPOINT", "/opt/docker/bin/scala-steward"),
-    ExecCmd("CMD", "")
-  ),
+  dockerBaseImage := Option(System.getenv("DOCKER_BASE_IMAGE")).getOrElse("openjdk:8-jre-alpine"),
+  dockerCommands ++= {
+    val getSbtVersion = sbtVersion.value
+    val sbtTgz = s"sbt-$getSbtVersion.tgz"
+    val sbtUrl = s"https://github.com/sbt/sbt/releases/download/v$getSbtVersion/$sbtTgz"
+    Seq(
+      Cmd("USER", "root"),
```
mwz: Could you switch back at the end to a non-privileged user to avoid running scala-steward as root?

fthomas: A PR for that would be welcome.

implmnt: Oh.. I was certain the setting `daemonUser` will be used in entrypoint. 
It doesn't work like the official description says:
- _daemonUser in Docker_
_The user to use when executing the application. Files below the install path also have their ownership set to this user._

There are two ways I can suggest to fix this. We can switch back to `daemonUser` in `dockerCommands` (just by adding another `Cmd(..)` to the end of the list) or modify `dockerEntrypoint` that will always start steward as `daemonUser`.
```scala
dockerEntrypoint := {
    val user = (Docker / daemonUser).value
    val runSteward = (dockerEntrypoint.value :+ "--disable-sandbox").toList
    "su" :: "-s" :: "/bin/bash" :: user :: "-c" :: runSteward
  }
```
Which way is more preferable @fthomas @mwz?

mwz: Personally, I'd switch by running a new cmd at the end of docker commands.

fthomas: Running a new cmd seems to be the shorter solution. I would prefer that too, since it means less code to maintain.

Issue #145, opened 2018-11-08
Also update scala version


erikvanoosten:
It looks like scala-steward doesn't try to update the scala version.

For example in the build.sbt of metrics-scala we have: `scalaVersion := "2.12.6"` and `crossScalaVersions := Seq("2.11.12", "2.12.6")`. Version 2.12.7 is already out.

fthomas:
I found that updating the Scala version is trickier and more error-prone than updating libraries or plugins. E.g in most of my projects the Scala version is only defined in the `.travis.yml` file. The naive approach also generated some bad PRs like this one: https://github.com/fthomas/singleton-ops/pull/76/files. If we find a reliable way to update the Scala version, I'm all for letting scala-steward do it.

erikvanoosten:
Neat. I have never seen that. I still prefer to have scala versions in my build.sbt.

What about a simple heuristic that only matches: `scalaVersion := "([^"]+)"` and `crossScalaVersions := Seq\(("([^"]+)")+\)`? Presumably this will catch most usages.

For the second pattern only minor versions should be suggested. If a new major version is available, the suggestion could be to add it to the sequence.

thesamet:
With all the 2.13 milestones and RCs coming out - it would be really nice to have scala-steward able to push us all to the latest versions. Since those milestones and RCs are not binary compatible, scala-steward would have to check that all dependencies have been published for that specific version before making a PR.

nafg:
I don't get the benefit of having the scala version in .travis.yml -- then needing to pass `++$TRAVIS_SCALA_VERSION` to sbt.

It seems much easier to use `crossScalaVersions` in `build.sbt` and then just put a `+` before the sbt command. Besides, the build is then much more portable and reproducible.

anilkumarmyla:
Is this something that can be done under a special command-line flag that can be turned on by people interested in this? Honestly I would settle for a bad PR (that breaks the build and needs intervention) than no PR at all for scala upgrades 🙂

fthomas:
@anilkumarmyla The code that prevents updating Scala is [here](https://github.com/fthomas/scala-steward/blob/381670a4d38978d1e8e725d45cb4d5e233e15e19/modules/core/src/main/scala/org/scalasteward/core/update/FilterAlg.scala#L78-L81). I'm open to adding a new flag to `.scala-steward.conf` that overrules these lines.

anilkumarmyla:
[...]

PR #1381, opened 2020-03-28
Update scala version

anilkumarmyla:
Attempts a stab at #145 by introducing a flag `updates.includeScala` which when set to true (default: false) updates scala version as well

Tested this on a personal repo and it works, although needs some more work to update unit tests and docs (and of course a thorough review since I still feel new to the codebase)

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala
@@ -47,7 +47,7 @@ final class PruningAlg[F[_]](
       case None => F.pure((false, List.empty))
       case Some(repoCache) =>
         val ignoreScalaDependency =
-          !repoCache.maybeRepoConfig.exists(_.updates.defaultIncludeScala)
+          !repoCache.maybeRepoConfig.flatMap(_.updates.includeScala).getOrElse(false)
         val dependencies = repoCache.dependencyInfos
           .flatMap(_.sequence)
           .collect {
--- a/modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala
@@ -23,7 +23,7 @@ import io.chrisdavenport.log4cats.Logger
 import org.scalasteward.core.data._
 import org.scalasteward.core.nurture.PullRequestRepository
 import org.scalasteward.core.repocache.{RepoCache, RepoCacheRepository}
-import org.scalasteward.core.repoconfig.{PullRequestFrequency, RepoConfig, RepoConfigAlg}
+import org.scalasteward.core.repoconfig.{PullRequestFrequency, RepoConfig}
 import org.scalasteward.core.update.PruningAlg._
 import org.scalasteward.core.update.data.UpdateState
 import org.scalasteward.core.update.data.UpdateState._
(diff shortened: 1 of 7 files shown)


anilkumarmyla:
This is ready to be reviewed

The codecov pr build failure however is a false positive since added code increases the coverage

modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala
```diff
@@ -37,20 +37,32 @@ final class PruningAlg[F[_]](
     implicit
     dateTimeAlg: DateTimeAlg[F],
     logger: Logger[F],
+    repoConfigAlg: RepoConfigAlg[F],
```
anilkumarmyla: just realized RepoConfig is already part of RepoCache, so injecting RepoConfigAlg doesn't  make sense here

exoego:
This is awesome job 👍 

It would be great if this feature also update `scala` versions in `.travis.yml`, since many Scala OSSs uses Travis as their CI.

anilkumarmyla:
Thanks @exoego , this is just the start, not planning to cover other enhancements here yet

Enhancements
- Updating scala version in other files (ex: .travis.yml)
- Running scalafix for scala major version upgrades

anilkumarmyla:
@fthomas please take a look and let me know your thoughts/comments

modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala
```diff
@@ -46,9 +46,14 @@ final class PruningAlg[F[_]](
     repoCacheRepository.findCache(repo).flatMap {
       case None => F.pure((false, List.empty))
       case Some(repoCache) =>
+        val ignoreScalaDependency =
+          !repoCache.maybeRepoConfig.flatMap(_.updates.includeScala).getOrElse(false)
```
fthomas: Could you add `UpdatesConfig.defaultIncludeScala: Boolean` and use that instead of `false` here? This way all default config values are defined under the `repoconfig` package which makes it easier to figure out what the default values of the various config options are.

anilkumarmyla: updated, thanks! let me know if the change is as intended

fthomas: Sorry, I should have been more clear. What I had in mind was this:
```scala
case class UpdatesConfig(
  ...
  includeScala: Option[Boolean]
)

object UpdatesConfig {
  ...
  val defaultIncludeScala: Boolean = false
}

// and then in PruningAlg
!repoCache.maybeRepoConfig.flatMap(_.updates.includeScala)
  .getOrElse(UpdatesConfig.defaultIncludeScala)
```

The optional `includeScala` field in the `UpdatesConfig` class was fine. If this is changed to a `Boolean` with a default value, it means that the default value will be persisted in `RepoCache` objects by `RepoCacheAlg`. But persisted default values make it harder to change the default later because we can't differentiate between a persisted default value and the same value that was explicitly set by the user. (#1339 and #1344 has been done for the same reason.)

anilkumarmyla: fixed, thanks for the detailed explanation

erikvanoosten:
Its not entirely working as I expected. It should not update the major scala version in `crossScalaVersions`. For example: https://github.com/erikvanoosten/metrics-scala/pull/245/files

anilkumarmyla:
yeah, we'd have many bad PRs like this since its experimental 😬

Issue #1340, opened 2020-02-20
Add link to .scala-steward.conf documentation to PRs


fthomas:
Recent PRs enhanced the capabilities of repo specific configuration via `.scala-steward.conf` (#1274, #1254, #1330, #1336). To increase the discoverability of these features, we could link to the [`.scala-steward.conf` documentation](https://github.com/fthomas/scala-steward/blob/master/docs/repo-specific-configuration.md) in PR bodies.

PR #1383, opened 2020-03-30
Add repo specific conf link to PR body

kiranbayram:
Addresses #1340 

 - I also changed the title of Ignore future updates section because we can use that part for repo specific conf examples

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,7 +78,13 @@ lazy val core = myCrossProject("core")
           defaultStrategy(otherwise)
       }
     },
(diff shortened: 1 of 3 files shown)


kiranbayram:
Reverted the hint about PR frequency @fthomas

fthomas:
I moved the link to the documentation into its own paragraph because I thought that it was to much content for the `<summary>` tag. I also used the current `HEAD` commit instead of `master` in the link so that the documentation is always in sync with the running Scala Steward instance.

PR #1406, opened 2020-04-20
Introduce an option in the repository config for the file extension

### LINKED ISSUE #1376: Look for corresponding changes in things like '.md' files
I don't know how common this use case is, but I have some dependencies that also appear in my markdown docs, for example:

```
how to install:

```scala
resolvers += "manyangled" at "https://dl.bintray.com/manyangled/maven/"

libraryDependencies ++= Seq(
  // basic coulomb dependencies
  "com.manyangled" %% "coulomb" % "0.3.6",
  "org.typelevel" %% "spire" % "0.17.0-M1",
  "eu.timepit" %% "singleton-ops" % "0.4.0",
  // pureconfig integration:
  "com.manyangled" %% "coulomb-pureconfig" % "0.3.5",
  "com.github.pureconfig" %% "pureconfig" % "0.12.0"
)
```

Using the example above, if scala steward updated the pureconfig dep in my sbt build, in theory it also might update it in my markdown.

mzuehlke:
A user option to make it possible to change version numbers in text files. This will hopefully create not as many unrelated PRs as the last attempt.

fixes #1376

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/io/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/io/package.scala
@@ -37,10 +37,7 @@ package object io {
 
   private def isGenericSourceFile(file: File, fileExtensions: Set[String]): Boolean = {
     val name = file.name
-    val allowedByExtension = file.extension.exists(fileExtensions)
-    def allowedBySuffix: Boolean =
-      Set(".sbt.shared").exists(suffix => name.endsWith(suffix) && !name.startsWith(suffix))
-    allowedByExtension || allowedBySuffix
+    fileExtensions.exists(suffix => name.endsWith(suffix) && !name.startsWith(suffix))
   }
 
   private def isSbtUpdate(update: Update): Boolean =
--- a/modules/core/src/main/scala/org/scalasteward/core/io/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/io/package.scala
@@ -21,21 +21,23 @@ import cats.implicits._
 import org.scalasteward.core.data.{GroupId, Update}
 
 package object io {
-  def isSourceFile(update: Update)(file: File): Boolean = {
+  def isSourceFile(update: Update, fileExtensions: Set[String])(file: File): Boolean = {
     val notInGitDir = !file.pathAsString.contains(".git/")
-    notInGitDir && isSpecificOrGenericSourceFile(update)(file)
+    notInGitDir && isSpecificOrGenericSourceFile(update, fileExtensions)(file)
   }
 
-  private def isSpecificOrGenericSourceFile(update: Update)(file: File): Boolean =
+  private def isSpecificOrGenericSourceFile(update: Update, fileExtensions: Set[String])(
+      file: File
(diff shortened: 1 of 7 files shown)


modules/core/src/main/scala/org/scalasteward/core/io/package.scala
```diff
@@ -21,21 +21,23 @@ import cats.implicits._
...
+      case _                                 => isGenericSourceFile(file, fileExtensions)
     }
 
-  private def isGenericSourceFile(file: File): Boolean = {
+  private def isGenericSourceFile(file: File, fileExtensions: Set[String]): Boolean = {
     val name = file.name
-    val allowedByExtension = file.extension.exists(Set(".scala", ".sbt", ".sc", ".yml"))
+    val allowedByExtension = file.extension.exists(fileExtensions)
     def allowedBySuffix: Boolean =
       Set(".sbt.shared").exists(suffix => name.endsWith(suffix) && !name.startsWith(suffix))
```
fthomas, line 42: It would be great if suffixes like `.sbt.shared` could be also handled by this feature. I guess we could just remove `val allowedByExtension` and use `fileExtensions` instead of `Set(".sbt.shared")` in `def allowedBySuffix`. WDYT?

modules/core/src/test/scala/org/scalasteward/core/repoconfig/RepoConfigAlgTest.scala
```diff
@@ -26,6 +26,7 @@ class RepoConfigAlgTest extends AnyFunSuite with Matchers {
          |updates.ignore = [ { groupId = "org.acme", version = "1.0" } ]
          |updates.limit = 4
          |updates.includeScala = true
+         |updates.fileExtensions = [ "*.txt" ]
```
fthomas, line 29: I know this is just a test, but maybe we should remove `*` here to not give the anyone the impression that it is needed for `fileExtensions`.

mzuehlke: I agree. It would be a bad example.

modules/core/src/main/scala/org/scalasteward/core/io/package.scala
mzuehlke: I agree, it makes sense. I merged these 2 tests.
Q: What should `!name.startsWith(suffix)` prevent ? That the suffix is the complete filename `name != suffix` ?

mzuehlke:
Thanks for comments. I improved the PR, butI kept the term `fileExtensions` in the config after merging it with suffix handling as I think is more widely used.

modules/core/src/main/scala/org/scalasteward/core/io/package.scala
fthomas: That code was added in #1194 but it does not mention the reason for that check. But yes, I think I wanted to make sure that the suffix is a proper suffix.

erikerlandson:
thanks @mzuehlke & @fthomas !

PR #1407, opened 2020-04-20
check for all artifact URLs return by coursier, if they exists.

### LINKED ISSUE #1403: PRs with broken urls
Hello there,

Today i saw a PR (repo is private so can't give a link) where urls were broken. So i was wondering if we can use `ExistenceClient` to check urls returned by `CoursierAlg.getArtifactIdUrlMapping` and either filter the ones with broken url out or show them as text (instead of link).

WDYT?

mzuehlke:
Include them only in the PR, if they exist.

( I didn't perform the filtering before generating the release related URLs. But I'm not sure about that)

fixes #1403

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
@@ -149,16 +149,17 @@ final class NurtureAlg[F[_]](
       artifactIdToUrl <- coursierAlg.getArtifactIdUrlMapping(
         Scope(data.update.dependencies.toList, resolvers)
       )
-      releaseRelatedUrls <- artifactIdToUrl
+      existingArtifactUrlsList <- artifactIdToUrl.toList.filterA(a => existenceClient.exists(a._2))
+      existingArtifactUrlsMap = existingArtifactUrlsList.toMap
+      releaseRelatedUrls <- existingArtifactUrlsMap
         .get(data.update.mainArtifactId)
         .traverse(vcsExtraAlg.getReleaseRelatedUrls(_, data.update))
       branchName = vcs.createBranch(config.vcsType, data.fork, data.update)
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -153,10 +154,11 @@ final class NurtureAlg[F[_]](
         .traverse(vcsExtraAlg.getReleaseRelatedUrls(_, data.update))
       branchName = vcs.createBranch(config.vcsType, data.fork, data.update)
       migrations = migrationAlg.findMigrations(data.update)
+      existingArtifactUrls <- artifactIdToUrl.toList.filterA(a => existenceClient.exists(a._2))
```
kiranbayram, line 157: I think we shouldn't filter out the whole artifact when url is broken. We can still write the artifact id in PR body but as regular text instead of a url.

mzuehlke: When there is no URL for an `artifactId` we write the `artifactId` without the URL:
https://github.com/fthomas/scala-steward/blob/master/modules/core/src/main/scala/org/scalasteward/core/vcs/data/NewPullRequestData.scala#L121-L124

kiranbayram: Oh i see, i wasn't aware of that 👍

fthomas:
Thanks for the PR, @mzuehlke!

If the filtering would be done before trying to find release related URLs, we could save a lot of calls to `HttpExistenceClient.exists` with URLs that most certainly do not exist. The URL from Coursier is the source URL from which we derive the release related URLs. If the source URL does not exists, I would expect that none of the derived URLs exist either.

mzuehlke:
I agree and changed the PR to perform the check before generating release related URLs.

PR #837, opened 2019-08-16
Complement SCM URL with Scaladex

exoego:
Unfortuately, number of `pom.xml` without SCM URL are not small.
Therefore Scala Steward can not include URL/diff URL in PR for such artifacts.

Scaladex neatly know SCM URLs even if `pom.xml` lacks it. 
So let's use Scaladex as a second resource.
Original idea: https://github.com/fthomas/scala-steward/issues/82#issuecomment-429421560.

This is not perferct solution since Scaladex do not index non-Scala artifacts (e.g. aws-sdk-java), but definitely improve situations !!

# Notes

Only artifact id is used as search query, since the below Scaladex issues.
https://github.com/scalacenter/scaladex/issues/577
https://github.com/scalacenter/scaladex/issues/578

# Example PRs

| Artifact| Before (no URL or homepage) | After (have SCM URL from Scaladex) |
| --- | --- | --- |
| sbt-scalatra | https://github.com/exoego/scala-steward-test-project/pull/39 | https://github.com/exoego/scala-steward-test-project/pull/62 |
| sbt-scalajs-crossproject | https://github.com/exoego/scala-steward-test-project/pull/37 | https://github.com/exoego/scala-steward-test-project/pull/59 |
| scalatest| https://github.com/exoego/scala-steward-test-project/pull/56 | https://github.com/exoego/scala-steward-test-project/pull/61 |

PR #1418, opened 2020-05-02
Complement SCM URL with Scaladex

exoego:
Rework of #837

I could not find Scaladex API endpoint for finding repository, so scraping is still used ...


fthomas:
This is great, thanks @exoego!

I'm a bit concerned about repeated calls to `HttpScaladexClient.extractFromPage` with the same URL. It seems that the result of `client.get(url)` is not cached and repeated calls could slow Scala Steward down, is that right? It would be nice if the `Client` that this class is using would cache its results. The same `Client` could then also be used in `HttpExistenceClient`.

fthomas:
On a second thought, adding a general purpose caching client in this PR is too much to ask for. A specialized cache for the response body that is only used in `HttpScaladexClient` is just fine. I guess the implementation could be similar to the cache in `HttpExistenceClient`.

exoego:
I will revisit

PR #1429, opened 2020-05-12
Prepare support for other build systems

fthomas:
The changes here are based on @stanislav-chetvertkov's excellent work in #1193. Instead of full support for another build system, this PR only adds abstractions to make supporting other build systems easier. To add support for a new build system should now only require changes to `Context` and `BuildSystemDispatcher` in addition to code that interacts with that new build system. One motivation for doing this PR, is to make #1193 smaller and to make reviewing it easier.

This PR basically contains the `BuildSystemAlg` changes from #1193 but without associating a `Repo` with a build system upfront in `StewardAlg`. The new `BuildSystemDispatcher` is now checking on every `getDependencies` or `runMigrations` call which build system needs to be called. This allows to call more than one build system per repo and doesn't require too much changes to the existing code base. It just replaced `SbtAlg` with `BuildSystemDispatcher` in `RepoCacheAlg` and `EditAlg`.


fthomas:
> To add support for a new build system should now only require changes to Context and BuildSystemDispatcher in addition to code that interacts with that new build system.

To make it more concrete: Support for a new build system only requires implementing [`BuildSystemAlg`](https://github.com/fthomas/scala-steward/blob/de932f0af5b0168da5069a44b768ae0b28521aba/modules/core/src/main/scala/org/scalasteward/core/buildsystem/BuildSystemAlg.scala) and plugging it into `BuildSystemDispatcher` [here](https://github.com/fthomas/scala-steward/blob/de932f0af5b0168da5069a44b768ae0b28521aba/modules/core/src/main/scala/org/scalasteward/core/buildsystem/BuildSystemDispatcher.scala#L35).

PR #1474, opened 2020-06-08
Fix labeling for versions with a leading zero

sideeffffect:
Properly handle labeling of changes where where the version is like `0.x.y`

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/SemVer.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/SemVer.scala
@@ -64,11 +64,24 @@ object SemVer {
     case object BuildMetadata extends Change("build-metadata")
   }
 
-  def getChange(from: SemVer, to: SemVer): Option[Change] =
-    if (from.major =!= to.major) Some(Major)
+  def getChange(from: SemVer, to: SemVer): Option[Change] = {
+    val zero = NonNegBigInt.unsafeFrom(0)
+    if (
+      from.major === zero && to.major === zero &&
+      (from.minor =!= zero ||
+      to.minor =!= zero ||
+      from.patch =!= zero ||
+      to.patch =!= zero)
+    )
+      getChange(
+        from.copy(major = from.minor, minor = from.patch, patch = zero),
+        to.copy(major = to.minor, minor = to.patch, patch = zero)
+      )
+    else if (from.major =!= to.major) Some(Major)
     else if (from.minor =!= to.minor) Some(Minor)
     else if (from.preRelease =!= to.preRelease) Some(PreRelease)
     else if (from.patch =!= to.patch) Some(Patch)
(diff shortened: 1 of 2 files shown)


fthomas:
Thanks for the PR, @sideeffffect! Could you please motivate why this is a more proper way to handle zero major versions? Is it because of [SemVer §4](https://semver.org/#spec-item-4)
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

?

sideeffffect:
The concrete motivation are libraries like http4s or circe (and probably we could find many other examples). They are both still < 1, i.e. `0.x.y`. They break compatibility when `x` increases, but maintain compatibility when only `y` increases. Essentially they mimic _SemVer_ where `x` would be the major version and `y` would be the minor version.

We use the resulting labels from Scala Steward for automatic merging of the PRs: We are happy with automatically merging patch and minor version upgrades. But we are more careful with major version upgrades and want a human to do the merge, not a bot.

Currently, Scala Steward labels a change from `0.1.*` to `0.2.*` as a minor upgrade, which is misleading, because, as you say, things can (and often do) break. This PR intends to improve that.

fthomas:
Thanks @sideeffffect for providing the motivation for this change. I can understand that you don't want to automatically merge bumps of minor versions if the major version is zero but I'm not comfortable with reinterpreting minor versions as major versions and patch versions as minor versions. I think it would be strange if Scala Steward would label an update from `0.1.*` to `0.2.*` as `semver-major` but an update of from `1.1.*` to `1.2.*` as `semver-minor`. I can imagine that other people expect that updates with `semver-major` actually bump the major version and would be confused if it didn't.

I would be more comfortable with adding an additional `semver-zero-major` label if both versions have a zero major version. That way we wouldn't reinterpret the version fields and it would still be possible to prevent automatically merging `semver-minor` updates if they are also labelled with `semver-zero-major`.

WDYT?

sideeffffect:
> I think it would be strange if Scala Steward would label an update from `0.1.*` to `0.2.*` as `semver-major`

I don't think that they would be surprised at all, since that's the philosophy how the libraries they use are versioned. And I thought this would be the simplest and most straightforward solution for Scala Steward, so that' why I suggested it at the start. But we don't have to change how `semver-*` labels work, if you would prefer not to.

There may be other, maybe even better ways to do it. Maybe we can introduce a new kind(s) of label(s)?
 * `changes-breaking`: for changes like `0.0.1` -> `0.0.2`, `0.1.x` -> `0.2.y`, `1.x.y` -> `2.z.w`
 * `changes-nonbreaking`: for other changes

example:
```
labels: test-library-update, changes-nonbreaking, semver-minor
```

eed3si9n:
FYI I'm interested in giving a name to the `0.x.y`-is-major variant of Semantic Versioning in https://github.com/coursier/versions/issues/7, with the aim of library authors being able to self-identify. Please comment if you have ideas.

Issue #1470, opened 2020-06-05
Update only specified dependency (cli argument for updates.allow?)


MPV:
## What

Could Scala Steward create pull-requests for only a specific dependency (across multiple downstream repositories)?

## Why?

If a dependency (in our case, an internal private library) is updated, we'd like to get create pull requests for this library (and only that) in downstream consumers of this library.

## What I've tried:

We use Scala Steward through the [scala-steward-action](https://github.com/scala-steward-org/scala-steward-action), so it would be helpful to be able to pass this as a CLI argument.

Just setting [`updates.allow` in .scala-steward.conf](https://github.com/fthomas/scala-steward/blob/master/docs/repo-specific-configuration.md) wouldn't help since Scala Steward checks out the repos itself, where I can't modify that file locally per runthrough.

## Notes

I understand that changing this would be a bit of a shift in which purpose Scala Steward fulfills, 
from: 
`update all dependencies in this/these repos` (consumer-centric)
into:
`update this dependency in all my consumers` (producer/dependency-centric)

Curious to hear your thoughts on this.

fthomas:
@mzuehlke noted that #1483 could solve this issue and I agree with Marco that it looks like a nice solution for this request. What do you think, @MPV?

MPV:
Looks great, thanks @mzuehlke & @fthomas!

ches:
Part of our use case for #1483 is pretty much the same, it'd be good perhaps to give it as motivation on the PR but I'll echo it here as also a testimonial ☺️ 

We have a number of common libraries within the organization that are widely adopted. In order to help a small number of volunteer maintainers support them, Scala Steward keeps our little private ecosystem moving forward.

With downstream consumers that are application projects with large dependency graphs, though, the full firehose of Maven Central updates is too much churn and risk for most app teams to tolerate day-to-day. We're aiming to run an instance of Steward scoped to updating our blessed libraries, which is hopefully much easier to opt into and even mandate.

Steward can bring us a little closer to library upkeep benefits of a monorepo, without a monorepo.

fthomas:
Thanks for sharing your motivation for #1483, @ches, that's very helpful!

Issue #1411, opened 2020-04-23
Configurable path for `.scala-steward.conf`


anilkumarmyla:
Currently scala steward config is looked up at the top of repo `<repo>/.scala-steward.conf` and there is no way to specify an alternate location for it. 

Our use-case (as bizarre as it seems) is we want to have two different scala steward configs and run two instances of scala steward with the configs at different times/frequency. 

Some dependencies update too frequently and we want to be able to stagger the updates by having an ignore rule in one config and omitting it in the other. This way we can run the first instance as much as we want and occasionally run the second instance (with different config) whenever we seem to care about an update to this dependency. What do you think?

fthomas:
If you could set the PR frequency for individual dependencies, would that solve the underlying issue?

anilkumarmyla:
Yes! but in its current shape I cannot fathom how we can define pull-request frequency for a specific dependency

anilkumarmyla:
PR #1483 provides a way to achieve our use-case

PR #1483, opened 2020-06-15
Add default .scala-steward.conf for Scala Steward itself

eugeniyk:
In some scenarios there is a need to limit dependencies you want to update for all repos. 

Currently you'd need to update repo configuration (`.scala-steward.conf`) for every project you are working on.
If the repo doesn't have configuration (`maybeRepoConfig`) it will fallback to empty `RepoConfig.default`.

This PR is adding ability to override fallback to `RepoConfig.default` by reading `default.scala-steward.conf` within main scala steward repo. It gives you all the features as described in [repo-specific-configuration.md](https://github.com/fthomas/scala-steward/blob/master/docs/repo-specific-configuration.md)

Specifically, you can limit dependencies by providing `allow` settings (which is helpful if you have 100+ repos and you want to propagate specific dependencies, only)

### Changes
- Added `--default-repo-config` setting (default to `default.scala-steward.conf`)
- Chain of configuration: check repoConf (`$repo/.scala-steward.conf`) => fallback to DefaultRepoConf (`root/default.scala-steward.conf`) => fallback to empty configuration
- Added extra tests

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -236,7 +236,7 @@ runSteward := Def.taskDyn {
   val args = Seq(
     Seq("--workspace", s"$projectDir/workspace"),
     Seq("--repos-file", s"$projectDir/repos.md"),
-    Seq("--repos-default-config", s"$projectDir/default.scala-steward.conf"),
+    Seq("--default-repo-config", s"$projectDir/default.scala-steward.conf"),
     Seq("--git-author-email", s"me@$projectName.org"),
     Seq("--vcs-login", projectName),
     Seq("--git-ask-pass", s"$home/.github/askpass/$projectName.sh"),
--- a/build.sbt
+++ b/build.sbt
@@ -236,7 +236,7 @@ runSteward := Def.taskDyn {
   val args = Seq(
     Seq("--workspace", s"$projectDir/workspace"),
     Seq("--repos-file", s"$projectDir/repos.md"),
-    Seq("--repos-default-config", s"$projectDir/.scala-steward.conf"),
+    Seq("--repos-default-config", s"$projectDir/default.scala-steward.conf"),
     Seq("--git-author-email", s"me@$projectName.org"),
     Seq("--vcs-login", projectName),
     Seq("--git-ask-pass", s"$home/.github/askpass/$projectName.sh"),
--- a/build.sbt
+++ b/build.sbt
@@ -236,6 +236,7 @@ runSteward := Def.taskDyn {
   val args = Seq(
     Seq("--workspace", s"$projectDir/workspace"),
     Seq("--repos-file", s"$projectDir/repos.md"),
+    Seq("--repos-default-config", s"$projectDir/.scala-steward.conf"),
     Seq("--git-author-email", s"me@$projectName.org"),
     Seq("--vcs-login", projectName),
     Seq("--git-ask-pass", s"$home/.github/askpass/$projectName.sh"),
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -37,7 +37,7 @@ object Cli {
   final case class Args(
       workspace: String,
       reposFile: String,
-      reposDefaultConf: String,
+      reposDefaultConf: String = ".scala-steward.conf",
       gitAuthorName: String = "Scala Steward",
       gitAuthorEmail: String,
       vcsType: SupportedVCS = SupportedVCS.GitHub,
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -37,7 +37,7 @@ object Cli {
   final case class Args(
       workspace: String,
       reposFile: String,
-      reposDefaultConf: String = "default.scala-steward.conf",
+      defaultRepoConf: String = "default.scala-steward.conf",
       gitAuthorName: String = "Scala Steward",
       gitAuthorEmail: String,
       vcsType: SupportedVCS = SupportedVCS.GitHub,
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -37,6 +37,7 @@ object Cli {
   final case class Args(
       workspace: String,
       reposFile: String,
+      reposDefaultConf: String,
       gitAuthorName: String = "Scala Steward",
       gitAuthorEmail: String,
       vcsType: SupportedVCS = SupportedVCS.GitHub,
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -37,7 +37,7 @@ object Cli {
   final case class Args(
       workspace: String,
       reposFile: String,
-      reposDefaultConf: String = ".scala-steward.conf",
+      reposDefaultConf: String = "default.scala-steward.conf",
       gitAuthorName: String = "Scala Steward",
       gitAuthorEmail: String,
       vcsType: SupportedVCS = SupportedVCS.GitHub,
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
@@ -24,6 +24,7 @@ import org.scalasteward.core.application.Cli.EnvVar
 import org.scalasteward.core.git.Author
 import org.scalasteward.core.util
 import org.scalasteward.core.vcs.data.AuthenticatedUser
+
 import scala.concurrent.duration.FiniteDuration
 import scala.sys.process.Process
 
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
@@ -32,7 +32,7 @@ import scala.sys.process.Process
   * == [[reposDefaultConfigFile]] ==
   * Location of default repo configuration file.
   * This will be used if target repo doesn't have custom configuration.
-  * Note if this file doesn't exist, [[org.scalasteward.core.repoconfig.RepoConfig.default]] will be applied
+  * Note if this file doesn't exist, empty configuration will be applied
   *
   * == [[vcsApiHost]] ==
   * REST API v3 endpoints prefix
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
@@ -29,7 +29,7 @@ import scala.sys.process.Process
 
 /** Configuration for scala-steward.
   *
-  * == [[reposDefaultConfigFile]] ==
+  * == [[defaultRepoConfigFile]] ==
   * Location of default repo configuration file.
   * This will be used if target repo doesn't have custom configuration.
   * Note if this file doesn't exist, empty configuration will be applied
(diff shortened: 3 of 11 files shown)


mzuehlke:
I really like this idea and it will solve #1470.

Have you had a look how difficult it is to merge local configuration with global one ? 
So that in the case of an available local configuration it would override the global one ?

The only minor comment I have after going through the PR is that the default global scala-teward.conf you are referencing in the docs and examples start with a dot. Which would hide i. I would prefer "default.scala-teward.conf" or "global.scala-teward.conf"

eugeniyk:
@mzuehlke thanks for feedback

I think it could be nice continuation of this feature (I see some complexity with "mergeability" for some settings, see below)

Seems so far we have `allow` / `ignore` / `pin` options to be merged, rest should be just repo (local) over global:

1. **Allow** - I'd vote for **superset** since each project knows which the only repos it want to get, and each steward knows which dependencies it wants to operate.
In case of different combinations of `artifact`, `artifact / module`, `artifact / module / revision` we should pick the most restrictive one.
If one of Local / global settings are empty, just pick another one.

2. **Ignore** - union
3. **pin** - union with Local preference in terms of revision

At the same time, let me rename config to be `global.scala-steward.conf`, makes sense

mzuehlke:
Thanks for the detailed investigation. Sound feasible.
But I'm totally fine with doing this in a follow up PR.

eugeniyk:
@mzuehlke updated

mzuehlke:
Thanks @eugeniyk 
I have no possibility to approve or merge.
This is up to @fthomas

MPV:
Author of #1470 here. Looks like this will solve that issue. Just wanted to reach out to say thank you for this PR/contribution. ❤️

scripts/run.sh
```diff
@@ -23,6 +23,8 @@ LOGIN="scala-steward"
 
 COMMON_ARGS=(
   --workspace "$STEWARD_DIR/workspace"
+  --repos-file "$STEWARD_DIR/repos.md"
+  --repos-default-conf "$STEWARD_DIR/default.scala-steward.conf"
```
fthomas, line 27: Could you please revert these two changes. I'm using this script to run my public Scala Steward instance. The `--repos-file` parameter is set later in this file and my instance will not use a default repo config.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
```diff
@@ -16,34 +16,49 @@
...
-      val maybeRepoConfig = OptionT(fileAlg.readFile(configFile)).map(parseRepoConfig).flatMapF {
-        case Right(config)  => logger.info(s"Parsed $config").as(config.some)
-        case Left(errorMsg) => logger.info(errorMsg).as(none[RepoConfig])
-      }
-      maybeRepoConfig.value
+    workspaceAlg
+      .repoDir(repo)
+      .flatMap(projectRootDir => readConfiguration(projectRootDir / repoConfigBasename))
+
+  private def readConfiguration(configFile: File): F[Option[RepoConfig]] = {
```
fthomas, line 55: I would prefer if this is called `readFromFile` or `readRepoConfigFromFile`.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
```diff
@@ -31,7 +31,7 @@ final case class RepoConfig(
 }
 
 object RepoConfig {
-  val default: RepoConfig = RepoConfig()
+  private[core] val default: RepoConfig = RepoConfig()
```
fthomas, line 34: What is the purpose of this change? Is it because we now have two defaults, `RepConfig.default` and `RepoConfigAlg.defaultRepoConfig`, and that is confusing? Would it be less confusing if this is renamed to `RepoConfig.empty`?

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
```diff
@@ -16,34 +16,49 @@
...
 ) {
   def readRepoConfigOrDefault(repo: Repo): F[RepoConfig] =
-    readRepoConfig(repo).map(_.getOrElse(RepoConfig.default))
+    readRepoConfig(repo).flatMap { config =>
+      config.map(F.pure).getOrElse(defaultRepoConfig())
+    }
+
+  /**
+    * Default configuration will try to read file specified in config.reposDefaultConfig first; if not found - fallback to empty configuration.
+    * Note it's not lazy since we want to get config updates if you decide to change config in the middle of the process
```
fthomas, line 45: Could you please break lines after 80 characters?

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
```diff
@@ -16,34 +16,49 @@
...
-    workspaceAlg.repoDir(repo).flatMap { dir =>
-      val configFile = dir / repoConfigBasename
-      val maybeRepoConfig = OptionT(fileAlg.readFile(configFile)).map(parseRepoConfig).flatMapF {
-        case Right(config)  => logger.info(s"Parsed $config").as(config.some)
-        case Left(errorMsg) => logger.info(errorMsg).as(none[RepoConfig])
-      }
-      maybeRepoConfig.value
+    workspaceAlg
+      .repoDir(repo)
+      .flatMap(projectRootDir => readConfiguration(projectRootDir / repoConfigBasename))
```
fthomas, line 53: Could you rename `projectRootDir` to `repoDir` or just `dir`? I think these names are almost exclusively used in the code base.

modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
```diff
@@ -37,6 +37,7 @@ object Cli {
   final case class Args(
       workspace: String,
       reposFile: String,
+      reposDefaultConf: String = "default.scala-steward.conf",
```
fthomas, line 40: I would prefer calling this `defaultRepoConf`.

modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
```diff
@@ -49,6 +54,7 @@ import scala.sys.process.Process
 final case class Config(
     workspace: File,
     reposFile: File,
+    reposDefaultConfigFile: File,
```
fthomas, line 57: Likewise, I would prefer calling this `defaultRepoConfigFile`.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
```diff
@@ -16,34 +16,49 @@
...
-    readRepoConfig(repo).map(_.getOrElse(RepoConfig.default))
+    readRepoConfig(repo).flatMap { config =>
+      config.map(F.pure).getOrElse(defaultRepoConfig())
+    }
+
+  /**
+    * Default configuration will try to read file specified in config.reposDefaultConfig first; if not found - fallback to empty configuration.
+    * Note it's not lazy since we want to get config updates if you decide to change config in the middle of the process
+    */
+  def defaultRepoConfig(): F[RepoConfig] =
```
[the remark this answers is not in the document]
eugeniyk: > Note it's not lazy since we want to get config updates if you decide to change config in the middle of the process

I can move it to be val - just thinking if you have 100+ repos and you found an issue or smth, it will give you hot reload

wdyt?

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
eugeniyk: Just changed visibility, actually I want it to be used only by `RepoConfigAlg`, so I'll make it `private[repoconfig]`
`RepoConfig.empty` makes more sense for me, will rename

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
eugeniyk: No problem
Do you want to change https://github.com/fthomas/scala-steward/blob/master/.scalafmt.conf#L5 to be 80?

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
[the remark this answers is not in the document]
fthomas: Since the type is still `F[RepoConfig]` it will not be eagerly evaluated if this is changed to a `val`.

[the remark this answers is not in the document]
eugeniyk: you mean even if I leave it as is it won't be eager even if it's `def`?

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
fthomas: I thought this setting has no effect on Scaladocs?
We shouldn't change that setting in this PR.

eugeniyk: it has, otherwise I'd have it formatted by `fmt`

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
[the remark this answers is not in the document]
eugeniyk: ok set it to val for now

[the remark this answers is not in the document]
fthomas: I mean that it doesn't semantically matter if this is a `def` or `val` because the effects of that statement are not executed when it is evaluated but when the `F` is run.

modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala
```diff
@@ -64,10 +70,12 @@ final class PruningAlg[F[_]](implicit
       repoCache: RepoCache,
       dependencies: List[Scope.Dependency]
   ): F[(Boolean, List[Update.Single])] = {
-    val repoConfig = repoCache.maybeRepoConfig.getOrElse(RepoConfig.default)
+
     val depsWithoutResolvers = dependencies.map(_.value).distinct
     for {
       _ <- logger.info(s"Find updates for ${repo.show}")
+      defaultConfig <- repoConfigAlg.defaultRepoConfig
+      repoConfig = repoCache.maybeRepoConfig.getOrElse(defaultConfig)
```
fthomas, line 78: One last thing I noticed now: If `repoCache.maybeRepoConfig` is a `Some` there is no need to load the default repo config. We could change these two lines to:
```scala
repoConfig <- OptionT.fromOption[F](repoCache.maybeRepoConfig)
  .getOrElseF(repoConfigAlg.defaultRepoConfig)
```

eugeniyk: actually, in this case we should also merge configs
let me change

fthomas:
@eugeniyk Btw, you can fix the formatting by running `sbt fmt`.

eugeniyk:
I actually fix formatted version ;)

eugeniyk:
@fthomas you have quite advanced defence here: wart, fmt, codacy ;)

anilkumarmyla:
Wow, this PR also solves our use-case in https://github.com/fthomas/scala-steward/issues/1411 , this is awesome. Thanks @eugeniyk

build.sbt
```diff
@@ -236,6 +236,7 @@ runSteward := Def.taskDyn {
   val args = Seq(
     Seq("--workspace", s"$projectDir/workspace"),
     Seq("--repos-file", s"$projectDir/repos.md"),
+    Seq("--default-repo-config", s"$projectDir/default.scala-steward.conf"),
```
anilkumarmyla, line 239: should this be `--default-repo-conf` sans `ig` ?

fthomas: Yes, good catch. I'll update it.

PR #191, opened 2018-12-11
Do not leak file handles in deleteForce

fthomas:
better.files.File#delete() leaks file handles which can be observed
by watching `lsof | grep steward | wc -l` grow while steward runs.
With `FileUtils.forceDelete` the count of open files remains nearly
constant.

PR #1484, opened 2020-06-15
Fix deleteForce to also delete dangling symlinks

fthomas:


fthomas:
Btw, `FileUtils.forceDelete` is still used here because of #191 and a feeling that `forceDelete` works better with large directories.

PR #1488, opened 2020-06-17
Mergeable repo configurations

eugeniyk:
Continuation of #1483, allows configuration(s) to be merged together: 

Target Repo (`.scala-steward.conf`) => Default Scala Steward (`default.scala-steward.conf`) => Empty
by types: `Option[RepoConfig]` |+| `Option[RepoConfig]` |+| `RepoConfig.default`.

Current use-case - when running `scala-steward` on-prem with default global configuration, some repos want to override some settings (say, ignore few dependencies) by keeping global one as the base / fallback. Current behavior will pick the project's one and completely ignore default, global conf.

This PR allows you to have them both by merging conf together


modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
```diff
@@ -89,6 +93,109 @@ object UpdatesConfig {
   val defaultFileExtensions: Set[String] =
     Set(".scala", ".sbt", ".sbt.shared", ".sc", ".yml", "pom.xml")
 
+  private[repoconfig] val nonExistingFileExtension: List[String] = List(".non-exist")
```
eugeniyk, line 96: since current types are not distinguish between empty as all allowed and empty as nothing allowed - I've added artificial list

eugeniyk:
@fthomas @mzuehlke ready for initial review

mzuehlke:
Thanks @eugeniyk for the PR.
Overall is looks great, **but** I'm a bit uneasy with the _abuse_ of the `Semigroup` for merging 2 configs.
The `combine` operator of a `Semigroup` has to be associative and here the 1st config takes precedence over the 2nd.
Maybe it would be better to have a dedicated function for merging 2 configs.

eugeniyk:
@mzuehlke thanks for your comment, it's a good point!

If we take associativity as `(A -> B) -> C` === `A -> (B -> C)`,  it's exactly our case, pls correct me if I'm wrong: 

### Combine commits as message
(as example we can take `A` = None, `B` = Some("b"), `C` = Some("c"))

`(A getOrElse B) getOrElse C` === `A getOrElse (B getOrElse C)`

### Combine update rules
- updates.ignore => implemented as union, union is associative
- updates.pin => implemented as union with revision preference (if A, B, C are same group and artifactID, then (A -> B) -> C === A -> (B -> C) - A will take over B/C in any case)
- updates.allow => implemented as intersection, intersection is associative

More over, we can have it as Monoid since default config is `RepoConfig.empty` can be represented (proven by [test](https://github.com/fthomas/scala-steward/pull/1488/files#diff-cb8c36eafce5a612325e0294312df592R22)) as identity element:
- `RepoConfig.empty` |+| ourConf === ourConf
- ourConf |+| `RepoConfig.empty` === ourConf

mzuehlke:
> Did you mean `commutativity` property is it imply `A x B === B x A`?

@eugeniyk you exactly spotted my mistake. I mixed up commutativity and associativity.
Thanks for your detailed description.

eugeniyk:
How can I restart CI? on my local it's passing

fthomas:
Closing and reopening will trigger a new CI build.

modules/core/src/main/scala/org/scalasteward/core/util/wart.scala
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.scalasteward.core.util
+
+object wart {
+  @SuppressWarnings(Array("org.wartremover.warts.Equals"))
```
fthomas, line 20: We normally deal with this wart by adding a `cats.Eq` instance for the types we want to compare and then using `import cats.implicits._` at the site where the comparison happens so that we can use Cats' `===` or `=!=` syntax. See for example [here](https://github.com/scala-steward-org/scala-steward/blob/00fef541a845f232fb30154b372742c3852cde3d/modules/core/src/main/scala/org/scalasteward/core/git/GitAlg.scala#L169) and [here](https://github.com/scala-steward-org/scala-steward/blob/00fef541a845f232fb30154b372742c3852cde3d/modules/core/src/main/scala/org/scalasteward/core/git/Sha1.scala#L42).

modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
fthomas: Shouldn't we change the type of `fileExtensions` then? Does it even makes sense to have an empty list that is interpreted as "nothing allowed". That would mean Scala Steward can't edit any files. Who would want that?

modules/core/src/test/scala/org/scalasteward/core/repoconfig/CommitsConfigTest.scala
```diff
@@ -0,0 +1,20 @@
...
+    emptyCfg |+| emptyCfg shouldBe emptyCfg
+
+    val cfg = CommitsConfig(Some("message"))
+    cfg |+| emptyCfg shouldBe cfg
+    emptyCfg |+| cfg shouldBe cfg
+
+    val cfg2 = CommitsConfig(Some("message2"))
+    cfg |+| cfg2 shouldBe cfg
+    cfg2 |+| cfg shouldBe cfg2
+  }
```
fthomas, line 19: In a later PR we could replace this whole test with cats-laws' `MonoidTests` like this:
`checkAll("Monoid[CommitsConfig]", MonoidTests[CommitsConfig].monoid)`

modules/core/src/test/scala/org/scalasteward/core/repoconfig/PullRequestsConfigTest.scala
```diff
@@ -0,0 +1,20 @@
...
+    emptyCfg |+| emptyCfg shouldBe emptyCfg
+
+    val cfg = PullRequestsConfig(Some(PullRequestFrequency.Asap))
+    cfg |+| emptyCfg shouldBe cfg
+    emptyCfg |+| cfg shouldBe cfg
+
+    val cfg2 = PullRequestsConfig(Some(PullRequestFrequency.Weekly))
+    cfg |+| cfg2 shouldBe cfg
+    cfg2 |+| cfg shouldBe cfg2
+  }
```
fthomas, line 19: This test could also be replaced in a later PR by `checkAll("Monoid[PullRequestsConfig]", MonoidTests[PullRequestsConfig].monoid)`.

eugeniyk:
I'll come back to it very soon, it got some demand in my organization

modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
eugeniyk: hypothetically if repo owner says pls change only `.scala` and `Scala Steward` sets to change only `.sbt` - you shouldn't have changes, right?

modules/core/src/main/scala/org/scalasteward/core/util/wart.scala
eugeniyk: Thanks, changed

eugeniyk:
image

Ideally you want implicits to be smaller scope as possible
Codacy doesn't think so

eugeniyk:
@fthomas pls review again

modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
eugeniyk: About changing type - can we do it in another PR? I'm afraid it will bring a lot of changes to the code base since new structure will be
```scala
final case class UpdatesConfig(
    pin: Option[List[UpdatePattern]] = None,
    allow: Option[List[UpdatePattern]] = None,
    ignore: Option[List[UpdatePattern]] = None,
    limit: Option[PosInt] = None,
    includeScala: Option[Boolean] = None,
    fileExtensions: Option[List[String]] = None
)
```

ckipp01:
This is exactly what I created a feature request for without realizing this was here! Really looking forward to this!

eugeniyk:
@fthomas @mzuehlke let's make it real? It's been a while

mzuehlke:
Thanks @eugeniyk

PR #1523, opened 2020-07-03
use scalafixAll

bjaglin:
Use https://github.com/scalacenter/sbt-scalafix/pull/126, released in [sbt-scalafix 0.9.18](https://github.com/fthomas/scala-steward/pull/1522).

This slightly improves runtime as migrations can be run in parallel for `Compile` and `Test`.
    
Note that this will NOT run migrations on `IntegrationTest` or any other custom configuration, as Scalafix needs to be [enabled explicitly on them](https://github.com/scalacenter/scalafix/blob/v0.9.18/docs/users/installation.md#integration-tests).

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
@@ -92,8 +92,7 @@ object SbtAlg {
             scalafixCmds = for {
               migration <- migrations
               rule <- migration.rewriteRules
-              cmd <- Nel.of(scalafix, testScalafix)
-            } yield s"$cmd $rule"
+            } yield s"$scalafixAll $rule"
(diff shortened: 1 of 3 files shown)


bjaglin:
> Note that this will NOT run migrations on `IntegrationTest` or any other custom configuration, as Scalafix needs to be enabled explicitly on them.

Maybe slightly off topic, but I wonder if there is a demand for that in the context of scala-steward? If so, it could be interesting to provide a `scalafixEnableAll`.

fthomas:
> Maybe slightly off topic, but I wonder if there is a demand for that in the context of scala-steward? If so, it could be interesting to provide a scalafixEnableAll command.

Nobody has asked for this yet or complained about it but I guess it makes sense to have something like `scalafixEnableAll` so that Scala Steward migrates all sources in a project and not only the sources of the `Compile` and `Test` configuration.

bjaglin:
FTR https://github.com/scalacenter/scalafix/issues/1197

PR #1526, opened 2020-07-06
javac, curl and mill

joan38:


build.sbt
```diff
@@ -136,11 +136,7 @@ lazy val `sbt-plugin` = myCrossProject("sbt-plugin")
 
 lazy val `mill-plugin` = myCrossProject("mill-plugin")
   .settings(
-    crossScalaVersions := Seq(Scala213, Scala212),
```
fthomas, line 139: I think `crossScalaVersions` is required so that the mill plugin is published for Scala 2.12 and 2.13.

joan38: Removed by accident. Fixed now.

project/Dependencies.scala
```diff
@@ -1,6 +1,11 @@
 import sbt._
+import sbt.Keys._
 
 object Dependencies {
+  val mill = Def.setting {
+    val version = if (scalaBinaryVersion.value == "2.12") "0.6.3" else "0.7.4"
```
lefou, line 6: Just some food for thought: I think at compile time you should bind to the lowest API denominator, `0.6.0` resp. `0.7.0`. And only use newer versions at runtime. (Can't tell you if and how that is possible with sbt.)

joan38: Good point. We should do that in a separate PR I think.

Issue #1534, opened 2020-07-10
Support custom heuristics


payurgin:
WDYT about support of custom heuristics ?

Example repo config: 
```
...
...
updates.customHeuristics = [
  {
    groupId = "org.platform"
    termsToReplace = ["platform"]
    // artifactName = "some" //  optional artifact name
  }
]
```

So it would be possible to update such dependencies with scala steward:
```sbt
"org.platform" %% "libA"  % platformVersion
"org.platform" %% "libB"  % platformVersion  % Test,
```

payurgin:
@fthomas WDYT ?
I prepared a draft. Let me know if you ready to accept this contribution and I will fix all failed tests ASAP :)

fthomas:
I do not yet understand why the custom heuristic is needed. I think the `groupId` heuristic would be able to update `platformVersion` above.

payurgin:
Oh, you right, I made a mistake in the example above 🤦

Right example is: 
```sbt
"com.organization" % "libA" % platformVersion
"com.organization" % "libB" % platformVersion
```

It's not the best way to organize artifact but this is what we have.

fthomas:
Ok, so the issue is that the `val` used for the version does not contain a word of the groupId or artifactId so Scala Steward can't associate `platformVersion` with the version of `"com.organization" % "libA"` or `"com.organization" % "libB"`. I think that Scala Steward could automatically (without manual help) find and use `platformVersion` as search term for the version number. We could add a new heuristic between `moduleId` and `strict` that first looks for strings like `"com.organization" % "libA" % X` in all files and then uses `X` as sole search term if it is not the version number.

I think this strategy could also fix #1184.

WDYT @payurgin, would this work in your case?

payurgin:
I think it will work for my case.
I'm not sure if it possible to write regexp for matching variable name.
It's relatively simple for:
```sbt
"com.organization" % "libA" % platformVersion
"com.organization" % "libB" % platformVersion
```

But what to do with:
``` lazy val deps = Seq("foo" % "bar" % `platform version`,  "foo" % "bar" % `to be, or nø† †o ∫e`, "bar" % "foo" % "0.01")
```

Maybe we can start from simple cases when version variable name is relatively simple ? Then it would be relatively simple and I can add this to my pr.

fthomas:
I would also focus on the simple cases and ignore the esoteric ones. I could imagine that the naive approach works for 95% or more of the cases where a variable used for the version.

PR #1535, opened 2020-07-10
added custom heuristics

payurgin:
resolves https://github.com/fthomas/scala-steward/issues/1534


fthomas:
I'm closing this PR since `UpdateHeuristic` has been removed in #2816 and the new code that replaces version numbers is much better at finding the correct positions that need to updated so that there is less need to tweak this via configuration.

Please open a new issue if Scala Steward still has trouble finding and bumping the correct versions.

Issue #1023, opened 2019-10-05
Scala Steward could send WIP PRs for selected pre-releases


larsrh:
Precise mechanism unclear, but I could envision something like this:

* cats wants to release a bleeding-edge, experimental version 3.0.0
* to test out how much breaks in the ecosystem, maintainers release 3.0.0-M1
* scala-steward sends non-mergeable WIP PRs to projects bumping to the pre-release in order to have CI builds run
* maintainers keep a close eye on all the PRs (maybe there could be a dashboard, although that's out of scope here: they can also be selected through GitHub search)

exoego:
> maintainers keep a close eye on all the PRs (maybe there could be a dashboard, although that's out of scope here: they can also be selected through GitHub search)

This sounds related to #961

fthomas:
A first step towards this feature would be the ability to create regular PRs that bump versions to pre-releases. This is currently prevented by the implementation of `Version.selectNext`. So the first thing to do would be to have a variant of `selectNext` that does not exclude pre-release versions of a different version series.

PR #1549, opened 2020-07-26
Treat versions with hashes as pre-releases

### LINKED ISSUE #1260: Next version selection doesn't work for specific versions with hashes
Versions with commit hash containing `b` are not considered at all.

Here's a sample, modified from `VersionTest`'s `"selectNext, table 2"`:
```scala
val current = Version("1.2.0+9-4a769501")
val next = Version("1.2.0+17-7ef9b061")
current.selectNext(List(next)) shouldBe Some(next)
```

It seems that `b061` part gets parsed as `BETA` pre-release version, triggering `Do not select pre-release versions of a different series` case.

A quick workaround could be removing `b` clause in [Version.scala#L119](https://github.com/fthomas/scala-steward/blob/master/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala#L119)

fthomas:
This changes `Version.selectNext` in a few ways:
  * `isPreRelease` now returns `true`, if the version contains a hash that
    starts with `-` or `+`. So a `b` in the hash does not change the
    value of `isPreRelease` anymore.
  * Updates to a pre-release from a non-pre-release of the same series
    are explicitly prevented. 1.2.0 to 1.2.0+17-7ef98061 for example.
  * No more ad hoc rules that prevent updates where the new version
    contains a `-` or `+` when the previous version does not.
    1.2.3 to 1.2.3-4 for example are allowed now.
    These rules were added in #1134 and #1148.

This feels much more principled since hashes designate versions now as
pre-releases in addition to the other pre-release identifiers like
SNAPSHOT, MILESTONE, RC, and others.

Closes #1260.


fthomas:
Note that updates like https://github.com/almond-sh/almond/pull/620, which went from 2.1.4-13-fb16e4e to 2.2.0-0-fe5ed67, are prevented now since it is a bump to a pre-release of a different series. Updates like this could be allowed again once #1023 is implemented.

fthomas:
This resulted in a few new PRs that were created by my public instance. These are the ones I find most interesting:
* https://github.com/scala-steward-org/scala-steward/pull/1550
* https://github.com/scala-steward-org/scala-steward/pull/1551
* https://github.com/scalapb/ScalaPB/pull/885
* https://github.com/precog/sbt-precog/pull/72
* https://github.com/L-space/L-space-NS/pull/78
* https://github.com/djspiewak/gll-combinators/pull/61
* https://github.com/betagouv/aplus/pull/684
* https://github.com/softwaremill/correlator/pull/78

mzuehlke:
That PR looks really good and if I would have looked earlier I would have approved it. Easier to understand than before.
Some of the linked PRs look totally ok, some really crazy and some in between:

* from a stable version there should be not upgrade to a pre-release (like https://github.com/scala-steward-org/scala-steward/pull/1551: `0.8.0 to 0.8.0-1-d81662.`)
* a hash as the major version should maybe be ignored ? (https://github.com/scala-steward-org/scala-steward/pull/1550: `0.6.3 to 289f9e3aa3f5014a5c64319da8e6ab993947ade2-0-289f9e`, https://github.com/betagouv/aplus/pull/684: `2.9.3 to 26962ce-1`)
* maybe we should not "upgrade from a RC to a SNAPSHOT / hash ? (https://github.com/scalapb/ScalaPB/pull/885: `2.0.0-RC6-21 to 2.0.0-RC6-23+21-ffe6d4de-SNAPSHOT`,  https://github.com/softwaremill/correlator/pull/78: `0.21.0-RC5 to 0.21.6+43-2c1c1172-SNAPSHOT`)

The other PR you have linked look totally valid to me:
* https://github.com/precog/sbt-precog/pull/72: `0.1-58d9629 to 0.8.0`
* https://github.com/L-space/L-space-NS/pull/78: `3.8 to 3.9.4`
* https://github.com/djspiewak/gll-combinators/pull/61: `0.9-a3bf234 to 0.14-9419610`
* https://github.com/nafg/cloud-logging-logback-scala/pull/70: `0.116.0-alpha to 0.118.1-alpha`

fthomas:
> from a stable version there should be not upgrade to a pre-release (like #1551: 0.8.0 to 0.8.0-1-d81662.)

That could be "fixed" by recognizing hashes that are only 6 characters long.

> a hash as the major version should maybe be ignored ? (#1550: 0.6.3 to 289f9e3aa3f5014a5c64319da8e6ab993947ade2-0-289f9e, betagouv/aplus#684: 2.9.3 to 26962ce-1)

Recognizing 6 char hashes would have also prevented the first but not the second PR.

> maybe we should not "upgrade from a RC to a SNAPSHOT / hash ? (scalapb/ScalaPB#885: 2.0.0-RC6-21 to 2.0.0-RC6-23+21-ffe6d4de-SNAPSHOT, softwaremill/correlator#78: 0.21.0-RC5 to 0.21.6+43-2c1c1172-SNAPSHOT)

Yeah. I think the http4s version should have been updated to the latest non-pre-release (0.21.6). This needs more fiddling. :-)

> L-space/L-space-NS#78: 3.8 to 3.9.4

That surprised me because it should have happened long before this PR. Something was wrong before that prevented it.

> nafg/cloud-logging-logback-scala#70: 0.116.0-alpha to 0.118.1-alpha

That should have been prevented by the "Do not select pre-releases of different series" rule but there is a bug in `preReleaseIndex` that resulted in this PR. I'll push a fix for this shortly.

PR #1552, opened 2020-07-27
Fix Version.preReleaseIndex

fthomas:
... and recognize hashes which are only 6 characters long.

`preReleaseIndex` reported the wrong index for versions with components
with a length greater than one.

This is a follow-up #1549.


mzuehlke:
This seems to fix nicely the problems that remained after #1549 👍

fthomas:
Right, the only remaining problems are updates like these:
* scalapb/ScalaPB#885: 2.0.0-RC6-21 to 2.0.0-RC6-23+21-ffe6d4de-SNAPSHOT
* softwaremill/correlator#78: 0.21.0-RC5 to 0.21.6+43-2c1c1172-SNAPSHOT

PR #1557, opened 2020-07-31
Gitlab merge when pipeline succeeds

cactauz:
This adds a configurable option to use the gitlab API's merge automatically when pipeline succeeds functionality on MRs created by scala-steward.  

I realize this may be too specialized to be accepted, and that stuff like this may warrant a reconsideration around config.  I'm more than happy to help in any of these efforts to get this contributed, let us know what you think!

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
@@ -170,19 +170,24 @@ class Http4sGitLabApiAlg[F[_]](
             case mr if (mr.mergeStatus === GitlabMergeStatus.CanBeMerged) =>
               for {
                 _ <- logger.info(s"Setting ${mr.webUrl} to merge when pipeline succeeds")
-                res <- client.put[MergeRequestOut](
-                  url.mergeWhenPiplineSucceeds(repo, mr.iid),
-                  modify(repo)
-                )
+                res <-
+                  client
+                    .put[MergeRequestOut](
+                      url.mergeWhenPiplineSucceeds(repo, mr.iid),
+                      modify(repo)
+                    )
+                    // it's possible that our status changed from can be merged already,
+                    // so just handle it gracefully and proceed without setting auto merge.
+                    .recoverWith { case UnexpectedResponse(_, _, _, status, _) =>
+                      logger
+                        .warn(s"Unexpected gitlab response setting auto merge: $status")
+                        .flatMap(_ => mergeRequest)
+                    }
               } yield res
             case mr =>
               logger.info(s"Unable to automatically merge ${mr.webUrl}").map(_ => mr)
           }
-          .recoverWith { case UnexpectedResponse(_, _, _, status, _) =>
-            logger
-              .warn(s"Unexpected gitlab response setting auto merge: $status")
-              .flatMap(_ => mergeRequest)
-          }
+
     updatedMergeRequest.map(_.pullRequestOut)
   }
 
--- a/modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
@@ -17,6 +17,7 @@
 package org.scalasteward.core.gitlab.http4s
 
 import cats.syntax.all._
+import io.chrisdavenport.log4cats.Logger
 import io.circe._
 import io.circe.generic.semiauto._
 import org.http4s.{Request, Status, Uri}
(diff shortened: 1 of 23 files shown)


tyrcho:
I'd definitively use this if it is merged ! 

For the moment I resolved to this ugly workaround : https://gitlab.com/colisweb-open-source/scala-steward/-/blob/master/.gitlab-ci.yml#L31

modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
```diff
@@ -36,22 +37,33 @@ final private[http4s] case class MergeRequestPayload(
...
     MergeRequestPayload(id, data.title, data.body, projectId, data.head, data.base)
 }
+
 final private[http4s] case class MergeRequestOut(
     web_url: Uri,
     state: PullRequestState,
-    title: String
+    title: String,
+    iid: Int,
+    merge_status: String
```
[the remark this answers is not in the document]
cactauz: thanks for the review @tyrcho, i did it this way to agree with the other json field mappings in here and to minimize the number of changes here, but i think it is best to fix it up to be idiomatic and make the auto review happy (circe has an easy enough mechanism for this).  i'll take care of it and add some more tests i've been working on here shortly.

modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
```diff
@@ -116,15 +141,49 @@ class Http4sGitLabApiAlg[F[_]: MonadThrowable](
...
+                  modify(repo)
+                )
+              } yield res
+            case mr =>
+              logger.info(s"Unable to automatically merge ${mr.webUrl}").map(_ => mr)
+          }
+          .recoverWith { case UnexpectedResponse(_, _, _, status, _) =>
+            logger
+              .warn(s"Unexpected gitlab response setting auto merge: $status")
+              .flatMap(_ => mergeRequest)
```
fthomas, line 184: I'm not sure I understand this. `mergeRequest` has already been used on line 167 and if that fails, I don't think it will succeed on a second try here. If the first try succeeded and the merge request has already been created but `waitForMergeRequestStatus` or `client.put[MergeRequestOut](...)` throws an `UnexpectedResponse` using `mergeRequest` here again seems to be unnecessary - and I hope it will not create a second merge request. It makes more sense to me to only recover from `UnexpectedResponse` in the combination of `waitForMergeRequestStatus` and `client.put[MergeRequestOut](...)` (i.e. lines 168 to 180).

fthomas (commented):
This looks good to me, many thanks @cactauz! I've just one question about recovery in `createPullRequest` and if that is addressed I'm happy to merge.

modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
cactauz: Thanks for the review @fthomas! All of this happens after the merge request was created in attempt to just update the status, and what we encountered in practice is that very rarely we would get here with a status of can_be_merged, but the status changed in the moments between that successful response and the attempt to set the auto merge setting would result in an error (typically the 405 response documented here (https://docs.gitlab.com/ee/api/merge_requests.html#accept-mr).  I reproduced this scenario in one of the new tests, which fails without this block here.  

The mergeRequest is just returned here so we can proceed as normal having given up on the attempt to set the auto-merge on it.  Let me know if I can clarify it further!

cactauz: Just thinking about it a bit more, it might make it clearer to put this recoverWith directly on the client.put on L173 as that is the source of these potential error responses - happy to make that change.

modules/core/src/main/scala/org/scalasteward/core/gitlab/http4s/Http4sGitlabApiAlg.scala
```diff
@@ -116,15 +141,54 @@ class Http4sGitLabApiAlg[F[_]: MonadThrowable](
...
+                    .put[MergeRequestOut](
+                      url.mergeWhenPiplineSucceeds(repo, mr.iid),
+                      modify(repo)
+                    )
+                    // it's possible that our status changed from can be merged already,
+                    // so just handle it gracefully and proceed without setting auto merge.
+                    .recoverWith { case UnexpectedResponse(_, _, _, status, _) =>
+                      logger
+                        .warn(s"Unexpected gitlab response setting auto merge: $status")
+                        .flatMap(_ => mergeRequest)
```
fthomas, line 184: Flatmapping `mergeRequest` will POST a `MergeRequestPayload` again. Can't we just return the `MergeRequestOut` from `mergeRequest` in line 167? That is:
```suggestion
                        .as(mr)
```

cactauz: Ah yeah, my mistake, thanks!

fthomas (approved):
LGTM. Thanks again, @cactauz!

PR #1562, opened 2020-08-03
Add scalafix rules for upcoming 10.2.0 release

jrudolph:
Not yet release but will be soon, so it's good to have it ready before.


jrudolph:
Great, thanks for the quick turnaround.

PR #1584, opened 2020-08-31
Allow setting compiler options for Scalafix migrations

fthomas:
In anticipation to https://github.com/typelevel/cats/pull/3566


fthomas:
This PR allows Scalafix migrations to set additional options for the Scala compiler. If a migration specifies `scalacOptions`, a `scala-steward-scalafix-options.sbt` file will be temporarily added to the root of the repo during the execution of Scalafix. This file adds the options like this:
```sbt
ThisBuild / scalacOptions ++= List("$opt1", "$opt2", ...)
```
This should add the additional options to every sbt (sub)project in the build.

Since every migration can use different options, each migration in `runMigrations` is now executed in a separate sbt session. Before this change, all migrations passed to `runMigrations` were executed in the same sbt session.

PR #1598, opened 2020-09-03
Use only fun-suite and should-matcher for tests

mzuehlke:
With the release of scalatest 3.2.0 (https://www.scalatest.org/release_notes/3.2.0) the testing library got modularized making it possible to limit the possibilities and ensuring a consistent testing style in a project.
This converts the only different test file and tightens the scalatest dependency.


mzuehlke:
Thanks for the review

Issue #1586, opened 2020-09-01
Found 1 update ... no files were changed


tyrcho:
Hi,

Scala steward works great on most of our projects. However some dependency updates do not get their MR.

It seems to be in the case where the version for the update is "shared" between several libraries from the same group.

The 3 scala-opentracing libraries should upgrade at the same time.

Here an extract of the sbt build:
```scala
import sbt._

object Versions {
  val cats                = "2.1.1"
  val catsEffect          = "2.1.4"
  val circe               = "0.13.0"
  val tracing             = "2.4.1"
}

object CompileTimeDependencies {
  val cats                   = "org.typelevel"                               %% "cats-core"                             % Versions.cats
  val catsEffect             = "org.typelevel"                               %% "cats-effect"                           % Versions.catsEffect
  val circe                  = "io.circe"                                    %% "circe-core"                            % Versions.circe
  val circeGeneric           = "io.circe"                                    %% "circe-generic"                         % Versions.circe
  val circeRefined           = "io.circe"                                    %% "circe-refined"                         % Versions.circe
  val tracing                = "com.colisweb"                                %% "scala-opentracing-core"                % Versions.tracing
  val tracingContext         = "com.colisweb"                                %% "scala-opentracing-context"             % Versions.tracing
  val tracingHttp4s          = "com.colisweb"                                %% "scala-opentracing-http4s-server-tapir" % Versions.tracing
}

```

And the logs:

```
2020-09-01 05:20:27,479 INFO  ──────────── Nurture colisweb/back/ddd-packing-api ────────────
2020-09-01 05:20:27,479 INFO  Clone and synchronize colisweb/back/ddd-packing-api
2020-09-01 05:20:27,658 INFO  Parsed RepoConfig(CommitsConfig(None),PullRequestsConfig(None),UpdatesConfig(List(),List(),List(),None),Always)
2020-09-01 05:20:27,658 INFO  Found 1 update:
  com.colisweb:{(scala-opentracing-context, scala-opentracing-context_2.12), (scala-opentracing-core, scala-opentracing-core_2.12), (scala-opentracing-http4s-server-tapir, scala-opentracing-http4s-server-tapir_2.12)} : 2.4.1 -> 2.5.0
[...]

PR #1599, opened 2020-09-03
fix issue when value with version is a part of a long artifactId

tyrcho:
reproduce #1586

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
@@ -171,7 +171,7 @@ object UpdateHeuristic {
   val sliding = UpdateHeuristic(
     name = "sliding",
     replaceVersion = defaultReplaceVersion(
-      _.mainArtifactId.toSeq.sliding(5).map(_.unwrap).take(20).filterNot(_ === "scala").toList
+      _.mainArtifactId.toSeq.sliding(5).map(_.unwrap).filterNot(_ === "scala").toList
     )
   )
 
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala
```diff
@@ -171,7 +171,7 @@ object UpdateHeuristic {
   val sliding = UpdateHeuristic(
     name = "sliding",
     replaceVersion = defaultReplaceVersion(
-      _.mainArtifactId.toSeq.sliding(5).map(_.unwrap).take(5).filterNot(_ === "scala").toList
+      _.mainArtifactId.toSeq.sliding(5).map(_.unwrap).take(20).filterNot(_ === "scala").toList
```
fthomas, line 174: That's interesting! I'm wondering why `take(5)` was there in the first place? Maybe we can get rid of the `take` call completely.

tyrcho: that's right ! I've removed it. Probably someone got confused with sliding(5).

fthomas: It was there ever since this heuristic is in Scala Steward: https://github.com/scala-steward-org/scala-steward/pull/527. Unfortunately I didn't mention why I've added it. I guess I thought that only words at the beginning of artifacIds should be used as search terms for the version. But I'm OK with removing this restriction now.

fthomas:
You can run `sbt fmt` to fix the formatting.

fthomas (approved):
LGTM, thanks @tyrcho!

PR #1608, opened 2020-09-08
Fix Mill version and run ./mill if present

joan38:


fthomas:
We've stopped using local Mill scripts since #1527 and I'd like to keep it that way. Are there good reasons to prefer them over the systemwide installed Mill?

joan38:
We have custom coursier resolvers and they are provided by:
`export COURSIER_REPOSITORIES="ivy2Local|https://artifacts.mycompany.com"` that lives in the script.
Maybe anyway that's not ok that there is a different behavior between running the build with the script vs the system Mill. I'm rolling back this one.

fthomas:
Thanks!

Btw, this changes is required if `millVersion` is a hash version:
```
$ curl -I -s -o /dev/null -w "%{http_code}\n" https://github.com/lihaoyi/mill/releases/download/0.8.0-8-0a8d46/0.8.0-8-0a8d46
404
$ curl -I -s -o /dev/null -w "%{http_code}\n" https://github.com/lihaoyi/mill/releases/download/0.8.0/0.8.0-8-0a8d46
302
```

fthomas (approved):

Issue #1615, opened 2020-09-12
Order[Version] is not transitive


fthomas:
One build of https://github.com/scala-steward-org/scala-steward/pull/1614 failed with:
```
failing seed for order.transitivity is jrEI318o6GIPatr-KDWf_A1_sf9ndnP7IRRq2mNh1ZD=
[info] - Order[Version].order.transitivity *** FAILED ***
[info]   GeneratorDrivenPropertyCheckFailedException was thrown during property evaluation.
[info]    (VersionTest.scala:18)
[info]     Falsified after 1 successful property evaluations.
[info]     Location: (VersionTest.scala:18)
[info]     Occurred when passed generated values (
[info]       arg0 = Version(),
[info]       arg1 = Version(0),
[info]       arg2 = Version(D+)
[info]     )
[info]     Label of failing property:
[info]       Expected: true
[info]   Received: false
```
I assume the build will succeed once it is restarted.

fthomas:
This issue can be reproduced with the following test
```scala
  test("issue 1615: broken transitivity") {
    val v1 = Version("")
    val v2 = Version("0")
    val v3 = Version("X")
    v1 should be < v2
    v2 should be < v3
    v1 should be < v3
  }
```
which fails with
```
[info] - issue 1615: broken transitivity *** FAILED ***
[info]   Version() was not less than Version(X) (VersionTest.scala:26)
```

PR #1617, opened 2020-09-15
Fix transitivity of Order[Version]

fthomas:
This fixes the broken transitivity of `Order[Version]` noticed in #1615 by removing cases in `Order[Component]`. It is concerning that that there are no test cases that were sensitive to the removed code. We should double-check if it is OK to remove that code or if we can write tests that are sensitive to that code.


fthomas:
I added a test case that was sensitive to the removed code. `1.0e` is now smaller than `1.0.0` but was greater than it before. The tests also demonstrate that this behaviour is now in line with Coursier so I'm more confident with merging this.

I'd also like to note that the removed code was introduced by https://github.com/scala-steward-org/scala-steward/pull/1064 and https://github.com/scala-steward-org/scala-steward/pull/1217 and that the test cases of these PRs are still here and passing. This too increases my confidence that this change is OK.

Issue #119, opened 2018-10-31
Close PRs of obsolete updates


fthomas:
As suggested by @fwbrasil in https://github.com/getquill/quill/pull/1195#issuecomment-433567944.

kiranbayram:
Have you had a solution idea for this? @fthomas

mslinn:
This is a good idea. For example, the AWS OSGI Java library puts out a revision every few days, and they really pile up. The current queue of PRs for my [awslib_scala project](https://github.com/mslinn/awslib_scala) is  output by a [bash script that I wrote to apply PRs](https://gist.github.com/mslinn/1caab61a8222dfc25f6d9df5989af818) called `prs`:
```
     #47  Update sbt to 1.3.7
     #46  Update aws-java-sdk-osgi to 1.11.707
     #45  Update aws-java-sdk-osgi to 1.11.706
     #44  Update httpcore to 4.4.13
     #42  Update aws-java-sdk-osgi to 1.11.704
     #41  Update aws-java-sdk-osgi to 1.11.703
     #40  Update aws-java-sdk-osgi to 1.11.702
     #39  Update aws-java-sdk-osgi to 1.11.701
     #37  Update commons-codec to 1.14
```
Until Scala Steward incorporates this feature request, I think I'll modify my `prs` script to:
 - Scan the output of the `hub pr list` command
 - For each PR:
    - If obsolete, delete the PR, otherwise apply it

daddykotex:
Just an update, I'll probably tackle this.

This is something really interesting when running Scala Steward within an organization. My company has a lot of libraries in independent repositories and they're updated frequently. Along with this, when we run a build storm, it produces a lot of new libraries.

Overtime, if we're not disciplined and merge quickly they add up and it's hard to maintain.

This issue is also very interesting to me but the problems @fthomas raises are real, so I'll wait out for it.

https://github.com/scala-steward-org/scala-steward/issues/412

daddykotex:
I'm done with the implementation and most of the testing, I just waiting a new version of https://github.com/aws/aws-sdk-java/tags to test all the way :P

Ideally, it should close: https://gitlab.com/daddykotex/scala-steward-test/-/merge_requests/7

daddykotex:
It worked (but I'll be honest, on the second try)

```
[info] 2020-10-15 20:08:26,465 INFO
[info]   ____            _         ____  _                             _
[...]

PR #1631, opened 2020-09-30
Ground work to close obsolete pull requests

daddykotex:
This is some ground work to bring in #119.

Let me know what you think. In an upcoming MR:
* implement close pr for gitlab/github/bitbucket
* write the domain logic to extract branches, find related pr and close em

Here is the initial feature document I have locally:

# Introduction
Currently, the Scala Steward creates a pull-request with a specific branch pattern. I hope to reuse this branch pattern to find opened PR on branches that matches this pattern. 

Suppose the bot is working with `org.typelevel:cats`. If there is an opened PR on the project to update from `2.0.0` to `2.0.1` (branch: `update/cats-2.0.1`) and when the bot runs again, it opens a new PR to update from `2.0.0` to `2.0.2` (branch: `update/cats-2.0.2`), we'd like the bot to close the older PR.

This is to prevent PRs to pile up.

# Details

The pattern looks like this: `update/${artifact}-${version}`. If the various providers give us a way to get only opened branches that match the first part of the pattern: `update/${artifact}` (i.e.: ignoring the version)

So the bot needs to:

1. list branches in the repository (locally using `git`)
2. filter out the current branch, and keep only branches related to the current main artifact id we're updating
3. foreach branch, check whether a pull request is open for it, if yes, close it

# References:

Github Pulls API: https://docs.github.com/en/free-pro-team@latest/rest/reference/pulls

Gitlab MRs API: https://docs.gitlab.com/ee/api/merge_requests.html#list-merge-requests

Bitbucket pullrequests API https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pullrequests

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/bitbucket/http4s/json.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/bitbucket/http4s/json.scala
@@ -39,9 +39,10 @@ private[http4s] object json {
 
   implicit val pullRequestOutDecoder: Decoder[PullRequestOut] = Decoder.instance { c =>
     for {
+      id <- c.downField("id").as[Int]
       title <- c.downField("title").as[String]
       state <- c.downField("state").as[PullRequestState]
       html_url <- c.downField("links").downField("self").downField("href").as[Uri]
-    } yield (PullRequestOut(html_url, state, title))
+    } yield (PullRequestOut(id, html_url, state, title))
   }
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/bitbucketserver/http4s/Http4sBitbucketServerApiAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/bitbucketserver/http4s/Http4sBitbucketServerApiAlg.scala
@@ -59,7 +59,7 @@ class Http4sBitbucketServerApiAlg[F[_]](
         reviewers = reviewers
       )
       pr <- client.postWithBody[Json.PR, Json.NewPR](url.pullRequests(repo), req, modify(repo))
-    } yield PullRequestOut(pr.links("self").head.href, pr.state, pr.title)
+    } yield PullRequestOut(pr.id, pr.links("self").head.href, pr.state, pr.title)
   }
 
   private def useDefaultReviewers(repo: Repo): F[List[Reviewer]] =
@@ -92,7 +92,9 @@ class Http4sBitbucketServerApiAlg[F[_]](
   override def listPullRequests(repo: Repo, head: String, base: Branch): F[List[PullRequestOut]] =
     client
       .get[Json.Page[Json.PR]](url.listPullRequests(repo, s"refs/heads/$head"), modify(repo))
-      .map(_.values.map(pr => PullRequestOut(pr.links("self").head.href, pr.state, pr.title)))
+      .map(
+        _.values.map(pr => PullRequestOut(pr.id, pr.links("self").head.href, pr.state, pr.title))
+      )
 
   def ni(name: String): Nothing = throw new NotImplementedError(name)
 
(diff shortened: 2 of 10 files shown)


daddykotex:
Note: I'm used to GitLab and it deals better (IMO) with rebase and so I use that because I like a clean history and I think squash merge translate in a loss of information. But if you prefer that I avoid rebasing my commits, let me know.

fthomas:
I hadn't yet time to look into this in detail but wanted to comment on this aspect:

> I hope to reuse this branch pattern to find opened PR on branches that matches this pattern.

Created PRs are also persisted in `PullRequestRepository`. It could also be used to find old PRs that update the same artifacts.

daddykotex:
> PullRequestRepository

Oh wow, I was not aware this was a thing. I've been operating Scala Steward for a while and I did not know it was writing things to a JSON file!

I'll look into it that's for sure. I think this PR can still be relevant nonetheless.

daddykotex:
After looking a little bit more into that, there would be a problem. If this PR is merged, `JSON` file generated by the bot prior to the new version do not contain the `id: Int`, and thus deserializing is most likely to fail.

There are two approaches that I can think of:

- discard the existing file
- write a migration to extract the PR id

The first one speaks for itself but I'm not sure it's an acceptable solution (it would probably be in my case, but I'm not sure it's true for every operator).

The second one is more work, but I think it would be possible because most of `html_url` (the key in the json object) most likely contains the PR number (and I believe that's true for the 3 different supported system: Gitlab, Github and Bitbucket).

Do you have a preferred solution?

(note that I based my reasoning on reading the source, not by looking at a _real_ `JSON` file and I might be wrong)

mzuehlke:
If you can get the information out of the already stored data, that would be the best solution. 

Is the `id` part of the `html_url` for all VCS types and is this the information we need to close a PR for that VCS?

daddykotex:
> Is the id part of the html_url for all VCS types and is this the information we need to close a PR for that VCS?

My understanding is that it is.

Currently, the data in a file per repo (the key is the `Repo`), in directory like so (gitlab):
```
.
├── pull_requests
│   └── v2
│       └── gitlab
│           └── GROUP
│               ├── PROJECT1
│               │   └── pull_requests.json
│               └── PROJECT2
│                   └── pull_requests.json
```

The data in one of these files currently has this shape (gitlab):
```json
{
  "https://gitlab.com/group/project/-/merge_requests/363": {
    "baseSha1": "5aa2aeb03a46a2f71d06bfa234d533a8ad98ca84",
    "update": {
      "Single": {
        "crossDependency": [
          {
            "groupId": "org.scalactic",
            "artifactId": {
              "name": "scalactic",
              "maybeCrossName": "scalactic_2.13"
            },
            "version": "3.2.0",
            "sbtVersion": null,
            "scalaVersion": null,
            "configurations": "test"
          },
          {
            "groupId": "org.scalactic",
            "artifactId": {
              "name": "scalactic",
              "maybeCrossName": "scalactic_2.13"
            },
            "version": "3.2.0",
            "sbtVersion": null,
            "scalaVersion": null,
            "configurations": "test, it"
          }
        ],
        "newerVersions": [
          "3.2.2"
        ],
        "newerGroupId": null
      }
    },
    "state": "closed",
    "entryCreatedAt": 1602089471000
  }
}
```

The PullRequestsRepository use `Repo` has a key (as I said earlier) and stores a `Map[Uri, PullRequestData]` in the json. Note that `PullRequestData` is not `PullRequestOut`.

In this PR, I'm trying to update `PullRequestOut`. This change will not affect the content of the JSON file, so there is no migration to run.

But that raises the question: is this change necessary? Given that we have a file to keep track of all the PR opened/closed, we would not need the ID coming out of the request made to the VCS system when we create a PR.

I think we can close this PR. I will reopen a new one where I grab my information from the file and extract the `id` from the url (used as a key).

Issue #1624, opened 2020-09-24
Should not hard-code SBT home


avdv:
Currently, scala-steward silently assumes that SBT's home directory is `~/.sbt`.

There are various launcher options for SBT (configured system wide in `/etc/sbtops` or via `SBT_OPTS`) which relocate that directory.

I for one, would like to use the `--no-share` option which sets `-Dsbt.global.base=project/.sbtboot` `-Dsbt.boot.directory=project/.boot` `-Dsbt.ivy.home=project/.ivy`.

It would be good if scala-steward would honor those settings too, or at least document that you cannot relocate your SBT folder.

PR #1633, opened 2020-10-01
Note how to run Scala Steward on-prem with Gitlab

avdv:
I have set up Scala Steward within our team using a self-hosted Gitlab instance. Since the procedure was not at all straight forward I thought I'd share the CI config and setup instructions here.

On a second thought, I could also set up an example project on Gitlab.com, in case you don't want to have this here. But that would mean a bit more work for me... ;-)

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -126,7 +126,73 @@ docker run -v $PWD:/opt/scala-steward \
 
 [migrations]: https://github.com/fthomas/scala-steward/blob/master/docs/scalafix-migrations.md
 
-### Running On-premise (GitHub Enterprise)
+### Running On-premise
+
+#### GitHub Enterprise
 
 There is an article on how they run Scala Steward on-premise at Avast:
 * [Running Scala Steward On-premise](https://engineering.avast.io/running-scala-steward-on-premise)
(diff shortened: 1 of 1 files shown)


fthomas:
LGTM. Thanks, @avdv!

> Since the procedure was not at all straight forward I thought I'd share the CI config and setup instructions here.

Do you think there is anything we could do to make the setup easier?

avdv:
> Do you think there is anything we could do to make the setup easier?

The main complication with setting Scala Steward up with Gitlab CI was that it does only cache files inside the project folder. So I tried to configure sbt with `--no-share` option (sort of, I used absolute paths) which puts the global sbt directory, the ivy2 cache folder and the sbt boot directory all below the project directory. 
 
Alas, I was not aware how Scala Steward gets the project dependencies and that it assumes the global sbt directory is in `~/.sbt/1.0` (see #1624) and that it puts a file there which must be loaded by sbt.

(that's why I need to create the symlinks `~/.sbt` and `~/.ivy2` pointing into the project directory as a workaround)

If anything, I guess documenting these kinds of hidden requirements would be cool.

PR #1667, opened 2020-10-16
Close obsolete pull requests in the NurtureAlg

daddykotex:
I've only test that against GitLab as I said in: https://github.com/scala-steward-org/scala-steward/issues/119#issuecomment-709653048

All 4 server implement this with a PUT/PATCH to the pull request (merge request for Gitlab API) endpoint of the given pull request. There are some distinction in how the update in serialized but, all in all, it's quite similar:

```
https://docs.gitlab.com/ee/api/merge_requests.html#update-mr
GITLAB => {"state_event": "close"}

https://docs.github.com/en/free-pro-team@latest/rest/reference/pulls#update-a-pull-request
GITHUB => {"state": "closed"}


https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pullrequests/%7Bpull_request_id%7D#put
https://docs.atlassian.com/bitbucket-server/rest/6.6.1/bitbucket-rest.html#idp271
BITBUCKET => {"state": "declined"}
```

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
@@ -35,6 +35,8 @@ import org.scalasteward.core.vcs.data.{NewPullRequestData, Repo, RepoOut}
 import org.scalasteward.core.vcs.{VCSApiAlg, VCSExtraAlg, VCSRepoAlg}
 import org.scalasteward.core.{git, util, vcs}
 
+import scala.util.control.NonFatal
+
 final class NurtureAlg[F[_]](implicit
     config: Config,
     editAlg: EditAlg[F],
@@ -81,10 +83,11 @@ final class NurtureAlg[F[_]](implicit
       baseSha1 <- gitAlg.latestSha1(repo, baseBranch)
       _ <- NurtureAlg.processUpdates(
         sorted,
-        update =>
-          processUpdate(
+        update => {
+          val updateData =
             UpdateData(repo, fork, repoConfig, update, baseBranch, baseSha1, git.branchFor(update))
-          ),
+          processUpdate(updateData) <* closeObsoletePullRequests(updateData)
+        },
         repoConfig.updates.limit
       )
     } yield ()
@@ -112,6 +115,40 @@ final class NurtureAlg[F[_]](implicit
         )
       }
     } yield result
+
+  def closeObsoletePullRequests(data: UpdateData): F[Unit] = {
+    val crossDependency = data.update match {
+      case Update.Single(crossDependency, _, _) =>
+        crossDependency
+      case Update.Group(crossDependencies, _) =>
+        crossDependencies.head
+    }
+    for {
+      _ <- logger.info(s"Looking to close obsolete PRs for ${data.update.name}")
+      prsToClose <- pullRequestRepository.getObsoleteOpenPullRequests(
+        data.repo,
+        crossDependency,
+        data.update.nextVersion
+      )
+      dataPerId = prsToClose.flatMap { case (url, data) =>
+        vcsExtraAlg
+          .extractPRIdFromUrls(config.vcsType, url)
+          .tupleRight(data)
+          .toList
+      }
+      _ <- dataPerId.traverse { case (id, pullRequestData) =>
+        closePR(data.repo, pullRequestData, id).handleErrorWith { case NonFatal(ex) =>
+          logger.warn(ex)(s"Failed to close obsolete PR #$id for ${data.updateBranch.name}")
+        }
+      }
+
+    } yield ()
+  }
+
+  private def closePR(repo: Repo, prUpdate: Update, id: Int): F[Unit] =
+    vcsApiAlg.closePullRequest(repo, id) *> logger.info(
+      s"Closed a PR for ${prUpdate.show}"
+    )
 
   def applyNewUpdate(data: UpdateData): F[ProcessResult] =
     (editAlg.applyUpdate(
--- a/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
@@ -21,6 +21,7 @@ import cats.implicits._
 import eu.timepit.refined.types.numeric.PosInt
 import fs2.Stream
 import io.chrisdavenport.log4cats.Logger
+import org.http4s.Uri
 import org.scalasteward.core.application.Config
 import org.scalasteward.core.coursier.CoursierAlg
 import org.scalasteward.core.data.ProcessResult.{Ignored, Updated}
@@ -31,7 +32,7 @@ import org.scalasteward.core.repocache.RepoCacheRepository
 import org.scalasteward.core.repoconfig.{PullRequestUpdateStrategy, RepoConfigAlg}
 import org.scalasteward.core.scalafix.MigrationAlg
 import org.scalasteward.core.util.{BracketThrow, HttpExistenceClient}
-import org.scalasteward.core.vcs.data.{NewPullRequestData, Repo, RepoOut}
+import org.scalasteward.core.vcs.data.{NewPullRequestData, PullRequestState, Repo, RepoOut}
 import org.scalasteward.core.vcs.{VCSApiAlg, VCSExtraAlg, VCSRepoAlg}
 import org.scalasteward.core.{git, util, vcs}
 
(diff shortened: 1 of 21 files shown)


fthomas:
That looks great, thanks for the PR, @daddykotex! One thing I'm wondering is if it is possible to extract and persist the PR id from `PullRequestOut` when a PR is created instead of parsing it later from the PR URL.

daddykotex:
Yes, that's totally possible. In fact the work I initially did in: https://github.com/scala-steward-org/scala-steward/pull/1631 was exactly for that.

The ID is available in the returned payload when we create the PR and so grabbing it and serializing it to the file is very easy.

The problem is that since we have a state (the pull_requests.json file), we need to migrate what's in it. That'd mean going over the file, extracting the id from the key (URL) and adding it to the body.

Should I do that?

fthomas:
The migration of the `pull_requests.json` files is a good point. For closing obsolete PRs it is especially important to keep the old data around.

I think it would be nicer to persist PR IDs instead of extracting them from the URLs but I'd say we do this later in a separate PR.

fthomas:
Just tried this with one of my repos. After creating https://github.com/fthomas/kartograffel/pull/349, Scala Steward closed https://github.com/fthomas/kartograffel/pull/296, https://github.com/fthomas/kartograffel/pull/318, and https://github.com/fthomas/kartograffel/pull/342. Very nice!

modules/core/src/main/scala/org/scalasteward/core/nurture/PullRequestRepository.scala
```diff
@@ -55,6 +55,20 @@ final class PullRequestRepository[F[_]](
...
 
+  def getObsoleteOpenPullRequests(
+      repo: Repo,
+      crossDependency: CrossDependency,
+      newVersion: String
+  ): F[List[(Uri, Update)]] =
+    kvStore.get(repo).map {
+      _.getOrElse(Map.empty).collect {
+        case (url, data)
+            if UpdateAlg.isForSameArtifacts(data.update, crossDependency) &&
```
fthomas, line 66: I think we should also check `data.state === PullRequestState.Open` here.

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -113,6 +116,40 @@ final class NurtureAlg[F[_]](implicit
...
+        }
+      }
+
+    } yield ()
+  }
+
+  private def closePR(repo: Repo, prUpdate: Update, id: Int): F[Unit] =
+    vcsApiAlg.closePullRequest(repo, id) *> logger.info(
+      s"Closed a PR for ${prUpdate.show}"
+    )
```
fthomas, line 151: I think we should also set the `PullRequestState` to `Closed` in `PullRequestRepository` for the closed PR.

daddykotex: Ahh that's true I forgot to update the store, will do

modules/core/src/main/scala/org/scalasteward/core/nurture/PullRequestRepository.scala
daddykotex: absolutely, will do

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -113,6 +116,40 @@ final class NurtureAlg[F[_]](implicit
       }
     } yield result
 
+  def closeObsoletePullRequests(data: UpdateData): F[Unit] = {
+    val crossDependency = data.update match {
+      case Update.Single(crossDependency, _, _) =>
+        crossDependency
+      case Update.Group(crossDependencies, _) =>
+        crossDependencies.head
+    }
```
fthomas, line 125: I would move that inside the implementation of `getObsoleteOpenPullRequests`. That method should just take a `Repo` and an `Update` and return all obsolete PRs.

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -113,6 +116,40 @@ final class NurtureAlg[F[_]](implicit
...
+          logger.warn(ex)(s"Failed to close obsolete PR #$id for ${data.updateBranch.name}")
+        }
+      }
+
+    } yield ()
+  }
+
+  private def closePR(repo: Repo, prUpdate: Update, id: Int): F[Unit] =
+    vcsApiAlg.closePullRequest(repo, id) *> logger.info(
+      s"Closed a PR for ${prUpdate.show}"
```
fthomas, line 150: It would be nice to also log the URL of the closed PR here.

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
daddykotex: will do

modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
daddykotex: will do

daddykotex:
I've added a commit resolving your concerns:

```
Author: David Francoeur <***@gmail.com>
Date:   Fri Oct 16 12:10:33 2020 -0400

    Address PR comments

    * pass repo and update only to getObsoleteRequest
    * make sure only open PR are retrieved (with test)
    * log the url of the closed PR
    * change the state in the store
```

modules/core/src/main/scala/org/scalasteward/core/nurture/PullRequestRepository.scala
```diff
@@ -55,6 +66,28 @@ final class PullRequestRepository[F[_]](
...
+      case Update.Single(crossDependency, _, _) =>
+        crossDependency
+      case Update.Group(crossDependencies, _) =>
+        crossDependencies.head
+    }
+
+    kvStore.get(repo).map {
+      _.getOrElse(Map.empty).collect {
+        case (url, data)
+            if UpdateAlg.isForSameArtifacts(data.update, crossDependency) &&
```
fthomas, line 83: I'm wondering if it wouldn't be easier to add `Update.withNewerVersions(versions: Nel[String]): Update` and then to use that here instead of `UpdateAlg.isForSameArtifacts` to find older updates, i.e.  `data.update.withNewerVersions(updateData.newerVersions) === updateData`? For `Update.Group` it would also mean that only other PRs that update the exact same list of artifacts are found. Currently all older PRs that contain the first artifactId in `updateData` are found.

daddykotex: I introduced the changes you proposed, it makes the code cleaner.

Meanwhile I had a question, I implemented like so:
```
  def withNewerVersions(versions: Nel[String]): Update = this match {
    case s @ Single(_, _, _) =>
      s.copy(newerVersions = versions)
    case g @ Group(_, _) =>
      g.copy(newerVersions = versions)
  }
```
That works in practice but we loose some type information and I find it disappointing, maybe you have a better solution. The problem is that if we have some `val update: Update.Single = ???` and we do `val newUpdate = update.withNewerVersions(updateData.newerVersions)`, well `newUpdate` will have the type `Update` and not `Update.Single`

Would you know how to implement `withNewerVersions` without the loss of the precise type?

fthomas: Thanks for the change! I try to test it again tomorrow with a few more repos and will merge if nothing serious pops up.

Regarding your question. You could leave the implementation in `Update` abstract and override it in `Single` and `Group` with more precise return types:
```scala
class Single { override def withNewerVersions(versions: Nel[String]): Single = ??? }
class Group { override def withNewerVersions(versions: Nel[String]): Group = ??? }
```

daddykotex:
I tried that, but for if you add `def withNewerVersions` in Update, its
return type has to be `Update` and thus, at call site, you'll get back an
Update (the generic one).

Anyway, it was more of a general question.

Thanks for your support

fthomas (approved):

mslinn:
This works well, thanks! However, GitHub send email when a PR is closed, which is redundant for the issue addressed by this PR because the newly closed PR has been superseded by a new PR. Is there some way to cut the noise in half by suppressing the email that notifies of the closed PR?

daddykotex:
I'm not sure it's possible. By glancing at the doc, I don't see anything to allow us to do that when we update a PR.

I've seen the `active` attribute on some other attributes and gives some control over notifications, but nothing for the `pulls` API.

PR #1672, opened 2020-10-17
Close obsolete PRs only if a PR was created or updated

fthomas:
This closes obsolete PRs only if a new PR was created or if an existing
PR was updated. Sometimes `processUpdate` does not create a new PR
because it can't figure out where to bump the version number. I think
that in these cases Scala Steward should not close the PRs of older
updates.

Follow-up to #1667. /cc @daddykotex


daddykotex:
I had that... I removed because on the first run, the bot created the MR, and the close MR code failed (poor json encoding). When I fixed the issue, I reran the bot but it did nothing, exactly because of that. The MR already existed and did not need to be updated. I removed the check and reran to see if my fix worked!

thanks for adding it back

PR #1673, opened 2020-10-18
Run scalafmt when upgrading scalafmt (opt-out)

### LINKED ISSUE #1056: When upgrading scalafmt, run scalafmt on the codebase
I think it makes sense that when scalafmt gets upgraded, steward should run it on the codebase after update. I'd also suggest a configuration toggle that repositories can add to their steward config in git that will disable that behavior, for people who prefer formatter updates to be executed as-needed instead of all files at once

exoego:
Closes #1056

This PR adds `scalafmt.runAfterUpgrading` config, as opt-oput (default `true`).
If `true` and when `scalafmt-core` is being updated, Scala Steward will run `scalafmt` CLI and add a separate commit if dirty.

Example PR: https://github.com/exoego/scala-steward-test-project/pull/173

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/scalafmt/ScalafmtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/scalafmt/ScalafmtAlg.scala
@@ -62,7 +62,7 @@ object ScalafmtAlg {
               F.pure(false)
             }
           _ <- F.whenA(shouldRunScalafmt) {
-            processAlg.execSandboxed(Nel.of("sbt", ";scalafmtAll;scalafmtSbt"), repoDir)
+            processAlg.exec(Nel.of("scalafmt"), repoDir)
           }
         } yield ()
     }
--- a/modules/core/src/main/scala/org/scalasteward/core/scalafmt/ScalafmtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/scalafmt/ScalafmtAlg.scala
@@ -20,10 +20,14 @@ import cats.data.Nested
 import cats.syntax.all._
 import cats.{Functor, Monad}
 import org.scalasteward.core.data.{Dependency, Version}
-import org.scalasteward.core.io.{FileAlg, WorkspaceAlg}
+import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
+import org.scalasteward.core.repoconfig.RepoConfigAlg
+import org.scalasteward.core.util.Nel
 import org.scalasteward.core.vcs.data.Repo
 
 trait ScalafmtAlg[F[_]] {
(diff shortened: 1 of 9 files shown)


fthomas:
This looks great, many thanks for the PR, @exoego! What do you think about making this opt-out instead of opt-in. I think it would be useful for a lot of people if Scala Steward reformats without adding a `.scala-steward.conf` or changing the default.

modules/core/src/main/scala/org/scalasteward/core/scalafmt/ScalafmtAlg.scala
```diff
@@ -43,5 +49,21 @@ object ScalafmtAlg {
...
+          shouldRunScalafmt <-
+            if (mainArtifactId === scalafmtArtifactId) {
+              repoConfigAlg
+                .readRepoConfigWithDefault(repo)
+                .map(_.scalafmt.exists(_.runAfterUpgrading))
+            } else {
+              F.pure(false)
+            }
+          _ <- F.whenA(shouldRunScalafmt) {
+            processAlg.execSandboxed(Nel.of("sbt", ";scalafmtAll;scalafmtSbt"), repoDir)
```
fthomas, line 65: Have you thought about using the [CLI version of Scalafmt](https://scalameta.org/scalafmt/docs/installation.html#cli) here. I think it could be as easy as `processAlg.exec(Nel.of("scalafmt"), repoDir)`. I see two advantages of using the CLI. It is probably faster than using sbt and it would also work with other build tools than sbt.

mzuehlke:
Thanks @exoego, 
could add the configuration to  https://github.com/scala-steward-org/scala-steward/blob/master/docs/repo-specific-configuration.md#repository-specific-configuration

exoego:
> What do you think about making this opt-out instead of opt-in. I think it would be useful for a lot of people if Scala Steward reformats without adding a `.scala-steward.conf` or changing the default.

I also think most of users, including myself, desire opt-out.
I will update codes.

>  could add the configuration to https://github.com/scala-steward-org/scala-steward/blob/master/docs/repo-specific-configuration.md#repository-specific-configuration

I will do so.

mzuehlke (approved):

modules/core/src/main/scala/org/scalasteward/core/scalafmt/ScalafmtAlg.scala
exoego: Updated code to use `scalafmt` https://github.com/scala-steward-org/scala-steward/pull/1673/commits/061d2f9455878b23df2e622a0ca34ec597e8fd18

fthomas (approved):

mzuehlke:
With these merged changes a local available `scalafmt` binary is required, isn't it ? Does the default docker image include one ?

fthomas:
Yes and no. We should add it to the docker image.

PR #1686, opened 2020-10-20
Add ability to search for updates with different artifact id and / or group id

scottrice10:
This builds from the [custom group migrations](https://github.com/scala-steward-org/scala-steward/pull/1298) work. The author there, @kubukoz, mentioned wanting to "tackle changing the artifact name... in another PR", and this is my attempt to do that.

I do not know of open source Scala projects that could be used to test upgrading with a different artifact id, or updating both the group id and artifact id. Ideally, someone in the community would know of such examples so that they can be added to tests here.

At my company, we are using scala-steward and need to be able to update a privately hosted library like this:
```
"com.company" %% "old-artifact-name" % "1.0.0" ->  "com.company" %% "new-artifact-name" % "2.0.0"
```
so that is my primary purpose for this PR.

I opted to rename the cli argument from `--group-migrations` to `--artifact-migrations`, and keep the logic for updating either / both of the artifacts in one place, since I think that was simpler and easier to handle the "both" case. However, this is a backwards incompatible change, so if this is undesired, I could break the artifact id rename into a separate cli argument and logic in code.


docs/artifact-migrations.md
```diff
@@ -0,0 +1,32 @@
...
+In this example, scala-steward will look in the project for dependencies with the before group id, "com.geirsson", and 
+artifact id "sbt-scalafmt". If found, scala-steward will search for updates using the after group id, "org.scalameta", 
+artifact id "sbt-scalafmt", and version greater than the initial version, 2.0.0. If found, scala-steward will update
+to the after group id and latest version.
+
+The fields `groupIdBefore` and `artifactIdBefore` are optional. If just `groupIdBefore` is specified, as in the previous
+example, then only the group id will get renamed. If just `artifactIdBefore` is specified, then only the artifact id
+will get renamed. Specifying both `groupIdBefore` and `artifactIdBefore` will rename both.
+
+[migrations]: https://github.com/fthomas/scala-steward/blob/master/modules/core/src/main/resources/artifact-migrations.conf
```
fthomas, line 31:
```suggestion
[migrations]: https://github.com/scala-steward-org/scala-steward/blob/master/modules/core/src/main/resources/artifact-migrations.conf
```

docs/running.md
```diff
@@ -196,3 +195,7 @@ echo "${SCALA_STEWARD_TOKEN}"
 ```
 7. add the `repos.md` file 
 8. (*optional*) create a new schedule to trigger the pipeline on a daily/weekly basis
+
+
+[scalafixmigrations]: https://github.com/fthomas/scala-steward/blob/master/docs/scalafix-migrations.md
+[artifactmigrations]: https://github.com/fthomas/scala-steward/blob/master/docs/artifact-migrations.md
```
fthomas, line 201:
```suggestion
[scalafixmigrations]: https://github.com/scala-steward-org/scala-steward/blob/master/docs/scalafix-migrations.md
[artifactmigrations]: https://github.com/scala-steward-org/scala-steward/blob/master/docs/artifact-migrations.md
```

modules/core/src/test/scala/org/scalasteward/core/mock/MockContext.scala
```diff
@@ -75,8 +75,8 @@ object MockContext {
   implicit val filterAlg: FilterAlg[MockEff] = new FilterAlg[MockEff]
   implicit val versionsCache: VersionsCache[MockEff] =
     new VersionsCache[MockEff](config.cacheTtl, new JsonKeyValueStore("versions", "1"))
-  implicit val groupMigrations: GroupMigrations =
-    GroupMigrations.create[MockEff].runA(MockState.empty).unsafeRunSync()
+  implicit val groupMigrations: ArtifactMigrations =
```
fthomas, line 78:
```suggestion
  implicit val artifactMigrations: ArtifactMigrations =
```

modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
```diff
@@ -58,7 +58,7 @@ object Context {
       implicit0(fileAlg: FileAlg[F]) = FileAlg.create[F]
       implicit0(migrationAlg: MigrationAlg) <-
         Resource.liftF(new MigrationsLoader[F].loadAll(config.scalafixCfg).map(new MigrationAlg(_)))
-      implicit0(groupMigration: GroupMigrations) <- Resource.liftF(GroupMigrations.create[F])
+      implicit0(groupMigration: ArtifactMigrations) <- Resource.liftF(ArtifactMigrations.create[F])
```
fthomas, line 61:
```suggestion
      implicit0(artifactMigrations: ArtifactMigrations) <- Resource.liftF(ArtifactMigrations.create[F])
```

modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
```diff
@@ -27,7 +27,7 @@ import scala.concurrent.duration.FiniteDuration
 final class UpdateAlg[F[_]](implicit
     filterAlg: FilterAlg[F],
     versionsCache: VersionsCache[F],
-    groupMigrations: GroupMigrations,
+    groupMigrations: ArtifactMigrations,
```
fthomas, line 30:
```suggestion
    artifactMigrations: ArtifactMigrations,
```

modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
```diff
@@ -40,7 +40,7 @@ final class UpdateAlg[F[_]](implicit
       maybeNewerVersions = Nel.fromList(versions.filter(_ > current))
       maybeUpdate = maybeNewerVersions
         .map(vs => Update.Single(CrossDependency(dependency.value), vs.map(_.value)))
-        .orElse(groupMigrations.findUpdateWithNewerGroupId(dependency.value))
+        .orElse(groupMigrations.findUpdateWithRenamedArtifact(dependency.value))
```
fthomas, line 43:
```suggestion
        .orElse(artifactMigrations.findUpdateWithRenamedArtifact(dependency.value))
```

fthomas (commented):
This looks good to me, thanks @scottrice10! I agree with you that having both migrations in the same place is nicer than to having them separated.

I added a couple of suggested changes to rename the old `groupMigrations` name to `artifactMigrations`.

docs/artifact-migrations.md
scottrice10: I did a Find and Repace from https://github.com/fthomas/scala-steward -> https://github.com/scala-steward-org/scala-steward in all places.

modules/core/src/test/scala/org/scalasteward/core/mock/MockContext.scala
scottrice10: Updated as suggested.

modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
scottrice10: Updated as suggested.

modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
scottrice10: Updated as suggested.

modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
scottrice10: Updated as suggested.

fthomas (approved):

Issue #595, opened 2019-06-19
execSandboxed should not inherit environment variables


fthomas:
As discussed in https://github.com/fthomas/scala-steward/pull/337#discussion_r264052558.

Issue #1698, opened 2020-10-29
Error on sbt installed by coursier/cs


yaroot:
I switched to `cs install sbt` for some time and didn't notice this.

The error is very obscure (unrelated message), but after deleting workspace this error shows up

```
[error] Expected symbol
[error] Not a valid command: -
[error] Expected end of input.
[error] Expected '--'
[error] Expected 'debug'
[error] Expected 'info'
[error] Expected 'warn'
[error] Expected 'error'
[error] Expected 'addPluginSbtFile'
[error] -batch
[error]  ^
```

fthomas:
It seems that sbt installed by coursier does not support the [`-batch` and `-no-colors` options](https://github.com/scala-steward-org/scala-steward/blob/1abad4bea5c3f3d9a3dece02e6531a0b61e0359c/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala#L111). Can you find a way to disable the supershell and colors that works with sbt's own shell script and coursier's launcher?

yaroot:
I switch back to stock sbt for now

fthomas:
Using `JAVA_OPTS` works for me for sbt's own script and coursier's launcher:
```sh
JAVA_OPTS="-Dsbt.color=false -Dsbt.supershell=false $JAVA_OPTS" sbt compile
```
Relevant documentation:
 * https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html#Command+Line+Options
 * https://get-coursier.io/docs/cli-install.html#java-options

fthomas:
Btw, the `-no-colors` option is there since the first commit of this repo (https://github.com/scala-steward-org/scala-steward/commit/dc938740f9036bc053d69756837cc1ff46daf59b#diff-61a5e78c575411a38676310ef7cc129c86fd859018daa28f2447b39ec2ec6a1eR9) and `-batch` was added in https://github.com/scala-steward-org/scala-steward/commit/de3360fe9be26a41ab996c6bc94ae40a40f4b8f7.

PR #1703, opened 2020-10-30
Allow passing environment variables to execSandboxed

fthomas:
This allows to pass extra environment variables to `execSandboxed` in
preparation for a fix for #1698. It also changes how environment
variables are used if firejail is enabled. All environment variables
(those from the config and those passed to `execSandboxed`) are
now passed via the `--env` parameter of the firejail command instead of
injecting them into the environment. This has been suggested earlier in
https://github.com/scala-steward-org/scala-steward/pull/337#discussion_r264052558 by @mwz.


fthomas:
The last commit also clears the inherited environment if firejail is used. This closes #595.

PR #1705, opened 2020-11-01
Run sbt without -batch and -no-colors options

### LINKED ISSUE #1698: Error on sbt installed by coursier/cs

fthomas:
This runs sbt without the `-batch` and `-no-colors` command line
options since not all launchers support them, see #1698.

I did some tests and everything works as before and I also didn't see
any colors or color escape sequences in the log output when the log
level is set to `TRACE`. Maybe these options were redundant the whole
time. Let's test this.

Closes #1698.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
@@ -79,8 +79,9 @@ object SbtAlg {
       override def getDependencies(repo: Repo): F[List[Scope.Dependencies]] =
         for {
           repoDir <- workspaceAlg.repoDir(repo)
-          commands = List(setOffline, crossStewardDependencies, reloadPlugins, stewardDependencies)
-          lines <- exec(sbtCmd(commands), repoDir)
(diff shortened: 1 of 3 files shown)


fthomas:
> I did some tests and everything works as before and I also didn't see
any colors or color escape sequences in the log output when the log
level is set to TRACE. Maybe these options were redundant the whole
time.

Turned out that they weren't redundant. I did my initial tests with sbt 1.4 only and didn't observe colors or the supershell there. I've now tested projects with sbt 0.13.18 and 1.2.8 and observed colors and fancy terminal stuff there. Both can be disabled by setting the properties `-Dsbt.color=false -Dsbt.log.noformat=true -Dsbt.supershell=false`.

PR #1710, opened 2020-11-03
add scalacheck-1-14 → scalacheck-1-15 migration

larsrh:
This is the first artifactId migration in the config file, so I'm not 100% sure this syntax is correct.

### DIFF
--- a/modules/core/src/main/resources/artifact-migrations.conf
+++ b/modules/core/src/main/resources/artifact-migrations.conf
@@ -24,7 +24,7 @@ changes = [
     initialVersion = 0.10.0
   },
   {
-    groupIdBefore = org.scalatestplus
+    groupIdAfter = org.scalatestplus
     artifactIdBefore = scalacheck-1-14
     artifactIdAfter = scalacheck-1-15
     initialVersion = 3.2.2.0
--- a/modules/core/src/main/resources/artifact-migrations.conf
+++ b/modules/core/src/main/resources/artifact-migrations.conf
@@ -22,5 +22,11 @@ changes = [
     groupIdAfter = org.typelevel
     artifactIdAfter = kind-projector
     initialVersion = 0.10.0
+  },
+  {
+    groupIdBefore = org.scalatestplus
+    artifactIdBefore = scalacheck-1-14
+    artifactIdAfter = scalacheck-1-15
+    initialVersion = 3.2.2.0
   }
 ]


modules/core/src/main/resources/artifact-migrations.conf
```diff
@@ -22,5 +22,11 @@ changes = [
     groupIdAfter = org.typelevel
     artifactIdAfter = kind-projector
     initialVersion = 0.10.0
+  },
+  {
+    groupIdBefore = org.scalatestplus
```
mzuehlke, line 27:
```suggestion
    groupIdAfter = org.scalatestplus
```
It's the other way around: https://github.com/scala-steward-org/scala-steward/blob/master/docs/artifact-migrations.md
The `*After` fields are mandantory.

mzuehlke (changes requested):

mzuehlke (approved):

mzuehlke:
thanks @larsrh

larsrh:
This does not appear to trigger, e.g. in Paiges: https://github.com/typelevel/paiges

fthomas:
Here is an excerpt from the log for Paiges:
```
2020-11-09 19:56:10,956 INFO  Ignore org.scalatestplus:(scalacheck-1-14, scalacheck-1-14_2.12) : 3.2.2.0 -> 3.3.0.0-SNAP2 (reason: no suitable next version)
2020-11-09 19:56:10,956 INFO  Ignore org.scalatestplus:(scalacheck-1-14, scalacheck-1-14_2.13) : 3.2.2.0 -> 3.3.0.0-SNAP2 (reason: no suitable next version)
2020-11-09 19:56:10,956 INFO  Ignore org.scalatestplus:(scalacheck-1-14, scalacheck-1-14_sjs1_2.12) : 3.2.2.0 -> 3.3.0.0-SNAP2 (reason: no suitable next version)
2020-11-09 19:56:10,956 INFO  Ignore org.scalatestplus:(scalacheck-1-14, scalacheck-1-14_sjs1_2.13) : 3.2.2.0 -> 3.3.0.0-SNAP2 (reason: no suitable next version)
```
The migration added here is not used because there are newer versions of that artifact available. Scala Steward currently only rewrites groupIds or artifactIds if there are no newer versions for the given artifact.

larsrh:
In that case, I think we're out of luck, since upstream will probably continue to release new versions.

PR #1716, opened 2020-11-05
Mergify: configuration update

fthomas:
This change has been made by @fthomas from https://mergify.io config editor.


fthomas:
Hit the wrong button.

PR #1719, opened 2020-11-06
Change default value of --disable-sandbox

fthomas:
This changes the default value of the `--disable-sandbox` option from
`false` to `true`. I think this is a better default since sandboxed
execution is only useful when the operator of Scala Steward does not
trusts the repositories it is working on. This is only true for public
instances and (almost) never true for private instances. And there are
a lot more private than public instances.

Other arguments for this change are that `firejail` does not work inside
Docker and that documentation and setup become simpler.

To enable sandboxed execution, `--disable-sandbox=false` or the newly
added `--enable-sandbox` option can be used.


modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
```diff
@@ -118,7 +118,7 @@ object Config {
         sandboxCfg = SandboxCfg(
           whitelistedDirectories = args.whitelist,
           readOnlyDirectories = args.readOnly,
-          disableSandbox = args.disableSandbox
+          enableSandbox = args.enableSandbox.getOrElse(args.disableSandbox)
```
xuwei-k, line 121: 🤔 

```diff
- enableSandbox = args.enableSandbox.getOrElse(args.disableSandbox)
+ enableSandbox = args.enableSandbox.getOrElse(!args.disableSandbox)
```

fthomas: :facepalm: Of course! Thanks for noticing! If you want to, you can submit a PR changing this. Or I'll do it later.

PR #1749, opened 2020-11-16
Fix the runSteward command

daddykotex:
Minor issue with the command. With sbt 1.4+ it fails with some weird class not found.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -316,7 +316,7 @@ runSteward := Def.taskDyn {
     Seq("--whitelist", s"$home/.mill"),
     Seq("--whitelist", s"$home/.sbt")
   ).flatten.mkString(" ", " ", "")
-  (core.jvm / Compile / run).toTask(args)
+  (core.jvm / run).toTask(args)
 }.value
 
 /// commands


fthomas:
Strange, it works for me. Could you paste the error you're getting?

daddykotex:
I'm sorry, it's my mistake. I tried again this morning and it worked fine.

I was getting a NoClassDefFound with `org.scalasteward.core.application.SupportedVCS$GitLab`, I'm not sure why it kept happening. I will close, sorry for the noise.

ckipp01:
For what it's worth, I was working on some changes this morning and was hitting the same exact `NoClassDefFound` issue you mentioned @daddykotex. I then made your change and things started working again. After changing _back_ now, I also no longer have the issue. Making that changed definitely did something for me whether is was some cache issue or something else.

daddykotex:
Ahh good to know! I could not find any traces of it. I cleared my shell history, so I'm happy you can support what I say!

fthomas:
`GitLab` was recently renamed from `Gitlab`. Maybe the error is related to this change.

PR #1760, opened 2020-11-22
Added sbt-spiewak to the post-update hooks

djspiewak:
The tests are failing locally for bizarre reasons. It seems like the paths are not be canonicalized within the mock state machinery.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
@@ -59,8 +59,28 @@ final class HookExecutor[F[_]](implicit
 }
 
 object HookExecutor {
-  val postUpdateHooks: List[PostUpdateHook] =
-    List(
+
+  private[this] val knownGitHubActions = List(
+    (GroupId("com.codecommit"), ArtifactId("sbt-github-actions")),
+    (GroupId("com.codecommit"), ArtifactId("sbt-spiewak")),
+    (GroupId("com.codecommit"), ArtifactId("sbt-spiewak-sonatype")),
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
```diff
@@ -59,24 +59,36 @@ final class HookExecutor[F[_]](implicit
...
+
+  val postUpdateHooks: List[PostUpdateHook] = {
+    val actionsActions = knownGitHubActions.map { case (gid, aid) =>
+      PostUpdateHook(
+        groupId = gid,
+        artifactId = aid,
+        command = Nel.of("sbt", "githubWorkflowGenerate"),
+        useSandbox = true,
+        commitMessage =
+          update => s"Regenerate workflow with sbt-github-actions ${update.nextVersion}",
```
fthomas, line 78:
```suggestion
        commitMessage = _ => "Regenerate workflow with sbt-github-actions",
```
Including the version in the commit message could be confusing because `update.nextVersion` is the version of `aid` which is not necessarily the version of sbt-github-actions.

djspiewak: Ah good point.

fthomas (approved):

fthomas:
It works: https://github.com/typelevel/cats-effect/pull/1434 \o/
Seeing these non-trivial PRs that folks can review and merge in seconds instead of spending minutes for creating them make me unreasonably happy. :)

djspiewak:
> make me unreasonably happy. :)

Me too! Getting these kinds of features into scala-steward makes it all-but magical tbh. Between this and a nice Mergify configuration, it's almost like the Scala Ecosystem is one big loosely-coupled monorepo.

PR #1771, opened 2020-11-27
Fix tests on macOS

fthomas:


fthomas:
I guess this fixes the test failure mentioned in https://github.com/scala-steward-org/scala-steward/pull/1760#issue-525323620.

PR #1777, opened 2020-12-01
support pagination in installation/repositories and app/installations

xuwei-k:
- https://tools.ietf.org/html/rfc8288
- https://docs.github.com/en/free-pro-team@latest/rest/guides/traversing-with-pagination


fthomas (approved):
LGTM.

I guess the next evolution of the `all` method would be to return a `Stream[F, A]` instead of a `F[List[A]]`.

PR #1779, opened 2020-12-01
Ignore errors while cloning Git submodules

fthomas:
This PR replaces `--recursive` from the `git clone` options with an
[equivalent](https://git-scm.com/docs/git-clone/2.11.4#Documentation/git-clone.txt---recursive) `git submodule` call which we allow to fail. We allow it to fail
because submodules might point to private repos which may not be
necessary for updating dependencies.

This change is motivated by https://github.com/scala-steward-org/repos/pull/514#issuecomment-734213126.


Qi77Qi:
things are working now for this repo...thanks for the quick fix!

PR #1789, opened 2020-12-06
Add sbt-http4s-org to knownGitHubActions

rossabaker:
Per @fthomas's [helpful suggestion](https://github.com/http4s/http4s/pull/3981#issuecomment-739419625)


rossabaker:
Drat. That's my second trailing comma today that scalafmt took away. [stares coldly at defaults]

Thanks for the idea on this!

PR #1795, opened 2020-12-08
Minor refinement to `updates.limit` documentation

cjllanwarne:
Based on feedback https://github.com/scala-steward-org/scala-steward/issues/1788#issuecomment-740158243 -

We bumped into this recently, expecting `updates.limit` to be a global limit on how many PRs scala-steward would make, rather than how many it would make *per run*. Hopefully this wording change makes that distinction a little bit clearer.


fthomas (approved):
That sounds clearer. Many thanks, @cjllanwarne!

PR #1798, opened 2020-12-09
Version: ignore pre-release identifier after a hash

### LINKED ISSUE #1790: Dependencies with hash no longer updated
Some of our libraries are not being updated anymore starting version 0.7.0 and up. The last working version seems to be 0.6.0.

We have a hash in our release versions. Template: `x.y.z-yyyyMMdd-HHmmss-<git hash>`

example: 
`1.0.0-20201119-091040-d8b7496c`

fthomas:
With this tweak Scala Steward updates versions with dates and hashes
again.

Closes #1790.


mzuehlke (approved):

mzuehlke:
I had recently a look into the `selectNext` logic , which is getting more and more complex and thought if it would help to have `HASH` and `TIMESTAMP` as dedicated `Version.Component` sub-classes ?

fthomas:
It could make some things clearer but I think it also makes parsing components a little bit harder. I would say it is worth a try.

PR #1802, opened 2020-12-11
Simplify version parsing

mzuehlke:
Parsing of versions now uses regex.
In addition a new component for hashes is added to make it easier to detect them.

After the short discussion in https://github.com/scala-steward-org/scala-steward/pull/1798#issuecomment-741892096 I gave it a try.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
@@ -153,24 +153,24 @@ object Version {
     final case class Separator(c: Char, startIndex: Int) extends Component
     case object Empty extends Component { override def startIndex: Int = -1 }
 
-    private val numeric = """^([0-9]+).*""".r
-    private val separator = """^([.\-_+]).*""".r
-    private val alpha = """^([^.\-_+0-9]+).*""".r
-    private val hash = """([-+])(g?\p{XDigit}{6,}).*""".r
+    private val numeric = """^(\d+)(.*)$""".r
+    private val separator = """^([.\-_+])(.*)$""".r
+    private val alpha = """^([^.\-_+\d]+)(.*)$""".r
+    private val hash = """([-+])(g?\p{XDigit}{6,})(.*)$""".r
 
     def parse(str: String, index: Int = 0): List[Component] =
       str match {
         case "" => List.empty
-        case hash(sep, value) if !startsWithDate(value) =>
+        case hash(sep, value, rest) if !startsWithDate(value) =>
           Separator(sep.head, index) +: Hash(value, index + 1) +:
-            parse(str.substring(value.length + 1), index + value.length + 1)
-        case numeric(value)   => Numeric(value, index) +: rest(value, str, index)
-        case alpha(value)     => Alpha(value, index) +: rest(value, str, index)
-        case separator(value) => Separator(value.head, index) +: rest(value, str, index)
+            parse(rest, index + value.length + 1)
+        case numeric(value, rest) =>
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
```diff
@@ -157,61 +149,34 @@ object Version {
...
+    def parse(str: String, index: Int = 0): List[Component] =
+      str match {
+        case "" => List.empty
+        case hash(sep, value) if !startsWithDate(value) =>
+          Separator(sep.head, index) +: Hash(value, index + 1) +:
+            parse(str.substring(value.length + 1), index + value.length + 1)
+        case numeric(value)   => Numeric(value, index) +: rest(value, str, index)
+        case alpha(value)     => Alpha(value, index) +: rest(value, str, index)
+        case separator(value) => Separator(value.head, index) +: rest(value, str, index)
+      }
```
fthomas, line 170: This looks great @mzuehlke! I wonder if it could further simplified by replacing the `rest` function with capturing the rest in each regex, i.e. `^([0-9]+)(.*)` instead of `^([0-9]+).*`.  WDYT?

modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
```diff
@@ -157,61 +149,34 @@ object Version {
           case _                               => 0
         }
     }
+    final case class Hash(value: String, startIndex: Int) extends Component
```
fthomas, line 152: Another thing that came mind: Should we add a `case` to the `Order[Version]` instance for comparing `Hash` with `Hash` by comparing their values lexicographically so that versions which are equal up to the hashes are not considered equal?

modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
```diff
@@ -157,61 +149,34 @@ object Version {
...
+          Alpha(value, index) +: parse(rest, index + value.length)
+        case separator(value, rest) =>
+          Separator(value.head, index) +: parse(rest, index + value.length)
+      }
 
     def render(components: List[Component]): String =
       components.map {
         case n: Numeric   => n.value
         case a: Alpha     => a.value
+        case h: Hash      => h.value
```
fthomas, line 179: This should be covered by by the tests if #1804 is merged and `master` is merged into this branch.

modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
mzuehlke: For the comparison we are only comparing `alnumComponents` which does not contain `Hash`.
So a `case (h1: Hash, h2: Hash) => h1.value.compare(h2.value)` to `Order[Component]` doesn't get triggered. 
But if I get your idea correctly you want to handle this directly in `Order[Version]` ? Feel free to add what you think.

On the other hand does it make sense to compare hashes lexicographically, this does not have to have a relationship with the right order.

fthomas: I was somehow concerned that we consider two versions equal even if their string representations are different. But since we are using `alnumComponents` in `Order[Version]` this is already the case now. I think you're also right that comparing hashes lexicographically doesn't make much sense for our purpose.

fthomas (approved):
This is a great improvement. It makes the version parsing a lot clearer and easier to understand. Thanks @mzuehlke!

I pushed a small change to the `hash` regex and will merge if the build keeps green.

PR #1807, opened 2020-12-13
Remove startIndex from Version.Component

fthomas:
Recent changes (#1802) to `Version.selectNext` and `Version.Component` made it
unnecessary to track the `startIndex` of version components. There is no
point in keeping it if it isn't used, right?

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
@@ -18,7 +18,6 @@ package org.scalasteward.core.data
 
 import cats.Order
 import cats.implicits._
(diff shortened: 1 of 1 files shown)


mzuehlke:
Another good simplification 👍

fthomas:
For the record: `startIndex` was added in #1554 to
> simplify and optimize the `alnumComponentsWithoutPreRelease` and `preReleaseIndex` functions.

`preReleaseIndex` has been removed in this PR and `alnumComponentsWithoutPreRelease` is now implemented as:
```scala
    alnumComponents.takeWhile {
      case a: Version.Component.Alpha => !a.isPreReleaseIdent
      case _                          => true
    }
```
It seems to me that the same implementation could have also been used in #1554 too. :shrug:

PR #1829, opened 2020-12-22
Add Max Buffer Size to Config

LaurenceWarne:
Hi! This PR addresses a problem I found happens with mill projects with large `build.sc` files. If the number of json lines output by `mill extractDeps` is greater than the buffer size limit for `appendBounded`, the json is truncated and is then invalid, causing a `io.circe.ParsingFailure`.

Doubling the buffer size is enough for the project causing problems for me, but I added the buffer size as a config option which I hope is more palatable (size defaults to the existing) :slightly_smiling_face: .

Thanks, Laurence


fthomas (approved):
LGTM. I would also increase the default size to 8192. I guess you're not the only one hitting this issue.

fthomas:
@LaurenceWarne Out of curiosity, how many dependencies has this project where 4K lines are not enough?

LaurenceWarne:
> LGTM. Thanks @LaurenceWarne. I would also increase the default size to 8192. I guess you're not the only one hitting this issue.

Cool I've bumped it up.

> @LaurenceWarne Out of curiosity, how many dependencies has this project where 4K lines are not enough?

I think it was not the number of dependencies per say, but the number of modules we have, which each have their own test modules, etc 😄 The json file is 4.5K in all.

Issue #1793, opened 2020-12-08
Delete remote branch after closing a PR


jan-pieter:
Since #1667 Scala Steward closes obsolete PRs after creating a new or updating an existing PR. The remote branch is not deleted. I think it would be a good idea to delete the remote branch as well (or to make this configurable).

fthomas:
Sounds like a good idea.

PR #1846, opened 2020-12-28
Remove branch if PR is closed by project maintainer

fthomas:
Removes the remote update branch if we detect that a PR has been closed without merging.

See also #1793.


mzuehlke (approved):
Another cleanup, but this removal happens when the PR gets merged, too. Which is totally fine in my opinion.

fthomas:
I don't think it will be removed if the PR is merged. Merging the PR means the dependency is no longer outdated so that the same  `Update` won't be passed to `NurtureAlg` on the next run and the branch won't be removed. I think the code added here is only triggered if a PR is closed without merging and there are new commits on the base branch.

fthomas:
Btw, an example PR whose branch has been removed by this change is https://github.com/scala-steward-org/test-repo-2/pull/9.

mzuehlke:
The simplest approach would be: 
```
git("push", "origin", "--delete", branch.name)(repo).attempt.void
```

If we want to preserve that problem in the logs:
```
logger.attemptLogError("Removing remote branch failed") {
  git("push", "origin", "--delete", branch.name)(repo)
} .void
```

PR #1847, opened 2020-12-29
Show all details in the PR body below each other

fthomas:
This puts `oldVersionDetails` directly above the `ignoreFutureUpdates`
details instead of directly under the release releated URLs.

https://github.com/scala-steward-org/test-repo-2/pull/14#issuecomment-751992453
shows how this looks in comparison the current arrangement. IMO it is
nicer if all details sections are below each other. On the other hand,
`oldVersionDetails` is not as easily missed if it is directly under
the URLs.

WDYT @mzuehlke, do you have a preference for one or the other arrangement?


mzuehlke (approved):
I had no preference and my thoughts had been along your lines:

At the top is is more visible and the bottom all details are close to each other. But as this note is a bit more urgent than the other used the position further up.

Let's give nicer layout a try 👍

PR #1850, opened 2020-12-30
Add minimal CLI help

mzuehlke:
A first attempt at documenting all CLI arguments.
I'm open for other formatting ideas. 
Sadly [casa-app](https://github.com/alexarchambault/case-app) does not provide a markdown help on its own.

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -1,0 +1,39 @@
+# CLI help
+
+All command line arguments for the `scala-steward` application.
+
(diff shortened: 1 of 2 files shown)


fthomas (approved):
This looks great!

It would be indeed awesome if the documentation could be generated from the same code that generates the `--help` or `--usage` output but I'm not aware of a library that is capable of doing that.

PR #1875, opened 2021-01-11
Feature multi project support

GrafBlutwurst:
This is a much simpler attemt to get support for multiple projects within a mono repo based on the feedback here:

https://github.com/scala-steward-org/scala-steward/pull/1093#issuecomment-692246532

Currently it's still missing tests since I haven't had a good idea how to do those without unnecessary codebloat.


modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/MavenAlg.scala
```diff
@@ -37,19 +37,29 @@ object MavenAlg {
...
   ): MavenAlg[F] =
     new MavenAlg[F] {
-      override def containsBuild(repo: Repo): F[Boolean] =
-        workspaceAlg.repoDir(repo).flatMap(repoDir => fileAlg.isRegularFile(repoDir / "pom.xml"))
+      override def containsBuild(buildRoot: BuildRoot): F[Boolean] =
+        workspaceAlg
+          .repoDir(buildRoot.repo)
+          .flatMap(repoDir =>
+            fileAlg.isRegularFile(repoDir / buildRoot.relativeBuildRootPath / "pom.xml")
+          )
```
fthomas, line 45: @GrafBlutwurst Could you add a new `buildDir` or `buildRootDir` method to `WorkspaceAlg` that is implemented like this
```scala
def buildDir(buildRoot: BuildRoot): F[File] =
  repoDir(buildRoot.repo).map(_ / buildRoot.relativeBuildRootPath)
```
With that method, a lot of the changes in this PR become two simple replacements: `Repo` -> `BuildRoot` and `workspaceAlg.repoDir` -> `workspaceAlg.buildDir`.

modules/core/src/main/scala/org/scalasteward/core/vcs/data/BuildRoot.scala
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.scalasteward.core.vcs.data
+
+final case class BuildRoot(repo: Repo, relativeBuildRootPath: String)
```
fthomas, line 19: I suggest to rename `relativeBuildRootPath` to just `relativePath`. I don't think repeating the class name there adds more information.

modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/MavenAlg.scala
GrafBlutwurst: Done

modules/core/src/main/scala/org/scalasteward/core/vcs/data/BuildRoot.scala
GrafBlutwurst: Done

modules/core/src/main/scala/org/scalasteward/core/buildtool/BuildToolDispatcher.scala
```diff
@@ -25,44 +25,73 @@ import org.scalasteward.core.data.{Resolver, Scope}
...
 import org.scalasteward.core.util.Nel
-import org.scalasteward.core.vcs.data.Repo
+import org.scalasteward.core.vcs.data.{BuildRoot, Repo}
+import org.scalasteward.core.repoconfig.RepoConfigAlg
 
-trait BuildToolDispatcher[F[_]] extends BuildToolAlg[F]
+trait BuildToolDispatcher[F[_]] extends BuildToolAlg[F] {
+  def runMigrationsForAllBuildRoots(repo: Repo, migrations: Nel[Migration]): F[Unit]
+  def getDependenciesForAllBuildRoots(repo: Repo): F[List[Scope.Dependencies]]
+}
```
fthomas, line 34: I find it unfortunate that `BuildToolDispatcher` has now methods that work with `Repo` and `BuildRoot` but the only methods used outside of `BuildToolDispatcher` are the ones with `Repo`. Have you tried to add a second type parameter to `BuildToolAlg` as mentioned in [my comment](https://github.com/scala-steward-org/scala-steward/pull/1093#issuecomment-692246532):

> BuildToolDispatcher and the build tool traits like SbtAlg inherit from BuildToolAlg which uses Repo for its methods. Since SbtAlg, MavenAlg, and MillAlg should now operate on BuildRoots instead of Repo but BuildToolDispatcher should still use Repo we need to add a new type parameter to the BuildToolAlg, i.e. BuildToolAlg[F, R] where R would be either a Repo or a BuildRoot.

?

I think with this change, the above additions aren't necessary and the `BuildToolDispatcher`'s callsites (in `EditAlg` and `RepoCacheAlg`) also do not need to be changed.

GrafBlutwurst: Right that might work, I totally didn't see that part of the comment somehow. I'll change this asap.

GrafBlutwurst: Switched over, yeah that is a much nicer approach. Technically it should be possible to have contravariant typeclasses on `BuildToolAlg` that would allow the dispatcher to work easily both for `Repo` and `BuildRoot` I didn't do that since I'm a bit pressed for time now but if you see any value in that I'll happily change it to that.

fthomas: Looks good to me now, thanks. I think there is no need for more changes since the dispatcher is only used with `Repo` for now.

fthomas (approved):
LGTM. Thanks @GrafBlutwurst!

PR #1878, opened 2021-01-12
Create commit for Scalafix migration even if it failed

fthomas:
https://github.com/johnynek/bosatsu/pull/648 contains changes from a
Scalafix migration that exited with a non-zero status code in the same
commit as the version bump. Migration changes and the version bump
are supposed to be in separate commits but weren't because we're
currently only creating the Scalafix commit if it exited without an
error. This change always commits after a Scalafix migration regardless
of its exit status.


mzuehlke:
What about having a note in the commit message of the Scalafix commit that the migrations aborted with an error ?

fthomas:
Good idea!

Issue #1890, opened 2021-01-18
Cannot run program "scalafmt" once for the workspace, once in case of a monorepo


robstoll:
I get the following error when running scala-steward (using 26591b84)
```
java.io.IOException: Cannot run program "scalafmt" (in directory "/builds/tegonal/scala-steward/data/workspace"): error=2, No such file or directory
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1128)
	at java.base/java.lang.ProcessBuilder.start(ProcessBuilder.java:1071)
	at org.scalasteward.core.io.process$.$anonfun$createProcess$1(process.scala:82)
	at flatMap @ org.scalasteward.core.io.process$.slurp(process.scala:44)
	at ifM$extension @ io.chrisdavenport.log4cats.slf4j.internal.Slf4jLoggerInternal$Slf4jLogger.info(Slf4jLoggerInternal.scala:76)
	at >>$extension @ org.scalasteward.core.scalafix.MigrationsLoader.loadMigrations(MigrationsLoader.scala:43)
	at flatMap @ org.scalasteward.core.scalafmt.ScalafmtAlg$$anon$1.version(ScalafmtAlg.scala:55)
	at map @ org.scalasteward.core.scalafmt.ScalafmtAlg$$anon$1.version(ScalafmtAlg.scala:56)
	at flatMap @ org.scalasteward.core.application.SelfCheckAlg.checkScalafmtBinary(SelfCheckAlg.scala:49)
	at flatMap @ org.scalasteward.core.application.SelfCheckAlg.$anonfun$checkAll$3(SelfCheckAlg.scala:38)
	at ifM$extension @ io.chrisdavenport.log4cats.slf4j.internal.Slf4jLoggerInternal$Slf4jLogger.info(Slf4jLoggerInternal.scala:76)
	at timeoutTo @ org.scalasteward.core.io.process$.$anonfun$slurp$3(process.scala:67)
	at flatMap @ org.scalasteward.core.io.process$.$anonfun$slurp$1(process.scala:45)
	at flatMap @ org.scalasteward.core.io.process$.slurp(process.scala:44)
	at ifM$extension @ io.chrisdavenport.log4cats.slf4j.internal.Slf4jLoggerInternal$Slf4jLogger.info(Slf4jLoggerInternal.scala:76)
	at >>$extension @ org.scalasteward.core.scalafix.MigrationsLoader.loadMigrations(MigrationsLoader.scala:43)
[...]

PR #1897, opened 2021-01-19
improve error message in case git or scalafmt is not on the path

robstoll:
incorporates the feedback of #1890


fthomas (approved):
LGTM. Thanks @robstoll!

PR #1899, opened 2021-01-22
Add config key for default maven repo and self check url

meier-christoph:
This PR adds a few additional arguments to cli and config for the default maven repo and the self check uri.

For a full offline run add :
```
--self-check-uri https://github.example.org \
--default-maven-repo https://nexus.example.org \
--scalafix-migrations /data/config/scalafix-migrations.conf \
--artifact-migrations /data/config/artifact-migrations.conf \
--disable-default-scalafix-migrations
```

We are using scala-steward in a closed environment with github enterprise and are currently running a fork of scala-steward v0.8, this is a cherry pick from our fork as we would like to switch over to the github apps variant asap


fthomas:
I've two questions. Why would you want to override the URL for `UrlChecker` and how do you override the default Maven repository in other tools like Coursier and sbt? I could imagine that if you're already overriding the default Maven repository for Coursier, maybe we could always query it from Coursier so that there would be no need to override it in Scala Steward too.

meier-christoph:
We have our own github enterprise running on-prem in an isolated network. The self check would just fail and timeout after a few seconds since github.com is not reachable.

We only use sbt (which is why I haven't touched mill and I am not very familiar with it) and you can actually override the repositories that are used via `SBT_OPTS` e.g. `-Dsbt.repository.config=<file with urls to enterprise repo>` and `-Dsbt.override.build.repos=true`.

If you are using coursier via cli you can also set the repository using args or as environment variable but we don't use coursier outside of sbt v1.

To be honnest I wasn't sure what scala-steward was using the default maven repo for. There was a lot of trial and error befor it worked. For example, I had to build a new container and pre-build scalafmt in standalone mode as by default it would try to fetch some dependencies on the first run via coursier.

fthomas:
Thanks for the explanation, @meier-christoph. I'm OK with merging this but could please change `--self-check-uri` to `--url-checker-test-url` or `--url-checker-self-check-url`? I think that makes it clearer that the purpose of this URL is to check that the `UrlChecker` works.

meier-christoph:
@fthomas thanks for the feedback, I made the changes as you suggested.

fthomas (approved):
Thanks @meier-christoph!

PR #1934, opened 2021-02-15
Extract correct PR URL from Bitbucket API response

cb372:
Also add a test for the JSON decoding of PRs from Bitbucket API responses


fthomas (approved):
Thanks Chris, this LGTM!

I'm wondering if the current behaviour prevents some other functionality from working correctly or if this is "just" about showing the correct URL in the logs?

cb372:
I thought it was only used for logging, but the `html_url` is also used as a key in the `PullRequestRepository`.

I'm not really familiar enough with the code to understand the impact if that key changes. I guess we'll end up with a load of orphaned PRs in that repository, and potentially open duplicate PRs for a given dependency update?

fthomas:
> I guess we'll end up with a load of orphaned PRs in that repository

I think this will be the only consequence of this change but that is harmless. There shouldn't be duplicated PRs because Scala Steward always asks the repo manager if an update PR already exists before it creates a new one.

PR #1966, opened 2021-02-26
use slash syntax instead of "in"

xuwei-k:
prepare sbt 1.5 https://github.com/sbt/sbt/pull/6328


fthomas (approved):
Thanks @xuwei-k!

fthomas:
It would be cool if we could run the scalafix on updates to 1.5.0.

PR #1967, opened 2021-02-26
Migrate log4cats artifacts

fthomas:
I was just made aware that log4cats is now published under `org.typelevel`.  This will rewrite the organization of log4cats-core, log4cats-noop, log4cats-slf4j, and log4cats-testing. The package names for these were also changed from `io.chrisdavenport.log4cats` to `org.typelevel.log4cats`, so if we merge this all update PRs will fail because of that source incompatibility. In theory this could be salvaged with a Scalafix migration that rewrites the imports and is used at the same time, but I'm not sure if it is worth the hassle.

/cc @ChristopherDavenport, @rossabaker

### DIFF
--- a/modules/core/src/main/resources/artifact-migrations.conf
+++ b/modules/core/src/main/resources/artifact-migrations.conf
@@ -24,6 +24,30 @@ changes = [
     initialVersion = 1.0.0
   },
   {
+    groupIdBefore = io.chrisdavenport
+    groupIdAfter = org.typelevel
+    artifactIdAfter = log4cats-core
+    initialVersion = 1.2.0
+  },
+  {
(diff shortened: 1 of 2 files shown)


rossabaker:
I solved it with a search and replace when upgrading http4s.  I think it's 100% source compatible but for the package name.

They're less prominent, but unique, vault and keypool all made the same organization and package move.

fthomas:
I'm now trying to write a Scalafix. Maybe it can be done in a few lines of code. Let's see.

fthomas:
I added a Scalafix and it worked: https://github.com/scala-steward-org/scala-steward/pull/1969 :tada:

The PR changed the organization in the build and updated all imports.

PR #1994, opened 2021-03-05
Do not use sbt's logger to print stewardDependencies

fthomas:
If sbt's logger is configured to not print info messages, for example
via `logLevel := Level.Warn`, Scala Steward will not receive any
information about dependencies from sbt. Using `println` here should
fix that.


fthomas:
This reverts #1180 but unfortunately that ticket gives no reason for switching from `println` to sbt's logger and I also don't remember it anymore.

PR #2018, opened 2021-03-19
artifact-migrations.conf: slick-migration-api[-flyway]

nafg:


nafg:
It wasn't clear if initialVersion should be the last version before the change or the first version after the change. I did the former.

fthomas:
> the first version after the change

`initialVersion` should be that.

nafg:
Ok updated. You can squash merge if you like

PR #2019, opened 2021-03-20
Only create distinct changesets

### LINKED ISSUE #1995: Deduplicate identical changesets
As groups of dependencies managed by the same constant or variable end up looking independently updatable, we invariably end up with duplicate PRs with the same content.

Example:

- https://github.com/guardrail-dev/guardrail/pull/1012
- https://github.com/guardrail-dev/guardrail/pull/1013
- https://github.com/guardrail-dev/guardrail/pull/1014
- https://github.com/guardrail-dev/guardrail/pull/1015

I'm not sure how best to resolve this, a naive approach would be to order by GAV then deduplicate via a sliding window, we'd only get the first "group" opened and subsequent runs would deduplicate. This presumes common prefix, but that may be a good starting place.

blast-hardcheese:
Attempting to resolve #1995 

### Problem

Currently, when multiple artifacts share the same version number pinned by the same `val` binding, we end up seeing multiple PRs with the exact same changeset

### Examples

- guardrail-dev/guardrail#1012
- guardrail-dev/guardrail#1013
- guardrail-dev/guardrail#1014
- guardrail-dev/guardrail#1015

### Strategy in this PR

Attempting to deduplicate before we got to `applyUpdate` wasn't a viable strategy, due to the various different heuristics that could result in different commits.

Instead, track a `Ref[F, List[Branch]]` of every branch that we have walked in this run, then immediately after `applyUpdate` but before `pushCommits`, ensure that there are no branches that we have touched so far in this run that have the same changeset (as determined by the line count of `git diff other-branch`).

This will result in a little more IO churn, diffing between every other open and curated branch, but it should have a positive impact on users.

If you've got advice on how to write a more effective test for this new functionality, I'm open to it -- it seemed as though `NurtureAlgTest` would be the right place to put this, but there's not a lot of structure there to base my attempt on.

### Thank you
As always, a tremendous thank you for this fantastic tool.


blast-hardcheese:
Codacity failed because I used `.get` on a `Ref[F, A]`? Baffling

blast-hardcheese:
😅 Gonna think about this one more. If there is something obvious about the project structure I'm missing, I'm all ears, but I'll need to revisit in a few days.

blast-hardcheese:
OK, I think `codecov/patch` is incorrect.

I think the blast radius required to appropriately test this feature ended up being significantly more than initially anticipated, so I humbly submit this for review.

The tests needed to be written against real git running on the real filesystem, which is why I had to largely forego `MockContext`, but I admit I could have misunderstood something along the way and am happy to make any changes to help increase the overall quality of code in this project while also achieving the new feature I'm attempting to add.

Thanks!

fthomas:
It seems to me that this will only delay subsequent PRs with the same changes to the next run. If Scala Steward creates a PR for an update and that PR is not merged before the next run, Scala Steward will not pass that update to `NurtureAlg` since there is nothing to do for this update. But the update which has been prevented in the first run is still passed to `NurtureAlg` so that the previously prevented PR can now be created.

I'm also concerned that there is no principle which PR gets created and which PR is prevented. Currently the first update wins over the others and updates are AFAIK sorted by groupId and artifactId. What if a later update would have resulted in a "more correct" PR?

blast-hardcheese:
> What if a later update would have resulted in a "more correct" PR?

Only exactly equivalent changesets are deduplicated, using git itself as the final arbiter instead of trying to guess without modifying the filesystem. We accumulate commits, but only gate the push using this approach.

To your first point, I didn't realize updates were filtered before hitting NurtureAlg. Short of more back channels with distinct data flows, I don't know how to proceed here.

There may be an avenue to pursue, just tracking active branches in the store. This would work in the case you describe, being the initial set of branches to compare against, as well as working in the case of rebasing off master.

What do you think about this?

PR #2028, opened 2021-03-25
Use single branch updates

studouglas-scx:
Goal: have a single PR/branch that contains all scala steward updates. This way devs only need to approve a single PR per week rather than being overloaded. This is something scala-steward does not intend to implement.

This _seems_ to work when I was playing around with it against the CRM repos, but it's pretty sketchy so I wouldn't be surprised if there are issues.


studouglas-scx:
Disregard, wrong repo.

PR #2053, opened 2021-04-04
Implement MockEff as Kleisli with Ref[IO, MockState]

fthomas:
This implements `MockEff` as
```scala
type MockEff[A] = Kleisli[IO, Ref[IO, MockState], A]
```
instead of
```scala
type MockEff[A] = StateT[IO, MockState, A]
```

The motivation for this change is the upgrade to Cats Effect 3 where
`Context.step1` will most likely require a `Concurrent[F]` and
`Async[MockEff]` (and all type classes below) can be derived automatically
if `MockEff` is based on `Kleisli`. This constraint cannot be satisfied
with the `StateT` based `MockEff` for which only `Sync[MockEff]` can be
derived automatically.


fthomas:
Included in #2059.

PR #2098, opened 2021-04-27
pom.xml single update not working

anilkumarmyla:
Two unit tests that demonstrate #2099 (one is ignored that doesn't work and one that passes)


exoego:
@anilkumarmyla 
Thanks for a repro !!
I opened #2232 to address this.

PR #2126, opened 2021-05-18
Add a migration rule for scalafix 0.9.28

mlachkar:
Hello, we would like to add an automatic rule for the migration, but I don't know how to test to make sure it will work/resolve correctly. Thanks for your feedbacks

### DIFF
--- a/modules/core/src/main/resources/scalafix-migrations.conf
+++ b/modules/core/src/main/resources/scalafix-migrations.conf
@@ -143,6 +143,12 @@ migrations = [
     rewriteRules: ["github:typelevel/cats-effect/v3_0_0?sha=series/3.x"],
     scalacOptions: ["-P:semanticdb:synthetics:on"]
   },
+    {
+    groupId: "ch.epfl.scala",
+    artifactIds: ["scalafix-testkit"],
+    newVersion: "0.9.28",
+    rewriteRules: ["github:scalacenter/scalafix.g8/v0_9_28"]
+  },
   {
     groupId: "org.typelevel",
     artifactIds: ["cats-parse"],
--- a/modules/core/src/main/resources/scalafix-migrations.conf
+++ b/modules/core/src/main/resources/scalafix-migrations.conf
@@ -147,7 +147,7 @@ migrations = [
     groupId: "ch.epfl.scala",
     artifactIds: ["scalafix-testkit"],
     newVersion: "0.9.28",
-    rewriteRules: ["github:scalacenter/scalafix.g8/v0_9_28?sha=dba15e3"]
+    rewriteRules: ["github:scalacenter/scalafix.g8/v0_9_28"]
   },
   {
     groupId: "org.typelevel",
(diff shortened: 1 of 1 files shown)


modules/core/src/main/resources/scalafix-migrations.conf
```diff
@@ -142,6 +142,12 @@ migrations = [
     newVersion: "3.0.0",
     rewriteRules: ["github:typelevel/cats-effect/v3_0_0?sha=series/3.x"],
     scalacOptions: ["-P:semanticdb:synthetics:on"]
+  },
+    {
+    groupId: "ch.epfl.scala",
+    artifactIds: ["scalafix-testkit"],
+    newVersion: "0.9.28",
+    rewriteRules: ["github:scalacenter/scalafix.g8/v0_9_28"]
```
bjaglin, line 150: I don't think the current layout will work, we should rename back to `scalafix` to honor https://scalacenter.github.io/scalafix/docs/developers/sharing-rules.html#using-github

bjaglin: https://github.com/scalacenter/scalafix.g8/pull/21

bjaglin: To be safe, should this be pinned to a specific commit (since there are no tags over there) ?

mlachkar: ok

mlachkar: something like this : `  rewriteRules: ["github:scalacenter/scalafix.g8/v0_9_28?sha=dba15e3"]`

fthomas (approved):
This LGTM, thanks @mlachkar! Our current tests only check the syntax of the `scalafix-migrations.conf` file but not if Scalafix can resolve the given `rewriteRules`. Running `sbt scalafix github:scalacenter/scalafix.g8/v0_9_28` manually should suffice for that.

Do you have repositories in mind where this rule should be applied? It probably works for repositories that have scalafix-testkit as dependency in the sbt build in the root directory but not for repositories where Scalafix rules live under a separate sbt build in a `scalafix/` directory. It wouldn't work in these cases because Scala Steward runs the Scalafix rules in the sbt build in the root directory.

mlachkar:
I have just discovered that the rule needs to be under rules like this 
`scalafix/rules/src/main/scala/fix/RuleName.scala`, which is not the case for this rule (it's under scalafix/v0.9.28) 
I will create a gist and fix this PR as soon as https://github.com/scalacenter/scalafix.g8/pull/22 is merged

bjaglin:
> `scalafix/rules/src/main/scala/fix/RuleName.scala`, which is not the case for this rule (it's under scalafix/v0.9.28)

Oops, indeed...

> I will create a gist and fix this PR as soon as [scalacenter/scalafix.g8#22](https://github.com/scalacenter/scalafix.g8/pull/22) is merged.

You don't need a gist, you can use https://raw.githubusercontent.com/scalacenter/scalafix.g8/dba15e3684b80dea03005d846712fc4bb46ed292/scalafix/v0.9.28/rules/src/main/scala/fix/v0_9_28.scala as other migrations in that file.

mlachkar:
This pull request is ready! @fthomas can you please approve it again. Thanks

PR #2130, opened 2021-05-20
scalafix-testkit is mostly used through sbt-scalafix

bjaglin:
I am not quite sure how the resolution works, but since most of the repos will be bumping scalafix-testkit transitively through sbt-scalafix, I guess we need to make it explicit?

The releases are being pushed to maven central now, so I guess if I am right, this rather be merged fast.

### DIFF
--- a/modules/core/src/main/resources/scalafix-migrations.conf
+++ b/modules/core/src/main/resources/scalafix-migrations.conf
@@ -145,7 +145,7 @@ migrations = [
   },
     {
     groupId: "ch.epfl.scala",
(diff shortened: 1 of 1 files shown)


bjaglin:
cc @mlachkar

bjaglin:
apologies for the noise, I realize the bump on scalafix-testkit is actually direct: https://github.com/scalacenter/sbt-scalafix/blob/5b9bcfe312f525a359888c52000445681665a002/src/sbt-test/sbt-1.5/testkit/build.sbt#L40

PR #2135, opened 2021-05-22
Support reading and updating .mill-version file

lefou:
This PR adds support to read and update `.mill-version` files.

----
Initial but outdate comment:

_This is a draft, as the feature isn't complete yet_

@fthomas Thanks for the detailed steps in https://github.com/scala-steward-org/scala-steward/issues/1504#issuecomment-832950344. I managed to implement step 1 and 2, but fail at step 3. I think I'm not familiar enough with the used APIs. I could imagine it's just a finger exercise for you, in which case I'd be glad if you could help me out with a snippet. Please don't take it as a demand, but probably it speeds things up, at least on my side. ;-) Thank you!

I tried to add tests for step 1 and 2, but have no idea how to test step 3.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
@@ -74,7 +74,7 @@ final case class UpdatesConfig(
 
 object UpdatesConfig {
   val defaultFileExtensions: Set[String] =
-    Set(".scala", ".sbt", ".sbt.shared", ".sc", ".yml", "pom.xml")
+    Set(".mill-version", ".scala", ".sbt", ".sbt.shared", ".sc", ".yml", "pom.xml")
 
   implicit val updatesConfigEq: Eq[UpdatesConfig] =
     Eq.fromUniversalEquals
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
@@ -16,10 +16,11 @@
 
 package org.scalasteward.core.buildtool.mill
 
+import better.files.File
 import cats.effect.MonadCancelThrow
 import cats.syntax.all._
 import org.scalasteward.core.buildtool.BuildToolAlg
-import org.scalasteward.core.data.Scope
+import org.scalasteward.core.data.{ArtifactId, Dependency, GroupId, Resolver, Scope}
 import org.scalasteward.core.data.Scope.Dependencies
 import org.scalasteward.core.edit.scalafix.ScalafixMigration
 import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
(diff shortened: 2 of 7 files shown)


modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
```diff
@@ -74,7 +74,7 @@ final case class UpdatesConfig(
 
 object UpdatesConfig {
   val defaultFileExtensions: Set[String] =
-    Set(".scala", ".sbt", ".sbt.shared", ".sc", ".yml", "pom.xml")
+    Set(".mill-version", ".scala", ".sbt", ".sbt.shared", ".sc", ".yml", "pom.xml")
```
lefou, line 77: I was assuming alphabetical order, that's why I choose to first position for the insert. But now I see, `pom.xml` isn't in order.

fthomas:
For step 3 https://github.com/scala-steward-org/scala-steward/blob/9c921461694d68ce1d65a9e3ce36c6050b7e0837/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala#L209-L216
needs to be changed to something like this:
```scala
  val specific: UpdateHeuristic = UpdateHeuristic(
    name = "specific",
    replaceVersion = {
      case update: Update.Single if isScalafmtUpdate(update) =>
        defaultReplaceVersion(_ => List("version"))(update)
      case update: Update.Single if isMillUpdate(update) =>
        content => if (content === update.currentVersion) Some(update.nextVersion) else None
      case _ =>
        _ => None
    }
```

You could then test it with a test similar to this one: https://github.com/scala-steward-org/scala-steward/blob/9c921461694d68ce1d65a9e3ce36c6050b7e0837/modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala#L168-L179

lefou:
Thank you, @fthomas ! I finally found time to complete this feature, thanks to your detailed help!

lefou:
This PR is ready for review. /cc @exoego

fthomas (approved):
LGTM. Many thanks @lefou!

lefou:
Looks like it doesn't work as expected. E.g. the created PR https://github.com/lefou/mill-jbake/pull/2  looks releated to this PR. Although it properly detected the currently configured mill version (0.6.1) from `.mill-version`, and also calculated the best upgrade version (0.9.9), it hasn't edited the `.mill-version` file, but instead updated the `.travis.xml` file. Maybe, I did something wrong in my PR. Are there any logs I can access or which you could send me?

lefou:
My feeling is, that the change in UpdateHeuristic has not enough context to decide. E.g I'm not aware of the dependency between the file name `.mill-version` and the fact, that it should be associated with changes to a mill dependency.

fthomas:
@lefou The problem in https://github.com/lefou/mill-jbake/pull/2 is that an `UpdateHeuristic` other then the `specific` heuristic is updating the version number in the `.travis.yml` file. All later heuristics are ignored if an earlier heuristic updates the version. If you add a test like https://github.com/scala-steward-org/scala-steward/blob/7f67e25ecb197c5a8186f5f03ae9336153a1590c/modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala#L255-L267 which also contains the `.travis.yml` file with the line that was changed by Scala Steward, you can reproduce the issue with the above PR.

lefou:
@fthomas Thanks. I updated the test case to reproduce the false update and added it to the follow up PR.

PR #2167, opened 2021-06-16
Switched default docker base image from openjdk8 to adoptopenjdk11

### LINKED ISSUE #2156: Docker image for jdk 11
The official scala steward docker image fails for one of my projects because part of the sbt file relies on a java11 feature. The official docker image fails to steward my project as it runs on java 8.

Would it be possible to update the image to java 11 or provide a separate Dockerfile for java 11?
I could build one myself, but I think it might be handy for other people as well.

froth:
fixes #2156 

Unfortunately there exists no openjdk11-alpine image from the openjdk organisation (https://github.com/docker-library/openjdk/pull/235#issuecomment-424599754 https://github.com/docker-library/openjdk/issues/313).

Therefore I decided to switch to the alpine variant of adoptopenjdk. Another approach would be to switch to something like openjdk/11-jdk-slim. My reasoning behind doing it this way was to keep the operation system constant (and good experience with adoptopenjdk). If you prefer the other way around: I am happy to change the PR.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -241,7 +241,8 @@ lazy val metadataSettings = Def.settings(
 )
 
 lazy val dockerSettings = Def.settings(
-  dockerBaseImage := Option(System.getenv("DOCKER_BASE_IMAGE")).getOrElse("openjdk:8-jdk-alpine"),
+  dockerBaseImage := Option(System.getenv("DOCKER_BASE_IMAGE"))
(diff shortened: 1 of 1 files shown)


fthomas:
LGTM, but the `build.sbt` needs to be reformatted with `sbt fmt`.

froth:
> LGTM, but the `build.sbt` needs to be reformatted with `sbt fmt`.

done

fthomas (approved):

PR #2179, opened 2021-07-09
Add sbt-davenverse to the plugins that depend on sbt-github-actions

fthomas:
This will run `sbt githubWorkflowGenerate` if necessary in projects that use sbt-davenverse.

/cc @ChristopherDavenport @JesusMtnez


JesusMtnez (approved):
Thank you @fthomas !!

PR #2181, opened 2021-07-12
Rename Monocle organization

julien-truffaut:


fthomas:
@julien-truffaut Each artifact needs to be listed explicitly in the `changes` list. So for each artifact that has a new groupId, there needs to be an entry like this in the file:
```
  {
    groupIdBefore = com.github.julien-truffaut
    groupIdAfter = dev.optics
    artifactIdAfter = monocle-core
    initialVersion = 3.0.0
  },
```

julien-truffaut:
Thanks @fthomas

PR #2183, opened 2021-07-15
Allow choosing a custom default branch

alejandrohdezma:
# What has been done in this PR?

A new `--default-branch` argument has been added to the Scala Steward cli. When this argument is present, its value will be used as the default branch when updating repositories instead of the repository's default branch.

# Why this change?

This change enables users who run their own instances of Scala Steward to customize the branch in which updates are performed. This will enable multiple use-cases, for example:

- Updating multiple branches in a repository.
- Merging updates into a "develop" branch regularly that will be merged into the real default branch every so often.

This change will be specially useful for users of the [Scala Steward GitHub Action](https://github.com/scala-steward-org/scala-steward-action) after scala-steward-org/scala-steward-action#268 is merged.

Example:

```yml
on:
  schedule:
    - cron: '0 0 * * 0'
name: Launch Scala Steward on `main`, `1.x` and `2.x`
jobs:
  scala-steward:
    runs-on: ubuntu-latest
    name: Launch Scala Steward
    strategy:
      matrix:
        branch: ['main', '1.x', '2.x']
    steps:
      - name: Launch Scala Steward
        uses: scala-steward-org/scala-steward-action@v2
        with:
          github-token: ${{ secrets.REPO_GITHUB_TOKEN }}
          branch: ${{ matrix.branch }}
```

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -37,4 +37,5 @@ All command line arguments for the `scala-steward` application.
   --github-app-key-file  FILE                        Github application key file
   --github-app-id  ID                                Github application id
   --refresh-backoff-period DURATION                  Period of time a failed build won't be triggered again, default: "7 days"
+  --default-branch BRANCH                            A fixed branch name to use as default instead of the repository's default branch
 ```
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -59,7 +59,8 @@ object Cli {
       githubAppId: Option[Long] = None,
       urlCheckerTestUrl: Option[Uri] = None,
       defaultMavenRepo: Option[String] = None,
-      refreshBackoffPeriod: FiniteDuration = 7.days
+      refreshBackoffPeriod: FiniteDuration = 7.days,
+      defaultBranch: Option[String] = None
   )
 
   final case class EnvVar(name: String, value: String)
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
@@ -33,6 +33,7 @@ import org.scalasteward.core.vcs.VCSType
 import org.scalasteward.core.vcs.data.AuthenticatedUser
 import org.scalasteward.core.vcs.github.GitHubApp
 import scala.concurrent.duration.FiniteDuration
+import org.scalasteward.core.git.Branch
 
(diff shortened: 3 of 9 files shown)


ChristopherDavenport:
I wonder if there is a way we could adjust it to do multiple passes if branches were set via the .scala-steward.conf file. I know its not a simple hop from here to there, but then it would make sense that projects might set this and then steward could run and help when repositories are in cross-branching situations.

Almost all open source projects multi-branch at some point in their project lifecycles.

alejandrohdezma:
> I wonder if there is a way we could adjust it to do multiple passes if branches were set via the .scala-steward.conf file. I know its not a simple hop from here to there, but then it would make sense that projects might set this and then steward could run and help when repositories are in cross-branching situations.
> 
> Almost all open source projects multi-branch at some point in their project lifecycles.
> 
> @fthomas do you think that would be feasible?

I run through this idea over the weekend without success. I don't see a way of allowing this kind of configuration from the `.scala-steward.conf` that doesn't involve a complete refactor of how Scala Steward operates, given that Steward creates and searches caches based on the commit hash from the default branch before cloning the repository (and thus reading the configuration).

Given all that, I strongly believe that this solution is the best one that involves minimal changes, and allow branch selection for a wide range of Scala Steward users.

Of course, please @fthomas correct me if I'm wrong 😸

alejandrohdezma:
Hey @exoego, any chance we could have this reviewed? This is a feature several libraries could benefit from.

exoego:
@alejandrohdezma Thanks for this PR 👍 
I will review on this weekend.

exoego:
@alejandrohdezma 
Code LGTM 💯 

Could you also consider update the list of CLI arguments in https://github.com/scala-steward-org/scala-steward/blob/master/docs/help.md 🙏

alejandrohdezma:
> @alejandrohdezma Code LGTM 💯
> 
> Could you also update the list of CLI arguments in [https://github.com/scala-steward-org/scala-steward/blob/master/docs/help.md](https://github.com/scala-steward-org/scala-steward/blob/master/docs/help.md?rgh-link-date=2021-10-02T10%3A53%3A49Z) 🙏

Done [here](https://github.com/scala-steward-org/scala-steward/pull/2183/commits/1451b1d503d955262ee48ff4010f3154a780e4d3) :)

exoego (approved):
LGTM

exoego:
I am going to merge this, since this is great addition and changes LGTM.
If any issues found later, I am fine to fix/revert.

PR #2194, opened 2021-08-09
Teads is now using Scala Steward internally

github-brice-jaglin:
And it's a real pleasure!


fthomas:
Thanks @github-brice-jaglin!

PR #2201, opened 2021-08-18
Remove removed scala update config

abuggin:
Noticed that in the tests there is a legacy option. Maybe worth removing?


fthomas (approved):
Good catch. Thanks @abuggin!

PR #2216, opened 2021-09-03
sbt-avro artifact migration

RustedBones:


RustedBones:
Hi @fthomas . I think I made an error on that:
- I release `com.github.sbt:sbt-avro:3.2.0`
- There existed a previous release  `com.cavorite:sbt-avro:3.2.0`

It seems the artifact migration is not picked up in that case.
Should I create a new release and update the initial version here so that `scala-steward` will be able to apply the migration rule ?

fthomas:
@RustedBones A new release and bumping the `initialVersion` here would be indeed the easiest way to trigger the migration. I guess the documentation should mention that the `initialVersion` needs to be greater than the latest version with the old groupId.

PR #2227, opened 2021-09-15
Update refined and case-app

exoego:
Supersedes #2187

This updates `refined` and `case-app` together, since `case-app` v2.1.0-M1 dependes on refined 0.9.26+.

Test code are slightly fixed, since `case-app` now outputs ANSI colors after https://github.com/alexarchambault/case-app/releases/tag/v2.1.0-M1
> New help / usage formats (picocli-inspired, with ANSI colors, etc.)


fthomas (approved):
Nice! This should also unblock #2144 and #2145.

PR #2230, opened 2021-09-15
Migrate to newer artifact version

RustedBones:
Trigger migration with new version of `sbt-avro` in `com.github.sbt` organisation
See #2216


exoego (approved):
LGTM 👍

PR #2237, opened 2021-09-20
Add artifact migration for better-tostring

kubukoz:
See https://github.com/polyvariant/better-tostring/pull/75.

I will un-draft when the version is synced to Central :)

### DIFF
--- a/modules/core/src/main/resources/artifact-migrations.conf
+++ b/modules/core/src/main/resources/artifact-migrations.conf
@@ -212,6 +212,6 @@ changes = [
   {
     groupIdAfter = org.polyvariant
     artifactIdAfter = better-tostring
-    initialVersion = 0.3.7
+    initialVersion = 0.3.8
   }
 ]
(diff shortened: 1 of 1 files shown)


modules/core/src/main/resources/artifact-migrations.conf
```diff
@@ -208,5 +208,10 @@ changes = [
     artifactIdBefore = argonaut-refined_6.2
     artifactIdAfter = argonaut-refined_6.3
     initialVersion = 1.3.0
+  },
+  {
+    groupIdAfter = org.polyvariant
```
fthomas, line 213: @kubukoz The `groupIdBefore` is missing. Too bad this does not result in parse / test error.

kubukoz: Thanks, I'll fix it later (need to figure out why publishing doesn't work first)

kubukoz:
Please review @fthomas :)

exoego (approved):
LGTM!

PR #2241, opened 2021-09-21
Revised .mill-version update support

lefou:
Follow up to PR https://github.com/scala-steward-org/scala-steward/pull/2135, hopefully correcting the update behavior. I extended the test case to reproduce and fix the seen issue.

### DIFF
--- a/modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala
@@ -259,21 +259,11 @@ class EditAlgTest extends FunSuite {
     )
     val original = Map(
       ".mill-version" -> """0.9.5""",
-      ".travis.yml" ->
-        """
-          |env
-          |  - TEST_MILL_VERSION=0.9.4
-          |  - TEST_MILL_VERSION=0.9.5
-          |""".stripMargin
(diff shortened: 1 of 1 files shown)


lefou:
Ready for review. Thanks for your helpful feedback @fthomas and @exoego.

modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala
```diff
@@ -258,10 +258,22 @@ class EditAlgTest extends FunSuite {
       newerVersions = Nel.of("0.9.9")
     )
     val original = Map(
-      ".mill-version" -> """0.9.5"""
+      ".mill-version" -> """0.9.5""",
+      ".travis.yml" ->
+        """
+          |env
+          |  - TEST_MILL_VERSION=0.9.4
+          |  - TEST_MILL_VERSION=0.9.5
```
fthomas, line 266: This is the only relevant line for this test, right? We're checking that the version number is only updated in `.mill-version` and not here. I would therefore delete all other lines to keep it lean:
`".travis.yml" -> "- TEST_MILL_VERSION=0.9.5"`

fthomas (approved):

exoego (approved):
LGTM

PR #2242, opened 2021-09-22
Introduce EditAttempt enum

fthomas:
This introduces the `EditAttempt` enum for the different types of edits Scala Steward can do in a PR: version bumps, Scalafix changes, and post-update hooks changes. The purpose of this change is that we can in the future use the list of `EditAttempts` returned by `EditAlg` to create more faithful PR bodies that better reflect what Scala Steward did: for example only mention Scalafix migrations if they result in changes, indicate if a Scalafix migration failed, or add labels for post-update hooks., etc.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
@@ -25,8 +25,8 @@ import org.scalasteward.core.application.Config
 import org.scalasteward.core.coursier.CoursierAlg
 import org.scalasteward.core.data.ProcessResult.{Created, Ignored, Updated}
 import org.scalasteward.core.data._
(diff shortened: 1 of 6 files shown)


modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -42,7 +41,6 @@ final class NurtureAlg[F[_]](config: Config)(implicit
     gitAlg: GitAlg[F],
     logger: Logger[F],
     pullRequestRepository: PullRequestRepository[F],
-    scalafixMigrationsFinder: ScalafixMigrationsFinder,
```
fthomas, line 45: That we can get rid of `ScalafixMigrationsFinder` in `NurtureAlg` is what I like most about this change.

Issue #1963, opened 2021-02-25
Handle artifact migrations similar to Scalafix migrations


fthomas:
See https://github.com/tototoshi/scala-ojisan/pull/1#commitcomment-47400009.

PR #1650, opened 2020-10-09
Load Scalafix migrations from this repository

fthomas:
This changes `MigrationAlg` so that it loads always the latest Scalafix
migrations from this repository instead of using the static list of
migrations that is shipped as resource in the JAR.

The following changes with this PR:
* The `--scalafix-migrations` option can now be used multiple times to
  load extra migrations not only from one but multiple files.
* Arguments to the `--scalafix-migrations` cannot only be local files
  but also remote URLs.
* Disabling loading the default Scalafix migrations is now done via a
  new `--disable-default-scalafix-migrations` option instead of using
  `disableDefaults` in an extra migrations file.
* Scala Steward now prints how many Scalafix migrations it loaded at
  start-up.

Closes: #1603

PR #2249, opened 2021-09-23
Load artifact migrations from this repository

fthomas:
This splits `ArtifactMigrations` into a loader and finder class. The
former loads the latest artifact migrations from this repository instead
of using the static list of migrations that is shipped as resource in
the JAR. That means that all Scala Steward instances use the artifact
migrations from this repository as soon as they are added here.

The following changes with this PR:
* The `--artifact-migrations` option can now be used multiple times to
  load extra migrations not only from one but multiple files.
* Arguments to the `--artifact-migrations` cannot only be local files
  but also remote URLs.
* Scala Steward prints how many artifact migrations it loaded at
  start-up.

The same has been done for Scalafix migrations in #1650.

Closes: #1963


modules/core/src/main/scala/org/scalasteward/core/update/artifact/ArtifactChange.scala
```diff
@@ -0,0 +1,37 @@
...
+import io.circe.generic.extras.{semiauto, Configuration}
+import org.scalasteward.core.data.GroupId
+
+final case class ArtifactChange(
+    groupIdBefore: Option[GroupId],
+    groupIdAfter: GroupId,
+    artifactIdBefore: Option[String],
+    artifactIdAfter: String,
+    initialVersion: String
+)
```
exoego, line 29: How about adding simple assertion here to address https://github.com/scala-steward-org/scala-steward/issues/2238 ?

```suggestion
) {
  require(groupIdBefore.isDefined || artifactIdBefore.isDefined,
    "Either or both of groupIdBefore and artifactIdBefore must be set")
}
```

fthomas: Let's do this in another PR. I'd like to keep this PR focused on the loading of migrations.

exoego (approved):
LGTM

Issue #2276, opened 2021-10-12
Log errors in CoursierAlg.getVersions


fthomas:
See https://github.com/scala-steward-org/scala-steward/issues/2202#issuecomment-940089050.

fthomas:
Sure!

Issue #2202, opened 2021-08-18
Find newer dependency versions for dependencies that reside in Gitlab Package Registry


L7R7:
We have a project that uses scala-steward for dependency upgrades. We recently switched the hosting of our own internal libraries to [Gitlab Package Registry](https://docs.gitlab.com/ee/user/packages/maven_repository/index.html) instead of a nexus. Scala Steward doesn't seem to check those dependencies (or it doesn't find the newer packages for some reason). I can't see anything in the logs (even on DEBUG level) on to why that is, so I'm not really sure how to proceed here.

I have the following ideas:

* We're using a [sbt plugin](https://github.com/gilcloud/sbt-gitlab) that handles all the tricky stuff regarding talking to Gitlab. Since the Gitlab API requires authentication, we probably have to provide the authentication information to the scala steward.I would either have to provide a file with the relevant information or set env variables that are there when sbt runs. However, I would have expected error logs ("can't authenticate ..." or something like that) if that was the problem
* The plugin requires us to [disable Coursier](https://github.com/gilcloud/sbt-gitlab#dependency-resolution). Maybe that's the problem?

Does Scala Steward support that use case? If so, what am I missing there?

PR #2277, opened 2021-10-12
Log errors in CoursierAlg#getVersions

L7R7:
This refers to #2276. As mentioned in #2202, it's only logged on TRACE level


L7R7:
I managed to get the scalafmt complaints out of the way, but I don't know how to deal with the failing codecov job yet

PR #2301, opened 2021-10-19
Add authors field to ScalafixMigration

### LINKED ISSUE #1914: Attribute Scalafix commits via "Co-authored-by" to the rule author?
Scalafix commits are currently attributed in the Git history to the Scala Steward user although it only did the easy part of running the rule. The hard part of writing the rule is done by someone else who's currently not attributed in any way. Wouldn't it be nice if we append [Co-authored-by](https://stackoverflow.com/a/7442255) attributions to the Scalafix commit if the rule author(s) opts into it. We could support this by adding a `authors: Option[Nel[GitAuthor]]` field to [`Migration`](https://github.com/scala-steward-org/scala-steward/blob/b5cacf994919cd95c08369518a99a389deb3c6c4/modules/core/src/main/scala/org/scalasteward/core/scalafix/Migration.scala).

fthomas:
Closes #1914.


Philippus:
It was a bit surprising seeing you as a co-author in a PR in one of our private repositories. But I see now where it came from.

Issue #1889, opened 2021-01-18
Run scalafmt on every update?


anilkumarmyla:
We over-ambitiously keep our build.sbt files formatted through scalafmt. Often times it happens that a library version string gets longer/shorter and PR builds fail on scalafmt since the alignment of sbt configuration (ex: `% Test`) is broken, see example below
 
```
  "net.javacrumbs.json-unit"    % "json-unit"               % "2.22.0"   % Test,
- "org.mockito"                 % "mockito-core"            % "3.6.28"   % Test,
+ "org.mockito"                 % "mockito-core"            % "3.7.7"   % Test,
  "org.scalatest"              %% "scalatest"               % "3.2.3"    % Test,
```

Are there any tips/tricks to avoid or workaround this issue? A quick fix is to run scalafmt to fix the alignment but currently scalafmt runs only if there is a scalafmt update. Thoughts?

fthomas:
> Are there any tips/tricks to avoid or workaround this issue?

Nothing I'm aware of.

> A quick fix is to run scalafmt to fix the alignment but currently scalafmt runs only if there is a scalafmt update. Thoughts?

I think it is reasonable that Scala Steward runs scalafmt after bumping the version and before creating the update commit. It should be configurable via the repo config and be disabled by default.

anilkumarmyla:
Thanks @fthomas for this enhancement and making scala-steward incrementally better with each iteration

PR #2319, opened 2021-10-29
Run scalafmt before creating the update commit

fthomas:
This runs `scalafmt --mode changed` after bumping the version but before
creating the update commit. With `--mode changed` only changed files are
reformatted. This reformatting is different to the `PostUpdateHook` that
runs `scalafmt` on scalafmt updates because formatting changes are not
recorded in a separate commit. This is not (yet) configurable via
`.scala-steward.conf` but that could be added later if needed.

We're also only running scalafmt if the project depends on it.

Example PR: https://github.com/scala-steward-org/test-repo-2/pull/59

Closes: #1889

### DIFF
--- a/modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala
@@ -5,13 +5,14 @@ import cats.effect.unsafe.implicits.global
 import munit.FunSuite
 import org.scalasteward.core.TestInstances.dummyRepoCache
 import org.scalasteward.core.TestSyntax._
-import org.scalasteward.core.data.{GroupId, RepoData, Update}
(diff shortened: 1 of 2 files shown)


fthomas:
Btw, here is a PR where a previously unformatted file gets reformatted: https://github.com/toddburnside/scalajs-react-virtuoso/pull/70

I also thought a little bit about a `.scala-steward.conf` flag for disabling this feature. Maybe that is not needed since one can already ignore files via the `.scalafmt.conf` file. So instead of disabling this feature in `.scala-steward.conf` just ignore the files you don't want reformatted in `.scalafmt.conf`.

fthomas:
And here is a PR where scalafmt fixed the formatting that Scala Steward messed up before: https://github.com/gemini-hlsw/lucuma-broadcast-channel/pull/56. These are the situations for which this feature has been implemented.

Issue #1518, opened 2020-07-02
`javac` missing in the docker image


joan38:
```
[error] ## Exception when compiling 72 sources to /opt/scala-steward/workspace/repos/SDE/playback-summarizer/target/scala-2.12/classes
[error] java.io.IOException: Cannot run program "javac" (in directory "/opt/scala-steward/workspace/repos/SDE/playback-summarizer"): error=2, No such file or directory
```
For repos containing some Java files.

To me this is because we should be using `openjdk:8-jdk-alpine` instead of `openjdk:8-jre-alpine` in https://github.com/fthomas/scala-steward/blob/master/build.sbt#L196

If people agree on this, I can make the PR.
Let me know.

fthomas:
I've no objection to changing the base image.

A related thought I had recently was if it would be easier to use [`cs setup`](https://www.scala-lang.org/2020/06/29/one-click-install.html) to install a JDK, sbt, maven, mill than what [we're currently doing](https://github.com/fthomas/scala-steward/blob/fcf491d2efe60a0fb4461da4ee461be9c6083d83/build.sbt#L198-L207).

joan38:
`curl: not found` curl is not even on openjdk alpine

fthomas:
I guess we could install `curl` with `apk` [here](https://github.com/fthomas/scala-steward/blob/7732e4c47125fe9ff4e2157d43cf6709532dbe94/build.sbt#L205).

joan38:
I've not been able to make the cs exec work on alpine. Ubuntu is ok tho.

fthomas:
We moved from a Debian based image to Alpine in #1370 because of the image size. It this point I'd accept a PR changing the base image to `openjdk:8-jdk-alpine` and defer using Coursier to another time.

PR #2322, opened 2021-10-31
Wait until currently used dependencies are sufficiently old before updating them

### LINKED ISSUE #1581: Time-based delay configuration for new dependencies
I found myself looking for a feature where you could say "Only update a dependency if a configured amount of time has passed" but I couldn't find an already existing solution to this. I imagine a feature like this could be set per groupId/dependency, but also globally.

The biggest motivation for me being that receiving daily PRs for dependencies that get released nightly (like any of the AWS SDK modules) can be a bit much, especially for expensive, long-running CI builds. If I could specify a "delay period" then I could effectively change these daily updates to be weekly and so on.

Also, sometimes brand new releases get unintentional breaking changes shipped out (which are usually later corrected within a couple of days). This feature could be useful for filtering out things like that. Like "I only want to trust a new dependency if X amount of time has passed without a new release".

rtyley:
This aims to fix #1581 - _"receiving daily PRs for dependencies that get released nightly (like any of the [AWS SDK modules](https://search.maven.org/artifact/software.amazon.awssdk/s3)) can be a bit much, especially for expensive, long-running CI builds."_ - ie not everybody wants to merge a new PR daily for one of these...! :

image

This PR will allow users to configure Scala Steward so that it _won't_ open PRs to update dependencies that are younger than the specified threshold - eg, if the artifact I'm _using_ is only a few days old, I don't want to get a PR to update it just yet!

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
@@ -32,6 +32,7 @@ import org.scalasteward.core.edit.hooks.HookExecutor
 import org.scalasteward.core.edit.scalafix.{ScalafixMigrationsFinder, ScalafixMigrationsLoader}
 import org.scalasteward.core.git.{GenGitAlg, GitAlg}
 import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
+import org.scalasteward.core.mavencentral.MavenCentralApiAlg
 import org.scalasteward.core.nurture.{NurtureAlg, PullRequestRepository}
 import org.scalasteward.core.persistence.{CachingKeyValueStore, JsonKeyValueStore}
 import org.scalasteward.core.repocache._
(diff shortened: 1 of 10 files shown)


fthomas:
I've not looked in detail at the changes but noticed that it only works for artifacts on Maven Central. There is another way to get the age (or an estimate of it) that works with all package repositories. Scala Steward has a [persistent cache of all artifact versions](https://github.com/scala-steward-org/scala-steward/blob/8599b276cfc1862b1feff2e74d501af4a7252160/modules/core/src/main/scala/org/scalasteward/core/coursier/VersionsCache.scala). When we get the list of available versions from Coursier we could attach the current time to versions that are not yet present in the cache. Versions that have been previously cached are already associated with an older timestamp. So over time each version is associated with a timestamp when Scala Steward has seen it for the first time. This is of course not the real age of an artifact but a lower limit that still can be used to implement this feature. The accuracy of that age would be proportional to how frequently Scala Steward runs and updates its version cache.

We could also have a look at Coursier's [`Versions#lastUpdated`](https://github.com/coursier/coursier/blob/525e0a52014ab1229199b820f374b4b4669b9084/modules/core/shared/src/main/scala/coursier/core/Versions.scala#L9). I don't know what it is exactly but maybe

fthomas:
@reibitto said in #1581:
> I think the 3rd thing in my issue that I seem to unintentionally be asking for is making it possible to set `pullRequests.frequency` at the groupId/artifactId level too. So that you can say "Update these AWS dependencies less frequently than the others".

@rtyley Would this work for you too? If we had `pullRequests.frequency` at the groupId/artifactId level, we wouldn't need to know the age of dependencies to implement it. We could just check that sufficient time has passed since the last PR of a dependency before proposing an update.

exoego:
Let me close this PR since the feature #1518 is realized in PR #2515.
rtyley, thanks for your effort !!

PR #2328, opened 2021-11-02
Allow running Scalafix migrations after version bump

fthomas:
This allows to run Scalafix migrations after the version bump. To run a
migration after the version bump, the `executionOrder` field has to be
set to `"post-update"`. The effective default value for this field is
`"pre-update"`.

Post-update migrations are useful if they consist of multiple rules
that only produce compiling code if the version is bumped before. One
example is the FS2 migration added here that consists of multiple
`replace:` rules. The symbols it renames are still present after the
bump to > 3.0.7. So it is OK to apply them successively after the
version bump but it would lead to compile errors if they were applied
before the bump because each rule could produce code that does not
compile with the old version. Running subsequent rules would then fail
with compile errors.

Here is an example PR where two rules where applied and that would have
failed without being `post-update`: https://github.com/scala-steward-org/test-repo-2/pull/71

### DIFF
--- a/modules/core/src/main/resources/scalafix-migrations.conf
+++ b/modules/core/src/main/resources/scalafix-migrations.conf
@@ -9,6 +9,8 @@ migrations = [
     groupId: "co.fs2",
     artifactIds: ["fs2-.*"],
     newVersion: "3.0.7",
+    // Note that the replace rule can generate broken code in some cases.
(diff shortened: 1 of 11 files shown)


fthomas:
Paging @mpilquist for the added FS2 migration.

Issue #1104, opened 2019-11-24
Wait for release announcement before creating PRs


fthomas:
This idea originated in: https://twitter.com/NicolasRinaudo/status/1198162957164777473

Scala Steward currently detects new releases by searching for newer JARs. So once a maintainer publishes newer JARs, Scala Steward starts creating PRs. For some projects publishing JARs does not mean that a new release is complete. Maintainers might still want to write release notes, create other packages, update their websites, etc. before making a release public. For these projects it would be nice if Scala Steward would wait with creating PRs until the new version is announced.

Unfortunately announcing a new release is not a standardized process and everyone does it differently - if at all. So detecting that a new version has been announced seems to the hardest part to support this feature.

Some ideas from the Twitter thread:
* Include metadata file in repos / JARs / POMs that tell Scala Steward how to detect if a new version is announced. (https://twitter.com/olafurpg/status/1198543160483491846, https://twitter.com/d6/status/1198623194036494336)
* Check that a release exists in the GitHub repo. (https://twitter.com/dwijnand/status/1198298739791806470, https://twitter.com/dwijnand/status/1198653955636703233)

In my ideal world, this would not only work with GitHub and could be used by other tools than Scala Steward.

dwijnand:
Perhaps there should be a grace period (24 hours? 72 hours?) after which ScalaSteward sends the PRs anyway?

SethTisue:
This came up again with the release of Scala 3.1. Library authors are merging the 3.1 bump, even though perhaps they shouldn't, because:

* https://twitter.com/SethTisue/status/1450260785762947076
* https://twitter.com/Kordyjan/status/1450406709579718660

and there isn't really official guidance on this yet, wording is still being discussed on https://github.com/scala/scala-lang/pull/1284

In the case of Scala, the release announcement _on GitHub_ does exist: https://github.com/lampepfl/dotty/releases/tag/3.1.0 , but it's just bullets, it's not the true, full release announcement, which is the blog post.

So... I don't know, this is hard. Maybe if there are only JARs on Maven Central but no GitHub release, wait 72 hours, and if there is a GitHub release, wait 24 hours?

[...]

PR #2337, opened 2021-11-09
Support reading a default repo config from this repository

fthomas:
This adds support for reading a defaultl repo config from this repository. Multiple repo configs can be loaded per Scala Steward installation with the `--repo-config` command-line option and are merged into one effective global repo config. Loading the default repo config from this repository can be disabled with `--disable-default-repo-config`. This is similar to the handling of Scalafix and artifact migrations config files.

The purpose of this change is to allow library maintainers to tell Scala Steward to ignore updates until a new version is announced (#1104). They can do this adding new entries to the `updates.ignore` list of the default repo config. All (public and private) Scala Steward installations that do not opt-out of loading the default repo config will then ignore the added updates until the default repo config is changed again.
 
The default repo config also contains ignored updates that were previously removed by `FilterAlg.removeBadVersions`. So this function was just a default `updates.ignore` list in disguise.

Note also that this is a breaking change for users that are using the old `--default-repo-conf` option. `--repo-config` is a drop-in replacement for that option.

### DIFF
--- a/modules/core/src/main/resources/default.scala-steward.conf
+++ b/modules/core/src/main/resources/default.scala-steward.conf
@@ -5,6 +5,12 @@ updates.ignore = [
   { groupId = "org.scala-lang", artifactId = "scala-reflect",  version = "2.13.8" },
   { groupId = "org.scala-lang", artifactId = "scalap",         version = "2.13.8" },
 
+  // Ignore the next Scala 2.12 version until it is announced.
+  { groupId = "org.scala-lang", artifactId = "scala-compiler", version = "2.12.16" },
+  { groupId = "org.scala-lang", artifactId = "scala-library",  version = "2.12.16" },
+  { groupId = "org.scala-lang", artifactId = "scala-reflect",  version = "2.12.16" },
+  { groupId = "org.scala-lang", artifactId = "scalap",         version = "2.12.16" },
+
   { groupId = "com.google.guava", artifactId = "guava", version = "r03" },
   { groupId = "com.google.guava", artifactId = "guava", version = "r05" },
   { groupId = "com.google.guava", artifactId = "guava", version = "r06" },
--- a/modules/core/src/main/resources/default.scala-steward.conf
+++ b/modules/core/src/main/resources/default.scala-steward.conf
@@ -18,11 +18,31 @@ updates.ignore = [
   // https://github.com/scala-steward-org/scala-steward/issues/1753
   { groupId = "commons-codec", artifactId = "commons-codec", version = "20041127.091804" },
 
+  { groupId = "commons-collections", artifactId = "commons-collections", version = "20030418.083655" },
+  // https://github.com/albuch/sbt-dependency-check/pull/107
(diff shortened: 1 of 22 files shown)


modules/core/src/main/resources/default.scala-steward.conf
```diff
@@ -0,0 +1,59 @@
+updates.ignore = [
+  // Ignore the next Scala 3 version until it is announced.
+  { groupId = "org.scala-lang", artifactId = "scala3-compiler",      version = { exact = "3.1.1" } },
+  { groupId = "org.scala-lang", artifactId = "scala3-library",       version = { exact = "3.1.1" } },
+  { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1",  version = { exact = "3.1.1" } },
```
fthomas, line 5: @Kordyjan @romanowski FYI: Here are the version numbers that need to be bumped once 3.1.1 is announced.

Btw, is 3.1.1 the next Scala 3 version that will be released? I'm asking because a 3.1.2-RC version seems to be already available.

modules/core/src/main/resources/default.scala-steward.conf
```diff
@@ -0,0 +1,59 @@
...
+  { groupId = "org.scala-lang", artifactId = "scala-compiler", version = { exact = "2.13.8" } },
+  { groupId = "org.scala-lang", artifactId = "scala-library",  version = { exact = "2.13.8" } },
+  { groupId = "org.scala-lang", artifactId = "scala-reflect",  version = { exact = "2.13.8" } },
+  { groupId = "org.scala-lang", artifactId = "scalap",         version = { exact = "2.13.8" } },
+
+  // Ignore the next Scala 2.12 version until it is announced.
+  { groupId = "org.scala-lang", artifactId = "scala-compiler", version = { exact = "2.12.16" } },
+  { groupId = "org.scala-lang", artifactId = "scala-library",  version = { exact = "2.12.16" } },
+  { groupId = "org.scala-lang", artifactId = "scala-reflect",  version = { exact = "2.12.16" } },
+  { groupId = "org.scala-lang", artifactId = "scalap",         version = { exact = "2.12.16" } },
```
fthomas, line 17: @SethTisue FYI: Here are the version numbers that need to bumped once 2.12.16 / 2.13.8 is announced.

modules/core/src/main/resources/default.scala-steward.conf
Kordyjan: I think you can see nightly builds of the compiler as 3.1.2-RC, as 3.1.2-RC1 will be the next release from the master branch. 3.1.1 is right now after the first RC and lives on `release-3.1.1` branch.

eugeniyk:
Is there a way to disable fetching default configuration from github? With on-prem & firewall rules new version can't be run

fthomas:
@eugeniyk Yes, with `--disable-default-repo-config` Scala Steward should not fetch the default config from GitHub.

eugeniyk:
ah ok, thanks!

PR #2339, opened 2021-11-11
Ignore the next Scala.js versions until they are announced.

sjrd:
This follows the idea used for Scala itself. The difference is that we don't know in advance whether the next release will be a patch or minor version, so we exclude both possibilities.

We exclude all the artifacts that are published by the core Scala.js repository.


modules/core/src/main/resources/default.scala-steward.conf
```diff
@@ -16,6 +16,38 @@ updates.ignore = [
   { groupId = "org.scala-lang", artifactId = "scala-reflect",  version = { exact = "2.12.16" } },
   { groupId = "org.scala-lang", artifactId = "scalap",         version = { exact = "2.12.16" } },
 
+  // Ignore the next Scala.js patch version until it is announced.
+  { groupId = "org.scala-js", artifactId = "sbt-scalajs",                   version = { exact = "1.7.2" } },
```
sjrd, line 20: Will this work? Or should it be `sbt-scalajs_1.0` or even `sbt-scalajs_1.0_2.12`?

fthomas: Yes, this will work. The strings here are the same as the ones used in PRs like https://github.com/typelevel/cats/pull/4011.

The above PR also lists `scalajs-library_2.13` and `scalajs-test-bridge_2.13` in addition to their names without the Scala binary suffix. I'm not sure why that is the case. I'll take a look - it could be a bug in Scala Steward or a quirck in the Cats build. To be on the safe side, these two should be added here too.

We could also choose to not list every artifact individually and just ignore all artifacts with the `org.scala-js` groupId and the versions 1.7.2 and 1.8.0:
```
{ groupId = "org.scala-js", version = { exact = "1.7.2" } },
{ groupId = "org.scala-js", version = { exact = "1.8.0" } },
```
This would be fine if there are no other `org.scala-js` artifacts that use these versions.

sjrd:
Thank you very much!

fthomas:
You are welcome!

modules/core/src/main/resources/default.scala-steward.conf
fthomas: > The above PR also lists `scalajs-library_2.13` and `scalajs-test-bridge_2.13` in addition to their names without the Scala binary suffix. I'm not sure why that is the case. I'll take a look - it could be a bug in Scala Steward or a quirck in the Cats build.

For posterity, here is the reason why both artifact names are listed in the PR:
```
sbt:cats> ++3.0.2
sbt:cats> testkitJS/libraryDependencies
...
[info] * org.scala-js:scalajs-library_2.13:1.7.0
...
sbt:cats> ++2.13.7
sbt:cats> testkitJS/libraryDependencies
...
[info] * org.scala-js:scalajs-library:1.7.0
...
```

PR #2359, opened 2021-11-16
Ignore bloop-config 121807cc

fthomas:


mzuehlke:
But why are we updating from a "normal" version to a hash in the first place ?

fthomas:
I think we currently require that hashes start with a `-` or `+`.

fthomas:
I agree that we should not propose such updates in the first place. I'm therefore closing this PR.

PR #2360, opened 2021-11-16
Adjust docs for default repo config cli options

### LINKED ISSUE #2344: Unrecognized argument: --default-repo-conf
I noticed recently that my own gitlab CI based run of scala-steward started to fail. 
It seems that the command line option `--default-repo-conf` has been removed recently (but still mentionned [here](https://github.com/scala-steward-org/scala-steward/blob/master/docs/running.md)  )

At the same time, a new `--repo-config` argument has been introduced, as well as a `--disable-default-repo-config ` boolean argument. 

Could you please clarify how is it supposed to work and how to point scala-steward to a default configuration.

mzuehlke:
Adjust docs for changes to the cli options when introducing default repo config

see #2337
closes #2344

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -8,7 +8,7 @@ sbt stage
 ./modules/core/.jvm/target/universal/stage/bin/scala-steward \
   --workspace  "$STEWARD_DIR/workspace" \
   --repos-file "$STEWARD_DIR/repos.md" \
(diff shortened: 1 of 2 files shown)


fthomas (approved):
Thanks Marco! I think we can also close #2344 when merging this.

PR #2377, opened 2021-11-26
Rework syntax for creating updates in tests

fthomas:
This reworks the syntax for creating updates in tests. It
* does not use the `Update.Single.apply` and `Update.Group.apply` methods anymore. This will make it easier to change both classes since we  need to update the `apply` method in a lot less places.
* gets rid of implicit conversions in `TestSyntax`. The syntax is now based on extension methods via implicit classes only.
* makes creating `Update`s more concise.


mzuehlke (approved):
That's easier to read 👍

PR #2415, opened 2021-12-20
Add sbt-mergify-github-actions to list of plugins that depend on sbt-github-actions

nafg:


nafg:
(It's a plugin to update .mergify.yml whenever `githubWorkflowGenerate` is run.)

PR #2434, opened 2021-12-26
Repo specific post update hooks

### LINKED ISSUE #1997: Feature request: run arbitrary (sbt) action after the update
I would like to run a custom sbt task after each update, before creating the update commit, that can make additional changes in the repository.

This is a very specific use case for zio-aws where scala-steward updates the full AWS SDK BOM that sometimes get new aws services included. In that case zio-aws's github actions definition must be regenerated to include the new aws libraries. I have an sbt task to do that, but it would be really nice to be able to trigger it automatically for each update.

If implemented as a general purpose extension point it would be probably useful for others too. For example it could also solve #1889

### LINKED ISSUE #2097: Adding Custom Plugin Behavior
I'm considering adding pom.xml files that can be automatically generated by SBT to my applications so github can do dependency analysis for me. Since scala-steward updates dependencies that would modify these generated files as well, and would fail if I put in place CI to make sure they were in alignment with generated code, I was wondering how I would/could add support similar to scalafmt to run a custom sbt command before commiting code. If that is even possible.

Many thanks for all your time, I love scala-steward. It's awesome and so are you!

987Nabil:
Fixes #1997 
fixes #2097

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
@@ -51,7 +51,10 @@ final class HookExecutor[F[_]](implicit
 
   private def execPostUpdateHook(repo: Repo, update: Update, hook: PostUpdateHook): F[EditAttempt] =
     for {
-      _ <- logger.info(s"Executing post-update hook for ${update.groupId}:${update.mainArtifactId}")
+      _ <- logger.info(
+        s"Executing post-update hook for ${update.groupId}:${update.mainArtifactId} with command " +
+          s"${hook.command.mkString_("'", " ", "'")}"
+      )
       repoDir <- workspaceAlg.repoDir(repo)
       result <- logger.attemptWarn.log("Post-update hook failed") {
         processAlg.execMaybeSandboxed(hook.useSandbox)(hook.command, repoDir)
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
@@ -39,7 +39,7 @@ final class HookExecutor[F[_]](implicit
     F: MonadThrow[F]
 ) {
   def execPostUpdateHooks(data: RepoData, update: Update): F[List[EditAttempt]] =
-    HookExecutor.postUpdateHooks
+    (HookExecutor.postUpdateHooks ++ data.config.postUpdateHooks.map(_.toHook))
       .filter { hook =>
         hook.groupId.forall(update.groupId === _)  &&
         hook.artifactId.forall(aid => update.artifactIds.exists(_.name === aid.name)) &&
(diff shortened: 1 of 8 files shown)


modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
```diff
@@ -51,7 +51,7 @@ final class HookExecutor[F[_]](implicit
 
   private def execPostUpdateHook(repo: Repo, update: Update, hook: PostUpdateHook): F[EditAttempt] =
     for {
-      _ <- logger.info(s"Executing post-update hook for ${hook.groupId}:${hook.artifactId.name}")
+      _ <- logger.info(s"Executing post-update hook for ${update.groupId}:${update.mainArtifactId}")
```
mzuehlke, line 54: What do you think about adding the `command` to the log ? Currently we would see multiple lines for the same update stating `"Executing post-update hook for gid:ai"` without mentioning the executed hook.

mzuehlke (approved):
With  a small remark regarding logging

modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
987Nabil: done

987Nabil:
@mzuehlke I tried to put this changes on a branch in my fork and let this run. I put settings into a file referenced via '--repo-config' but even when defining a hook without a group/artifact id, the hook does not run. Do you have an idea why? Because I don't 🤷

987Nabil:
Ok, I guess I found it. The monoid def of `RepoConfig` was missing a combine of the hooks

mzuehlke:
> Ok, I guess I found it. The monoid def of `RepoConfig` was missing a combine of the hooks

Good catch

mzuehlke (approved):

modules/core/src/main/scala/org/scalasteward/core/repoconfig/PostUpdateHookConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+import org.scalasteward.core.data.{ArtifactId, GroupId}
+import org.scalasteward.core.edit.hooks.PostUpdateHook
+import org.scalasteward.core.git.CommitMsg
+import org.scalasteward.core.util.Nel
+
+final case class PostUpdateHookConfig(
+    groupId: Option[GroupId],
+    artifactId: Option[String],
+    command: String,
+    useSandbox: Boolean,
```
fthomas, line 31: The `PostUpdateHookConfig#useSandbox` field should be removed and `PostUpdateHook#useSandbox` in `toHook` should be always set to `true`. It is a security risk to let users of Scala Steward decide if their commands should be executed in a sandbox or not. If the Scala Steward instance runs with `--enable-sandbox`, all user-defined commands should be executed in a sandbox.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/PostUpdateHookConfig.scala
```diff
@@ -0,0 +1,60 @@
...
+    commitMessage: String
+) {
+  def toHook: PostUpdateHook =
+    Nel
+      .fromList(command.split(' ').toList)
+      .fold(
+        throw new Exception("Post update hooks must have a command defined.")
```
fthomas, line 38: @987Nabil @mzuehlke The exception here would crash Scala Steward hard because it is not caught anywhere. Also `command.split(' ')` cannot handle arguments with spaces, for example if `command` is `sbt "argument1 with spaces" argument2` it would be split into `sbt`, `"argument1`, `with`, `spaces"`, `argument2`. I'm wondering if it would have been easier if `command` would be of type `Nel[String]` so that malformed commands would have been handled by the config file parser and we could support arguments with spaces.

987Nabil: @fthomas I was actually think about using `Nel[String]` but was not sure how it would be handled. And I had already an issue with white spaces...
I'll open a follow up pr soon™

PR #2435, opened 2021-12-26
Fix sandbox disabling

### LINKED ISSUE #2430: Disabling sandbox is broken
For private repos I check out master and run steward with `--disable-sandbox`. Now I get an error `Cannot run program "firejail"` which is afaik the sandbox tool. Seems to me related to #2414

987Nabil:
Fixes #2430


mzuehlke (approved):
Thanks Nabil for contributing to Scala Steward

PR #2441, opened 2021-12-30
User supplied hooks should always be executed in a sandbox

mzuehlke:
Spotted by @fthomas:https://github.com/scala-steward-org/scala-steward/pull/2434#discussion_r776669075


fthomas (approved):
Thank you, @mzuehlke!

987Nabil:
@fthomas @mzuehlke why should they?

PR #2446, opened 2022-01-02
Spread config parsing failures into the pull request body

danicheg:
This PR adds the spreading of config parsing failures into the pull request body. That should be useful for the OSS projects maintainers. It was born from the situation we got in  https://github.com/typelevel/keypool/pull/328.
The implementation is pretty straightforward, so I'm open to any suggestions/points about it.
And yeah, thanks for the Scala Steward! So cool to have it in our OSS ecosystem.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/StewardAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/StewardAlg.scala
@@ -80,7 +80,7 @@ final class StewardAlg[F[_]](config: Config)(implicit
       logger.attemptError.label(util.string.lineLeftRight(label), Some(label)) {
         F.guarantee(
           repoCacheAlg.checkCache(repo).flatMap { case (data, fork) =>
-            pruningAlg.needsAttention(data).flatMap {
+            pruningAlg.needsAttention(data.repoData).flatMap {
               _.traverse_(states => nurtureAlg.nurture(data, fork, states.map(_.update)))
             }
           },
--- a/modules/core/src/main/scala/org/scalasteward/core/data/RepoData.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/RepoData.scala
@@ -27,8 +27,8 @@ final case class RepoData(
 )
 
 final case class RepoDataWithMeta(
-  repoData: RepoData,
-  configParsingError: Option[String]
+    repoData: RepoData,
+    configParsingError: Option[String]
 ) {
   def repo: Repo = repoData.repo
   def cache: RepoCache = repoData.cache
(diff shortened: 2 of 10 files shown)


fthomas:
I think it is a good idea to notify users about malformed config files. :+1: In the current implementation, it would report parse errors only if it refreshed the `RepoCache` before creating PRs. But it often already has fresh `RepoCache`s and therefore wouldn't notify users about existing parse errors. If it should always work, the parse errors must be added to `RepoCache` [here](https://github.com/scala-steward-org/scala-steward/blob/00553da4912d791841326c76109514379a2bcee8/modules/core/src/main/scala/org/scalasteward/core/repocache/RepoCache.scala#L29).

Another idea would be to create an issue as soon as Scala Steward notices that the config is malformed. The advantage would be that users are notified soon after they have committed the malformed config and not (maybe much) later when Scala Steward creates a PR. We currently do not create issues, so I don't know how hard it would be to implement.

WDYT about using issues vs. PRs for this?

danicheg:
@fthomas Thanks for checking this!
Before I answer, wanna ask first - am I right that Scala Steward is updating its cache *only* before creating a PR? If so, then for frequently updated projects it will write about config parsing errors in almost every PR. Except for the case of multiple dependencies being updated, then yeah, only the first PR in a sequence will contain the config parsing error message. It's not so bad, I think. I especially was about not storing errors in the cache.
The functionality of creating issues with config parsing errors is sounded interesting, but that needs much-much more work. An algorithm should be the same as for PRs (keeps open only 1). And there is a case when projects keep closing their issues. So probably we should have a backport with messaging within the PR body.

danicheg:
@fthomas Hi! What do you think about this PR? Should I change something to get another review iteration?

fthomas:
> Before I answer, wanna ask first - am I right that Scala Steward is updating its cache _only_ before creating a PR?

Updating the cache and creating PRs are independent. Scala Steward updates the cache if it is out-of-date and if it creates PRs, the cache is up-to-date. But this does not necessarily mean that the cache is updated right before PRs are created. The cache could have been updated in a prior run.

danicheg:
Thus you are voting to store a parsing error in the cache, am I right? Sorry for probably my odd questions, but I'm actually didn't dig into Scala Steward internals well.

fthomas:
> Thus you are voting to store a parsing error in the cache, am I right?

Yes, I think that is the right thing to do regardless if the parsing errors are reported via issues or pull requests. Doing that should making `RepoDataWithMeta` unnecessary since the `RepoCache` is already available when creating PRs.

danicheg:
Thanks for your advice. I will dig into that in a few days probably.

modules/core/src/main/scala/org/scalasteward/core/repocache/RepoCache.scala
```diff
@@ -26,7 +26,8 @@ import org.scalasteward.core.repoconfig.RepoConfig
 final case class RepoCache(
     sha1: Sha1,
     dependencyInfos: List[Scope[List[DependencyInfo]]],
-    maybeRepoConfig: Option[RepoConfig]
+    maybeRepoConfig: Option[RepoConfig],
+    maybeConfigParsingError: Option[String]
```
fthomas, line 30: I would name this field `maybeRepoConfigParsingError` to make it crystal clear that it is related to the above `maybeRepoConfig`.

The rest looks great to me. :+1:

danicheg: Good, thanks. I've renamed that field.

fthomas (approved):

fthomas:
FYI: https://github.com/chipsalliance/chisel3/pull/2369 seems to be the first @scala-steward PR that reported a parsing error.

danicheg:
That's great :)

PR #2451, opened 2022-01-06
Ensure post update hooks having a `command` set via it's type

987Nabil:
- use `Nel[String]` instead of throwing an Exception

see https://github.com/scala-steward-org/scala-steward/pull/2434#discussion_r779472755_


987Nabil:
@fthomas I found the config syntax now a bit strange with having a list for the command. Is there a better way?

fthomas:
Thanks for the follow-up PR, @987Nabil!

> @fthomas I found the config syntax now a bit strange with having a list for the command.

I'm OK with the new syntax. It should be clearer now how to separate arguments.

> Is there a better way?

I've no idea how this could be made nicer.

mzuehlke (approved):

PR #2465, opened 2022-01-15
Attach credentials as http headers required for Gitlab's package registry authentication

cipriansofronia:
Gitlab's package registry auth mechanism is based on [headers](https://docs.gitlab.com/ee/user/packages/maven_repository/#authenticate-with-a-personal-access-token-in-maven). 
This change will add credentials as headers so artifacts from Gitlab's package registry can be downloaded. More details about the issue [here](https://gitter.im/scala-steward-org/scala-steward?at=61e2a64d5dc6213cd4e2580c).
I'm open to other suggestions as well.


modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -133,6 +133,7 @@ object CoursierAlg {
 
   private def toCoursierAuthentication(credentials: Credentials): Authentication =
     Authentication(credentials.user, credentials.pass)
+      .withHttpHeaders(Seq((credentials.user, credentials.pass)))
```
fthomas, line 136: @cipriansofronia This looks strange to me. If you are not using Gitlab's package registry, this will use a username as header name. What if the username is not a valid header name? I think this header should only be added if the username is `Private-Token`.

I also feel that Coursier should handle this case for us. It feels wrong that we need this code in Scala Steward.

cipriansofronia: I see, I can update the code to only add the header if the username is `Private-Token` or by some other mechanism but I think this should happen in ScalaSteward.. I don't see how Coursier can know that the registry is in gitlab and it should handle the authorization via headers. I'm open to other suggestions as well.

fthomas: But Scala Steward also doesn't know that the credentials are for a GitLab registry and Coursier could do the same as you proposed and only add the header if the username is `Private-Token`.

How are these credentials are handled in the sbt build?

cipriansofronia: In our projects we're using `sbt-gitlab` plugin in order to address this issue.

julienrf: From what I understand, Scala Steward does not know about the authentication credentials of Gitlab package registries. Such package registries are reported as follows by `stewardDependencies`:

~~~ json
{ "MavenRepository": { "name": "my-gitlab-repo", "location": "https://gitlab.my.company.com/api/v4/groups/my-group/-/packages/maven" } }
~~~

As far as I understand, the credentials are not exported by `stewardDependencies` because they are not provided by the `credentials` setting in sbt, but by `csrConfiguration`:

~~~ scala
val gitlabAuthentication =
  lmcoursier.definitions.Authentication(Seq("Private-Token" -> "my-token"))
resolvers += "my-gitlab-repo" at s"https://gitlab.my.company.com/api/v4/groups/my-group/-/packages/maven"
csrConfiguration ~= (_.addRepositoryAuthentication("my-gitlab-repo", gitlabAuthentication))
~~~

I believe Scala Steward should be made aware of the authentication process. However, as far as I understand, we’d have to modify the current model for credentials, which only supports Basic authentication currently.

Did I understand well? Would you be interested in such a contribution?

markvandertol: This PR of my addresses using `csrConfiguration` for sending headers as auth: https://github.com/scala-steward-org/scala-steward/pull/2533

cipriansofronia: In my scenario, adding auth to `csrConfiguration` will only make scala-steward work because inside the project `sbt-gitlab` is doing all the job, but I guess could add `csrConfiguration ~= (_.addRepositoryAuthentication("my-gitlab-repo", gitlabAuthentication))` as a [global config](https://www.scala-sbt.org/1.x/docs/Global-Settings.html#Basic+global+configuration+file) in my [scala-steward config](https://github.com/scala-steward-org/scala-steward/blob/main/docs/running.md#gitlab) so I don't update every project we have and also make scala-steward happy. So, I'll close this PR in favor of #2533.

cipriansofronia:
Closing in favor of #2533.

PR #2466, opened 2022-01-17
Add artifact migration for cakemix

Philippus:


fthomas:
LGTM. @Philippus, is version 1.2.0 already released? If yes, we can merge this right away. If not, we need to wait until it is released.

Philippus:
> LGTM. @Philippus, is version 1.2.0 already released? If yes, we can merge this right away. If not, we need to wait until it is released.

It was released two years ago already, so that should be fine. :)

PR #2470, opened 2022-01-22
Refactor UpdateAlg

fthomas:
Some refactorings in `UpdateAlg` to reduce the line count.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
@@ -17,6 +17,7 @@
 package org.scalasteward.core.update
 
 import cats.Monad
+import cats.data.OptionT
 import cats.syntax.all._
 import org.scalasteward.core.coursier.VersionsCache
 import org.scalasteward.core.data._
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
```diff
@@ -79,13 +50,39 @@ final class UpdateAlg[F[_]](implicit
     updates.flatMap(filterAlg.localFilterMany(repoConfig, _))
```
mzuehlke, line 50: You could inline `updates` and get rid of the curly braces, but wouldn't save a line

modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala
```diff
@@ -37,38 +37,9 @@ final class UpdateAlg[F[_]](implicit
...
-              CrossDependency(dependency.value),
-              newerVersions.map(_.value),
-              Some(artifactChange.groupIdAfter),
-              Some(artifactChange.artifactIdAfter)
-            )
-          )
-      }
+    findUpdateWithoutMigration(dependency, maxAge)
+      .orElse(findUpdateWithMigration(dependency, maxAge))
+      .value
```
mzuehlke, line 42: I really like the abstraction you have here now, easy to grasp what is going on.

mzuehlke (approved):
You made my initial change much better. Thanks a lot

Issue #2471, opened 2022-01-22
Syntax error when trying to load artifact migrations


lucaskjaero:
Hi Scala Steward team,
Thanks for maintaining a great tool, it saves a lot of time for me. Unfortunately, it looks like a recent update broke scala steward runs for me (and potentially others). Using `0.13.0`, I now get this error:
```
  2022-01-22 19:03:39,332 INFO   
    ____            _         ____  _                             _
   / ___|  ___ __ _| | __ _  / ___|| |_ _____      ____ _ _ __ __| |
   \___ \ / __/ _` | |/ _` | \___ \| __/ _ \ \ /\ / / _` | '__/ _` |
    ___) | (_| (_| | | (_| |  ___) | ||  __/\ V  V / (_| | | | (_| |
   |____/ \___\__,_|_|\__,_| |____/ \__\___| \_/\_/ \__,_|_|  \__,_|
   v0.13.0
   
  Error: Exception in thread "main" java.lang.Throwable: Failed to load artifact migrations from https://raw.githubusercontent.com/scala-steward-org/scala-steward/master/modules/core/src/main/resources/artifact-migrations.conf
  Error: 	at org.scalasteward.core.update.artifact.ArtifactMigrationsLoader$$anonfun$decodeMigrations$1.applyOrElse(ArtifactMigrationsLoader.scala:50)
  Error: 	at org.scalasteward.core.update.artifact.ArtifactMigrationsLoader$$anonfun$decodeMigrations$1.applyOrElse(ArtifactMigrationsLoader.scala:50)
  Error: 	at scala.PartialFunction$AndThen.applyOrElse(PartialFunction.scala:263)
  Error: 	at cats.ApplicativeError.$anonfun$recoverWith$1(ApplicativeError.scala:106)
  Error: 	at adaptErr$extension @ org.scalasteward.core.update.artifact.ArtifactMigrationsLoader.decodeMigrations(ArtifactMigrationsLoader.scala:50)
  Error: 	at blocking @ org.scalasteward.core.io.FileAlg$$anon$1.$anonfun$readSource$1(FileAlg.scala:138)
  Error: 	at blocking @ org.scalasteward.core.io.FileAlg$$anon$1.readUri(FileAlg.scala:134)
  Error: 	at make @ org.http4s.okhttp.client.OkHttpBuilder$.defaultOkHttpClient(OkHttpBuilder.scala:198)
[...]

PR #2472, opened 2022-01-22
Add initialVersion again to artifact-migrations.conf

fthomas:
These are needed so that older versions of Scala Steward can read this
file again.

Closes: #2471


lucaskjaero:
Thank you!

PR #2455, opened 2022-01-08
check if the version of the migrated artifact exists

mzuehlke:
Check if the version of the migrated artifact exists using the `versionCache`. As now the version of the migrated artifact is detected `ArtifactChange#initialVersion` is no longer needed.

fixes #2450 

This should prevent problems like in #2447

PR #2474, opened 2022-01-22
Add new version of artifact-migrations file

fthomas:
This adds a new default file for artifacts migrations that does not contain the `initialVersion` fields. Since older Scala Steward instances load the list of artifact migrations from this repository, removing fields is a backward-incompatible change so that we cannot remove them in the old default file.

See #2455 and #2472 for context.

@mzuehlke What do you think about this change. Is there a better way to keep older Scala Steward versions happy and still getting rid of the `initialVersion` field?

### DIFF
--- a/docs/artifact-migrations.md
+++ b/docs/artifact-migrations.md
@@ -31,6 +31,6 @@ to be greater than the latest version with the old groupId / old artifact id.
 
 Pull requests that added artifact migrations can be found [here][migration-prs].
 
-[migrations]: https://github.com/scala-steward-org/scala-steward/blob/master/modules/core/src/main/resources/artifact-migrations.conf
+[migrations]: https://github.com/scala-steward-org/scala-steward/blob/master/modules/core/src/main/resources/artifact-migrations.v2.conf
 [migration-prs]: https://github.com/scala-steward-org/scala-steward/pulls?q=label%3Aartifact-migration
 [HOCON]: https://github.com/lightbend/config/blob/master/HOCON.md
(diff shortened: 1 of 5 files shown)


docs/artifact-migrations.md
```diff
@@ -31,6 +31,6 @@ to be greater than the latest version with the old groupId / old artifact id.
 
 Pull requests that added artifact migrations can be found [here][migration-prs].
 
-[migrations]: https://github.com/scala-steward-org/scala-steward/blob/master/modules/core/src/main/resources/artifact-migrations.conf
+[migrations]: https://github.com/scala-steward-org/scala-steward/blob/master/modules/core/src/main/resources/artifact-migrations.v2.conf
```
mzuehlke, line 34: In this file are still reference to `initialVersion`. I had forgotten to remove them I guess.

mzuehlke (approved):
I can't think of any other solution. 
Old clients need the old format and only newer clients can handle the format without `initialVersion`. I see you added `sbt-twirl` already only to `v2` so PRs will only be created after a release.  👍

docs/artifact-migrations.md
fthomas: Good point. I removed the reference in the latest commit.

fthomas:
> I see you added `sbt-twirl` already only to `v2` so PRs will only be created after a release.

That's right for Scala Steward instances that do not track master. I think I'm going to publish 0.14.0 because of this soon.

PR #2475, opened 2022-01-23
dynamic sbt-scalafix version

### LINKED ISSUE #2357: Do not hardcode sbt-scalafix version
We currently hardcode the version of sbt-scalafix that we add to the global sbt plugins temporarily while applying Scalafix migrations: 
https://github.com/scala-steward-org/scala-steward/blob/cc04fcf0e92a469b600d647631d1a417b1bf035d/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/package.scala#L41

This is not ideal for Scala Steward instances that do not track the latest changes from this repository. If a new sbt-scalafix version is released that supports newer Scala versions, Scala Steward instances that not yet include that latest sbt-scalafix version might not successfully apply Scalafix migrations in projects that are already on the new Scala version. I think it would be better to use the latest sbt-scalafix version from the `VersionsCache`.

mzuehlke:
Retrieve the latest `sbt-scalafix` plugin version from coursier (`versionCache`) instead of relying on a hardcoded version.

Fixes #2357

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/package.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/package.scala
@@ -18,8 +18,8 @@ package org.scalasteward.core.buildtool
 
 import cats.Functor
 import cats.syntax.all._
-import org.scalasteward.core.buildtool.sbt.data.SbtVersion
-import org.scalasteward.core.data.{ArtifactId, Dependency, GroupId, Version}
+import org.scalasteward.core.buildtool.sbt.data.{SbtVersion, ScalaVersion}
+import org.scalasteward.core.data.{ArtifactId, Dependency, GroupId, Resolver, Scope, Version}
 import org.scalasteward.core.io.{FileAlg, FileData}
 
 package object sbt {
@@ -35,10 +35,22 @@ package object sbt {
       Dependency(sbtGroupId, sbtArtifactId, sbtVersion.value)
     }
 
-  val scalaStewardScalafixSbt: FileData =
+  val sbtScalaFixDependency: Scope[Dependency] =
+    Scope(
+      Dependency(
+        GroupId("ch.epfl.scala"),
+        ArtifactId("sbt-scalafix"),
+        "",
+        Some(SbtVersion("1.0")),
+        Some(ScalaVersion("2.12"))
+      ),
+      List(Resolver.mavenCentral)
+    )
+
+  def scalaStewardScalafixSbt(version: String): FileData =
     FileData(
       "scala-steward-scalafix.sbt",
-      """addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.34")"""
+      s"""addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "$version")"""
     )
 
   def scalaStewardScalafixOptions(scalacOptions: List[String]): FileData = {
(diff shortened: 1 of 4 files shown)


fthomas:
This is great!

modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/package.scala
```diff
@@ -35,10 +35,22 @@ package object sbt {
...
+  val sbtScalaFixDependency: Scope[Dependency] =
+    Scope(
+      Dependency(
+        GroupId("ch.epfl.scala"),
+        ArtifactId("sbt-scalafix"),
+        "",
+        Some(SbtVersion("1.0")),
+        Some(ScalaVersion("2.12"))
+      ),
+      List(Resolver.mavenCentral)
```
fthomas, line 47: This should be `Config#defaultResolver` which is `Resolver.mavenCentral` by default.

modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/package.scala
```diff
@@ -35,10 +35,22 @@ package object sbt {
...
+    )
+
+  def scalaStewardScalafixSbt(version: String): FileData =
     FileData(
       "scala-steward-scalafix.sbt",
-      """addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.34")"""
+      s"""addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "$version")"""
```
fthomas, line 53: We can now also remove the plugin from the build:
https://github.com/scala-steward-org/scala-steward/blob/b0699bf8b234bfbdb6444c8d4e5558d28b062261/project/plugins.sbt#L16-L17

mzuehlke:
Thanks for the comments @fthomas 
I'm not sure how to get the test reliable. Depending on wether the data is in the cache or not the executed command are different. One idea wold be to run the cmd `sbtAlg.runMigration...` twice.
The 1st to make sure the cache is properly filled but don't look at the state afterwards and the 2nd to assert the expect change. Doesn't look that nice. Or reduce the TTL to 0 then we in tests always expect value to be written to the cache. But this would make the test slower....

fthomas:
Another option could be to add a
`/tmp/scala-steward/workspace/store/versions/v2/https/repo1.maven.org/maven2/ch/epfl/scala/sbt-scalafix_2.12_1.0/versions.json` to the initial `MockState` so that `VersionsCache` always reads the versions from that file.

modules/core/src/test/scala/org/scalasteward/core/buildtool/sbt/SbtAlgTest.scala
```diff
@@ -133,4 +151,13 @@ class SbtAlgTest extends FunSuite {
     )
     assertEquals(state, expected)
   }
+
+  private val sbtScalafixVersionJson =
+    s"""|{
+        |  "updatedAt" : 9999999999999,
```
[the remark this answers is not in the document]
mzuehlke: thanks for the good and fast feedback on the initial versions

bjaglin:
Cool feature! Does this pick up RC releases as well? I am curious to know if/how I could test the impact of https://github.com/scalacenter/sbt-scalafix/releases/tag/v0.10.0-RC1 (particularly for https://github.com/scalacenter/sbt-scalafix/pull/292).

PR #2481, opened 2022-01-24
Rename `master` to `main`

fthomas:
I'd like to rename the default branch of this repo from `master` to `main`. If this is done, I'd check that older Scala Steward versions can still read the default `artifact-migrations.conf` and `scalafix-migrations.conf` files using the old URLs. I think GitHub redirects those.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
   publish:
     name: Publish Artifacts
     needs: [build]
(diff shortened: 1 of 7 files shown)


fthomas:
The branch has been renamed and I've checked that an older version can still read the conf files from this repo.

These commands can be used to update a local clone:
```
git branch -m master main
git fetch origin
git branch -u origin/main main
git remote set-head origin -a
```

PR #2492, opened 2022-01-30
Add custom `PostUpdateHook` for sbt-typelevel

armanbilge:
This adds a custom `PostUpdateHook` for sbt-typelevel, which calls a `tlPrePrBotHook` command alias. The idea is that:
1. Since sbt-typelevel subsumes several other plugins, it may require multiple post-update steps to be run.
2. We can easily update the command in sbt-typelevel without needing to change Scala Steward.
3. Downstream plugins can update the alias with their own custom steps, without needing to add their own `PostUpdateHook` to Steward.

Additional background in:
- https://github.com/typelevel/sbt-typelevel/pull/111.

Let me know what you think. Thanks in advance! :)

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
@@ -124,7 +124,7 @@ object HookExecutor {
 
   private def sbtTypelevelHook(
       groupId: GroupId,
-      artifactId: ArtifactId,
+      artifactId: ArtifactId
   ): PostUpdateHook =
     PostUpdateHook(
       groupId = Some(groupId),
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
```diff
@@ -72,11 +72,14 @@ object HookExecutor {
     (GroupId("com.codecommit"), ArtifactId("sbt-spiewak-bintray")),
     (GroupId("io.github.nafg.mergify"), ArtifactId("sbt-mergify-github-actions")),
     (GroupId("io.chrisdavenport"), ArtifactId("sbt-davenverse")),
-    (GroupId("org.http4s"), ArtifactId("sbt-http4s-org")),
-    (GroupId("org.typelevel"), ArtifactId("sbt-typelevel")),
     (GroupId("org.typelevel"), ArtifactId("sbt-typelevel-ci-release"))
```
armanbilge, line 75: `sbt-typelevel-ci-release` is a lower-level plugin which only needs the sbt-github-actions post-update step.

fthomas (approved):
LGTM

mzuehlke:
@armanbilge In case you didn't know since #2434 Scala Steward can have repo specific post update hooks.
But each user of the `sbt-typelevel` plugin would have to add them herself/himself.

armanbilge:
@mzuehlke thanks, that config works well for single repositories but is difficult to maintain across orgs.

For example, sbt-http4s-org extends sbt-typelevel and would like to add its own customized post-update hooks for all the repositories in http4s org. With these changes, all it needs to do is update the `tlPrePrBotHook` command alias with its own commands.

PR #2498, opened 2022-02-01
Don't use versions that only consist of a hash.

### LINKED ISSUE #2483: Prevent updates from "normal" versions to hashes
See #2359 and https://github.com/scalameta/metals-gitpod-sample/pull/85 for example.

bloop-config and sbt-bloop are example artifacts that have a few hash versions:
* https://repo1.maven.org/maven2/ch/epfl/scala/bloop-config_2.12/
* https://repo1.maven.org/maven2/ch/epfl/scala/sbt-bloop_2.12_1.0/

mzuehlke:
@fthomas do you think this is too drastic

Fixes #2483

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
@@ -63,10 +63,9 @@ final case class Version(value: String) {
           v.alnumComponents === alnumComponents ||
           // Do not select a version with hash if this version contains no hash.
           (v.containsHash && !containsHash) ||
+          (v.isOnlyHash && !isOnlyHash) ||
           // Don't select "versions" like %5BWARNING%5D.
-          !v.startsWithLetterOrDigit ||
-          // Don't select "versions" that are only a hash
-          v.isOnlyHash
+          !v.startsWithLetterOrDigit
         }.sorted
       }
       .lastOption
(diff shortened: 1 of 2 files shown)


fthomas:
> do you think this is too drastic

I'm not sure. It would prevent updates with very long versions that only consists of digits and are not dates. For example, a bump like `10000000 -> 20000000` would be prevented. If we change it slightly from `v.isOnlyHash` to `isOnlyHash =!= v.isOnlyHash` (or even `!isOnlyHash && v.isOnlyHash`), we would allow this bump but still prevent things like `1.4.12 -> 1032048a`.

modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
```diff
@@ -63,12 +64,17 @@ final case class Version(value: String) {
...
+          !v.startsWithLetterOrDigit ||
+          // Don't select "versions" that are only a hash
+          v.isOnlyHash
         }.sorted
       }
       .lastOption
   }
 
+  private def isOnlyHash: Boolean =
+    Rfc5234.hexdig.rep(8).string.filterNot(startsWithDate).parseAll(value).isRight
```
fthomas, line 76: Does this only match hashes with exactly 8 digits? If yes, I would vote for also matching longer hashes.

fthomas (approved):

fthomas:
Here is the next challenge: https://github.com/spotify/scio/pull/4235 - a bump from `1.4.12` to `7d2bf0af+20171218-1522` ;-)

Maybe we should redefine `containsHash` such that it does not require a leading separator. :thinking:

PR #2515, opened 2022-02-08
Add granular frequency control by groupId and artifactId

### LINKED ISSUE #1581: Time-based delay configuration for new dependencies

exoego:
Closes #1581
May supersedes #2322

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -39,7 +39,7 @@ pullRequests.frequency = "7 days"
 # `pullRequests.frequency` is used as default frequency, if none match.
 #
 # Default: empty `[]`
-pullRequests.frequencyPerGroup = [
+pullRequests.perGroup = [
     { frequency = "30 days", pattern = { groupId = "com.example", artifactId = "foo", version = "2." } },
     { frequency = "30 days", pattern = { groupId = "com.example", artifactId = "bar" } },
     { frequency = "14 days", pattern = { groupId = "com.example" } },
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -32,19 +32,6 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 #pullRequests.frequency = "0 0 ? * 3" # every thursday on midnight
 pullRequests.frequency = "7 days"
 
-# Granular control of frequency based on the given pattern.
-# Each pattern must have `groupId`, and may have `artifactId` and `version`.
-# First-matched entry is used.
-# More-specific entry should be placed before less-specific entry.
-# `pullRequests.frequency` is used as default frequency, if none match.
-#
-# Default: empty `[]`
-pullRequests.perGroup = [
-    { frequency = "30 days", pattern = { groupId = "com.example", artifactId = "foo", version = "2." } },
-    { frequency = "30 days", pattern = { groupId = "com.example", artifactId = "bar" } },
-    { frequency = "14 days", pattern = { groupId = "com.example" } },
-]
-
 # Only these dependencies which match the given patterns are updated.
 #
 # Each pattern must have `groupId`, and may have `artifactId` and `version`.
@@ -106,6 +93,28 @@ postUpdateHooks = [{
   groupId = "com.github.sbt",
   artifiactId = "sbt-protobuf"
 }]
+
+# You can override some config options for dependencies that matches the given pattern.
+# Currently, "pullRequests" can be overridden.  
+# Each pattern must have `groupId`, and may have `artifactId` and `version`.
+# First-matched entry is used.
+# More-specific entry should be placed before less-specific entry.
+#
+# Default: empty `[]`
+dependencyOverrides = [
+  {
+    pattern = { groupId = "com.example", artifactId = "foo", version = "2." },
+    pullRequests = { frequency = "1 day" },
+  },
+  {
+    pattern = { groupId = "com.example", artifactId = "foo" },
+    pullRequests = { frequency = "30 day" },
+  },
+  {
+    pattern = { groupId = "com.example" },
+    pullRequests = { frequency = "14 day" },
+  }
+]
 ```
 
 The version information given in the patterns above can be in two formats:
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -103,15 +103,15 @@ postUpdateHooks = [{
 # Default: empty `[]`
 dependencyOverrides = [
   {
-    pattern = { groupId = "com.example", artifactId = "foo", version = "2." },
+    dependency = { groupId = "com.example", artifactId = "foo", version = "2." },
     pullRequests = { frequency = "1 day" },
   },
   {
-    pattern = { groupId = "com.example", artifactId = "foo" },
+    dependency = { groupId = "com.example", artifactId = "foo" },
     pullRequests = { frequency = "30 day" },
   },
   {
-    pattern = { groupId = "com.example" },
+    dependency = { groupId = "com.example" },
     pullRequests = { frequency = "14 day" },
   }
 ]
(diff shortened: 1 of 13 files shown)


fthomas:
This looks great on a first glance, I'll take a closer look later.

modules/core/src/main/scala/org/scalasteward/core/util/dateTime.scala
```diff
@@ -31,19 +31,25 @@ object dateTime {
...
+    }
+
+  private def showDurationList(d: FiniteDuration): List[String] =
+    splitDuration(d).map(d1 => d1.length.toString + symbol(d1.unit))
+
+  def showDuration(d: FiniteDuration): String =
+    showDurationList(d).mkString(" ")
+
+  def showDurationShort(d: FiniteDuration): String =
+    showDurationList(d).take(3).mkString(" ")
```
fthomas, line 52: I like this change. When we show the duration, there is probably never the need to be as precise as we are now. I would just change the old `showDuration` to only use the first three elements so that we don't need a new `showDurationShort`.

Oh, and it would be nice to have this change in a separate PR, since this feature is independent of the change in `showDuration` and the other PR could be merged right away.

exoego: 👍 
Extracted PR https://github.com/scala-steward-org/scala-steward/pull/2517

docs/repo-specific-configuration.md
```diff
@@ -32,6 +32,17 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 #pullRequests.frequency = "0 0 ? * 3" # every thursday on midnight
 pullRequests.frequency = "7 days"
 
+# Granular control of frequency based on group id and optional artifact id.
+# First-matched entry is used.
+# More-specific entry should be placed before less-specific entry.
+# `pullRequests.frequency` is used as default frequency, if none match.
+#
+# Default: empty
+pullRequests.frequencyPerGroup = [
```
exoego, line 41: Maybe `pullRequests.perGroup` for new feature ? 🤔

modules/core/src/main/scala/org/scalasteward/core/repoconfig/GroupPullRequestFrequency.scala
```diff
@@ -0,0 +1,36 @@
...
+import io.circe.generic.semiauto.deriveCodec
+import org.scalasteward.core.data.{GroupId, Update}
+
+final case class GroupPullRequestFrequency(
+    frequency: PullRequestFrequency,
+    groupId: GroupId,
+    artifactId: Option[String] = None
+) {
+  def matches(update: Update.Single): Boolean =
+    this.groupId === update.groupId && this.artifactId.forall(_ === update.mainArtifactId)
```
fthomas, line 30: I'm wondering if we can use the already existing `UpdatePattern` class here since that already has the functionality to match `Update.Single`. That class would replace the `groupId` and `artifactId` fields here.

docs/repo-specific-configuration.md
fthomas: Do you mean something like this
```
pullRequests.perGroup = [
  { pattern = { groupId = "com.example", artifactId = "foo" }, config = { frequencey = "14 days", <other fields> } },
  ...
]
```
?
It looks a bit complicated but would allow to add other fields besides `frequency` that could be set per artifact.

exoego: Yes, I mean that.
Maybe more flat 
```
{ groupId = "com.example", artifactId = "foo", frequencey = "14 days", foo = "bar", buz = 42 }
```

modules/core/src/main/scala/org/scalasteward/core/repoconfig/GroupPullRequestFrequency.scala
exoego: That sounds nice and consistent. Done in https://github.com/scala-steward-org/scala-steward/pull/2515/commits/370f3f84b76397c2c2cc13a810bb1830c77ad7f5

docs/repo-specific-configuration.md
exoego: One possible use case is https://github.com/scala-steward-org/scala-steward/issues/1879
```
  pattern = { groupId = "com.example", artifactId = "foo" },
  frequencey = "14 days",
  resolver = "https://repo1.maven.org/maven2/",
}
```

mzuehlke: `frequencyPerGroup` is a bit misleading, as it can be more specific.
And it would be great to find a better name than `pattern`

exoego: > And it would be great to find a better name than pattern
Maybe `matcher`, or  `condition`.
I think `pattern` is not a bad choice since it corresponds to `case class UpdatePattern`.

exoego: Renamed to more generic name `pullRequests.perGroup`.

fthomas: I think the `perGroup` change goes into the right direction since it allows to override more than one config option for dependencies that match a given pattern. How about going one step further and move the `GroupPullRequestConfig` class from `PullRequestConfig` to `RepoConfig` so that we could potentially override all config options for matching dependencies?

I imagine that it would be similar how one can override options in `.scalafmt.conf` for specific files via [`fileOverride`](https://scalameta.org/scalafmt/docs/configuration.html#fileoverride). For example, we could specify a different pull request frequency and commit message for a given dependency via a new `RepoConfig.dependencyOverride` field like this:
```
dependencyOverride = [
  {
    dependency: { groupId = ..., artifactId = ...},
    override: { pullRequests.frequency = ..., commits.message = ... }
  }, ...
]
```
The only problem I see here is that the word `override` is not appropriate
* if we introduce options that can ony be set in a `override` section
* since it implies that all options can be overridden, while it is only true for options where overriding is implemented

WDYT @exoego, @mzuehlke?

exoego: Done in https://github.com/scala-steward-org/scala-steward/pull/2515/commits/4605e78153d6bfb11d43703afb5229b126055b81

I keep this PR for only `pullRequests` override.
`commits` support can be added in subsequent PR. 

> The only problem I see here is that the word override is not appropriate

As an user, it won't be a problem for me if it is documented that which options can be overridden

mzuehlke: You went with slightly different wording:
```
dependencyOverride = [
  {
    pattern = { groupId = ..., artifactId = ...},
    pullRequests.frequency = ..., 
    commits.message = ... // not there, yet
  }, ...
]
```
How about renaming `pattern` to `dependency` as in the comment from @fthomas. But for the rest keep your structure (without the `override` as `dependencyOverride` signals that we re dealing with overrides:
```
dependencyOverride = [
  {
    dependency = { groupId = ..., artifactId = ...},
    pullRequests.frequency = ..., 
    commits.message = ... // not there, yet
  }, ...
]
```
Other than this, looks good.

exoego: Done in https://github.com/scala-steward-org/scala-steward/pull/2515/commits/cd549f040d5dc7a57ebb9acb2831271938cbca71

mzuehlke (approved):

docs/repo-specific-configuration.md
fthomas: This LGTM, @exoego. And sorry for the late reply.

fthomas (approved):

fthomas:
https://github.com/scalacenter/scalafix/pull/1569 is a cool usage of this feature. Scalameta updates are proposed ASAP while all other updates are proposed twice a month.

PR #2522, opened 2022-02-10
Adapt the check for hashes in versions

### LINKED ISSUE #2502: Prevent updates to versions that start with hashes
See https://github.com/scala-steward-org/scala-steward/pull/2498#issuecomment-1028090848.

mzuehlke:
From checking if the complete version value is a hash to only checking if it contains a hash (at least 8 characters long).
The hash component detected by the parser only requires 6 byte, but in addition some leading separator.

Fixes #2502
Improves predecessor #2498
Original issue #2483

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
@@ -87,8 +87,8 @@ final case class Version(value: String) {
   private def containsHash: Boolean =
     components.exists {
       case _: Version.Component.Hash => true
-      case _ => Rfc5234.hexdig.rep(8).string.filterNot(startsWithDate).parse(value).isRight
-    }
+      case _                         => false
+    } || Rfc5234.hexdig.rep(8).string.filterNot(startsWithDate).parse(value).isRight
 
   private[this] def alnumComponentsWithoutPreRelease: List[Version.Component] =
     alnumComponents.takeWhile {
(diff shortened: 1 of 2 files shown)


exoego (approved):
LGTM

modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
```diff
@@ -91,7 +87,7 @@ final case class Version(value: String) {
   private def containsHash: Boolean =
     components.exists {
       case _: Version.Component.Hash => true
-      case _                         => false
+      case _ => Rfc5234.hexdig.rep(8).string.filterNot(startsWithDate).parse(value).isRight
     }
```
fthomas, line 91: If the version contains no hash this will check for hashes in `value` on every component. How about moving this after the `exists` call:
```scala
    components.exists {
      case _: Version.Component.Hash => true
      case _                         => false
    } || Rfc5234.hexdig.rep(8).string.filterNot(startsWithDate).parse(value).isRight
```
?

mzuehlke: makes sense 👍

fthomas:
The fun isn't over yet: https://github.com/davegurnell/unindent/pull/69 :-)

mzuehlke:
> from `7d2bf0af+20171218-1522` to `7e30aec7`

This is an update from a version with a hash to another version with a hash.
So versions that are only a hash are worse then versions that only contain a hash (in addition to other numbers)  🤔

fthomas:
Maybe we should just ignore versions that start with a hash. A bump like `7d2bf0af+20171218-1522 to 8d2bf0af+20171218-1522` would not make any sense either, although both versions contain other numbers besides the hash.

PR #2528, opened 2022-02-14
feat:  Adding support for setting required number of reviewers on a S…

### LINKED ISSUE #2519: Support for setting required number of reviewers on GitLab merge requests
This feature would be useful for helping automate merging ScalaSteward MRs in the context where most other MRs are subject to minimum reviewer rules.  

This would be mostly to support running ScalaSteward over private repositories where credentials provided through runtime parameters would require the sufficient privileges for this operation through the GitLab API.

dkichler:
…calaSteward managed GitLab MR (#2519)

Closes #2519


exoego:
@dkichler Sorry for the late response. Could you fix the conflict?
Code looks great to me.

dkichler:
Conflict addressed.

docs/help.md
```diff
@@ -70,6 +70,9 @@ Options and flags:
         Whether to assign the default reviewers to a bitbucket pull request; default: false
     --gitlab-merge-when-pipeline-succeeds
         Whether to merge a gitlab merge request when the pipeline succeeds
+    --gitlab-required-reviewers <integer>
+        When set, the number of required reviewers for a merge request will be set to this number (non-negative integer).  Is only used in the context of gitlab-merge-when-pipeline-succeeds being enabled, and requires that the configured access token have the appropriate privileges.
+    --gitlab-required-reviewers <integer>
```
exoego, line 75:
```suggestion
```

dkichler:
@exoego I think this one is finally ready to go in.

exoego:
Thanks!

henricook:
@exoego Thanks for this. OOI why does it only take effect when merge if pipeline succeeds is in play?

This would be helpful for me because we like to allow only one approver for dependency bumps instead of two (for all other code changes)

exoego:
Sorry, I don't use GitLab so I can not tell.

dkichler:
> @exoego Thanks for this. OOI why does it only take effect when merge if pipeline succeeds is in play?
> 
> This would be helpful for me because we like to allow only one approver for dependency bumps instead of two (for all other code changes)

Hi Henri.  So the `mergeWhenPipelineSucceeds` flag perhaps has a slightly misleading name.  If you inspect the code, the flag is used to [gate whether the MR in question is automatically marked to be merged](https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/scala/org/scalasteward/core/vcs/gitlab/GitLabApiAlg.scala#L193) by the update algorithm.  When set, the MR will also be adjusted with the configured number of required reviewers, as per this MR, prior to attempting the merge.  The flag was named as such b/c in the attempt to merge, the query param for [merging if pipeline succeeds](https://docs.gitlab.com/ee/api/merge_requests.html#merge-a-merge-request) is [hardcoded](https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/scala/org/scalasteward/core/vcs/gitlab/Url.scala#L41-L43).  

I didn't want to change the semantics of the existing `mergeWhenPipelineSucceeds` flag, to avoid changing behaviour of any existing implementations, but I think it would be more appropriately named something like `mergeUponAllCriteriaMet`.   Certainly there is some room to make the Gitlab algo implementation a bit cleaner/more explicit/configurable, but even as is I would expect it should account for most needs/use-cases.  Given my explanation, does it still cover your use-case?

henricook:
Thanks David. I think what you're saying here (to read it back) is that `mergeWhenPipelineSucceeds` can be set to true, but won't error if the MR is not in a _ready to merge_ state (e.g. it still has outstanding approvals), so enabling it in my situation won't hurt anything - because if approvers are set to 1, 1 approver will be outstanding and no auto merge will be attempted. 

Assuming I've understood that right then it sounds good for my use case, thanks!

henricook:
Lovely, thanks!

henricook:
Sadly this didn't work out for me, but I don't blame Steward. The /approvals API that this MR used is only available in Gitlab Premium and we're currently on the old 'Gitlab Starter' package (that's no longer sold, but we're still in subscription for).

The only way to change approvers is by POSTing to the MR and the approvals API 404s until we upgrade.

image

dkichler:
Aww shucks, indeed it appears this feature _does_ depend on Gitlab Premium.  I think at the very least the documentation for the flag ought to be updated to state as much.  To be fair, the Gitlab API is a bit of a mess to work with in the first place, it looks like the endpoint this feature is based on has [already been deprecated ](https://docs.gitlab.com/ee/api/merge_request_approvals.html#change-approval-configuration-deprecated) and will be removed in an upcoming release.  

Unfortunately, the [replacement API ](https://docs.gitlab.com/ee/api/merge_request_approvals.html#create-merge-request-level-rule)(create/update/delete MR level rules) appears to also only have support in Premium 😞.

PR #2533, opened 2022-02-20
Collect HTTP Headers from Coursier for authentication

markvandertol:
This PR lets Scala Steward also pick up custom headers used for authentication. This is for example needed for Gitlab. I think looking at `csrConfiguration.value.authenticationByRepositoryId` is the cleanest and most generic way to achieve this.

Example snippet of a .sbt file from a project that sets a custom HTTP-header for authentication with Coursier:
```
resolvers += "gitlab-repo" at "https://gitlab.example.com/api/v4/groups/1/-/packages/maven",
csrConfiguration ~= _.addRepositoryAuthentication("gitlab-repo", Authentication(Seq( ("private-token", "xxx") )))
```

It works for SBT and Mill, but doesn't work for Maven

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -199,7 +199,7 @@ There is multiple articles on how to run Scala Steward on-premise:
   
 #### GitLab
 
-The following describes a setup using GitLab Docker runner, which you have to setup seperately.
+The following describes a setup using GitLab Docker runner, which you have to set up separately.
 
 1. create a "scalasteward" user in GitLab
 2. assign that user "Developer" permissions in every project that should be managed by Scala Steward
@@ -260,6 +260,33 @@ echo "${SCALA_STEWARD_TOKEN}"
 7. add the `repos.md` file 
 8. (*optional*) create a new schedule to trigger the pipeline on a daily/weekly basis
 
+Scala Steward is compatible with Coursier authentication using headers. To authenticate
+using the [Gitlab CI/CD job token](https://docs.gitlab.com/ee/ci/jobs/ci_job_token.html), while also supporting your own private token when performing
+local development, use the following snippet:
+```scala
+import lmcoursier.CoursierConfiguration
+import lmcoursier.definitions.Authentication
+
+lazy val gitlabToken: Option[(String, String)] = {
+  //The Gitlab runner sets CI_JOB_TOKEN automatically as part of running inside a build job
+  val jobToken = sys.env.get("CI_JOB_TOKEN").map(t => ("Job-Token", t)) 
+  //When running on your local machine, set the environment variable GITLAB_PRIVATE_TOKEN
+  val privateToken = sys.env.get("GITLAB_PRIVATE_TOKEN").map(t => ("Private-Token", t))
+
+  jobToken.orElse(privateToken)
+}
+
+def addGitlabToken(current: CoursierConfiguration): CoursierConfiguration = {
+  gitlabToken.fold(current) { token =>
+    current.addRepositoryAuthentication("gitlab-repo", Authentication(Seq(token)))
+  }
+}
+
+resolvers += "gitlab-repo" at s"https://gitlab.example.com/api/v4/groups/1/-/packages/maven"
+csrConfiguration ~= addGitlabToken
+updateClassifiers / csrConfiguration ~= addGitlabToken
+updateSbtClassifiers / csrConfiguration ~= addGitlabToken
+```
 
 [scalafixmigrations]: https://github.com/scala-steward-org/scala-steward/blob/main/docs/scalafix-migrations.md
 [artifactmigrations]: https://github.com/scala-steward-org/scala-steward/blob/main/docs/artifact-migrations.md
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -126,15 +126,35 @@ object CoursierAlg {
 
   private def toCoursierRepository(resolver: Resolver): Either[String, coursier.Repository] =
     resolver match {
-      case Resolver.MavenRepository(_, location, creds) =>
-        Right(coursier.maven.MavenRepository.apply(location, creds.map(toCoursierAuthentication)))
-      case Resolver.IvyRepository(_, pattern, creds) =>
+      case Resolver.MavenRepository(_, location, creds, headers) =>
+        Right(
+          coursier.maven.MavenRepository
+            .apply(location, toCoursierAuthentication(creds, headers))
+        )
+      case Resolver.IvyRepository(_, pattern, creds, headers) =>
         coursier.ivy.IvyRepository
-          .parse(pattern, authentication = creds.map(toCoursierAuthentication))
+          .parse(pattern, authentication = toCoursierAuthentication(creds, headers))
     }
 
-  private def toCoursierAuthentication(credentials: Credentials): Authentication =
-    Authentication(credentials.user, credentials.pass)
+  private def toCoursierAuthentication(
+      credentials: Option[Credentials],
+      headers: List[Resolver.Header]
+  ): Option[Authentication] =
+    if (credentials.isEmpty && headers.isEmpty) {
+      None
+    } else {
+      Some(
+        new Authentication(
+          credentials.fold("")(_.user),
+          credentials.map(_.pass),
+          headers.map(h => (h.key, h.value)),
+          optional = false,
+          None,
+          httpsOnly = true,
+          passOnRedirect = false
+        )
+      )
+    }
 
   private def getParentDependency(project: Project): Option[coursier.Dependency] =
     project.parent.map { case (module, version) =>
--- a/modules/core/src/test/scala/org/scalasteward/core/update/PruningAlgTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/update/PruningAlgTest.scala
@@ -105,7 +105,8 @@ class PruningAlgTest extends FunSuite {
           |        {
           |          "MavenRepository" : {
           |            "name" : "public",
-          |            "location" : "https://foobar.org/maven2/"
+          |            "location" : "https://foobar.org/maven2/",
+          |            "headers" : []
           |          }
           |        }
           |      ]
@@ -192,7 +193,7 @@ class PruningAlgTest extends FunSuite {
             DependencyInfo("org.scala-lang".g % "scala-library".a % "2.12.14", List("build.sbt")),
             DependencyInfo("org.scala-lang".g % "scala-library".a % "2.13.5", List("build.sbt"))
           ),
-          List(MavenRepository("public", "https://repo5.org/maven/", None))
+          List(MavenRepository("public", "https://repo5.org/maven/", None, Nil))
         )
       )
     )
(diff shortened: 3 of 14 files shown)


julienrf (commented):
There are a couple of minor details that could be discussed (e.g., the usage of `Option[List[Header]]` vs `List[Header]`, and the way headers are serialized in JSON — I would just write `["foo", "bar"]` instead of `{ "key": "foo", "value": "bar" }`), but overall it looks correct to me, and it would unblock all the people that use a gitlab package registry!

markvandertol:
Thanks for having a look. I replaced the `Option[List[Header]]` with `List[Header]`. That indeed simplifies the code.
I feel that encoding `{ "key": "foo", "value": "bar" }` as object instead of array is clearer, the same as the credentials object is encoded as `{ "user": "foo", "pass", "bar" }`. Also, because you could have more than one header. If you want I can change this.

julienrf:
May I kindly request a review from the maintainers?

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -126,15 +126,35 @@ object CoursierAlg {
...
+  private def toCoursierAuthentication(
+      credentials: Option[Credentials],
+      headers: List[Resolver.Header]
+  ): Option[Authentication] =
+    if (credentials.isEmpty && headers.isEmpty) {
+      None
+    } else {
+      Some(
+        new Authentication(
+          credentials.fold("")(_.user),
```
exoego, line 148: Q) Does `""` for user works ?

modules/core/src/test/scala/org/scalasteward/core/update/PruningAlgTest.scala
```diff
@@ -192,7 +193,7 @@ class PruningAlgTest extends FunSuite {
             DependencyInfo("org.scala-lang".g % "scala-library".a % "2.12.14", List("build.sbt")),
             DependencyInfo("org.scala-lang".g % "scala-library".a % "2.13.5", List("build.sbt"))
           ),
-          List(MavenRepository("public", "https://repo5.org/maven/", None))
+          List(MavenRepository("public", "https://repo5.org/maven/", None, Nil))
```
exoego, line 196: Some tests on this file requires updates too
```
==> X org.scalasteward.core.update.PruningAlgTest.needsAttention: group-specific frequency (monthly) vs repo frequency (asap)  0.003s scala.MatchError: Left(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(headers), DownField(MavenRepository), DownArray, DownField(resolvers), DownArray, DownField(dependencyInfos)))) (of class scala.util.Left)
    at org.scalasteward.core.update.PruningAlgTest.$anonfun$new$11(PruningAlgTest.scala:220)
==> X org.scalasteward.core.update.PruningAlgTest.needsAttention: group-specific frequency (asap) vs repo frequency (monthly)  0.001s scala.MatchError: Left(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(headers), DownField(MavenRepository), DownArray, DownField(resolvers), DownArray, DownField(dependencyInfos)))) (of class scala.util.Left)
    at org.scalasteward.core.update.PruningAlgTest.$anonfun$new$14(PruningAlgTest.scala:331)
==> X org.scalasteward.core.update.PruningAlgTest.needsAttention: dependency can have artifactId and version  0.0s scala.MatchError: Left(DecodingFailure(Attempt to decode value on failed cursor, List(DownField(headers), DownField(MavenRepository), DownArray, DownField(resolvers), DownArray, DownField(dependencyInfos)))) (of class scala.util.Left)
```

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
markvandertol: Yes, that what's Coursier also does internally when just specifying headers. See: https://github.com/coursier/coursier/blob/2cb552ea934a100f52ce79f94278304c90d808a4/modules/util/shared/src/main/scala/coursier/core/Authentication.scala#L88

modules/core/src/test/scala/org/scalasteward/core/update/PruningAlgTest.scala
markvandertol: I pushed a fixed version to the branch. My bad.

Thanks for having a look

cipriansofronia:
@markvandertol, could u also update the [gitlab docs](https://github.com/scala-steward-org/scala-steward/blob/main/docs/running.md#gitlab) on how we can use this new feature?

markvandertol:
> @markvandertol, could u also update the [gitlab docs](https://github.com/scala-steward-org/scala-steward/blob/main/docs/running.md#gitlab) on how we can use this new feature?
@cipriansofronia I've added a few lines explaining how to authenticate against a Gitlab repository. Does this clarify enough how it works? No specific code for Scala Steward is needed.

docs/running.md
```diff
@@ -260,6 +260,33 @@ echo "${SCALA_STEWARD_TOKEN}"
 7. add the `repos.md` file 
 8. (*optional*) create a new schedule to trigger the pipeline on a daily/weekly basis
 
+Scala Steward is compatible with Coursier authentication using headers. To authenticate 
+using the [Gitlab CI/CD job token](https://docs.gitlab.com/ee/ci/jobs/ci_job_token.html), while also supporting your own private token when performing
```
exoego, line 264: 👍

jchapuis:
Looking forward to this one being released, thanks a lot @markvandertol

cipriansofronia:
@fthomas can this PR get some ❤️ from you? 😇

markvandertol:
Is there anything preventing this PR from being merged? 🙂

exoego:
@fthomas Do you have a time to test this with your test Steward instance ??

modules/sbt-plugin/src/main/scala/org/scalasteward/sbt/plugin/StewardPlugin.scala
```diff
@@ -56,14 +56,22 @@ object StewardPlugin extends AutoPlugin {
           Try(Credentials.forHost(sbtCredentials, url.getHost)).toOption.flatten
             .map(c => Resolver.Credentials(c.userName, c.passwd))
 
+        def getHeaders(name: String): List[Resolver.Header] =
+          for {
+            (repositoryId, authentication) <-
+              csrConfiguration.value.authenticationByRepositoryId.toList
+            if repositoryId == name
+            (headerKey, headerValue) <- authentication.headers
+          } yield Resolver.Header(headerKey, headerValue)
```
987Nabil, line 65: I think you should not only take the headers from the `csrConfiguration` but also username and password.
Setting these via `csrConfiguration` is not wrong. And I would prefer `csrConfiguration` over `credentials` if present, since it is more specific/explicit.

Funny enough, I did not see this pr and coded it myself that way. Maybe you can have a [look](https://github.com/987Nabil/scala-steward/blob/import-csr-http-headers/modules/sbt-plugin/src/main/scala/org/scalasteward/sbt/plugin/StewardPlugin.scala) ;)

987Nabil: I forgot the @markvandertol :)

jchapuis:
👋  is there anything holding up the merge at this point?

kaffepanna:
> wave is there anything holding up the merge at this point?

I am wondering that to. This would really be a good thing to have when managing dependencies in private gilab repos

987Nabil:
> > wave is there anything holding up the merge at this point?
> 
> I am wondering that to. This would really be a good thing to have when managing dependencies in private gilab repos

It is not only gitlb. Also Google Cloud Storage or Artifactregistry or managing packages in AWS need this feature.

modules/sbt-plugin/src/main/scala/org/scalasteward/sbt/plugin/StewardPlugin.scala
exoego: I am going to merge as-is since markvandertol seems busy now.
@987Nabil You can submit a PR for that change 😉

exoego:
@markvandertol Sorry for waiting so long. Could you resolve the conflict? 🙇

markvandertol:
No worries. I resolved the merge conflict. 🙂

modules/sbt-plugin/src/main/scala/org/scalasteward/sbt/plugin/StewardPlugin.scala
markvandertol: Sorry for not replying to this earlier. It indeed looks more consistent, but I'm not entirely sure how large of a breaking change would be, e.g. for people who disabled _coursier_ in their projects. Retrieving the credentials old way, but looking for headers in `csrConfiguration` shouldn't be breaking, so I propose to do that in a different PR so the impact can better be assessed.

987Nabil: We can do this in another PR for sure. But I think, deactivating coursier and then relying on `csrConfiguration` where csr = coursier and the [description](https://github.com/sbt/sbt/blob/1.7.x/main/src/main/scala/sbt/Keys.scala#L436) in sbt also explicitly mentions coursier, seems to be odd.

michaltomanski:
Thanks Mark for this PR! It looks promising.

Any chance to have it merged some time soon?

fthomas:
> Thanks for having a look. I replaced the `Option[List[Header]]` with `List[Header]`. That indeed simplifies the code.

Having the fields optional allows to read in persisted `RepoCache`s of previous Scala Steward runs. If the header fields are non-optional, Scala Steward fails to read these files:
```
[info] 2022-07-14 14:48:58,983 INFO  ──────────── Steward scala-steward-org/scala-steward ────────────
[info] 2022-07-14 14:48:58,983 INFO  Check cache of scala-steward-org/scala-steward
[info] 2022-07-14 14:48:59,080 ERROR Failed to parse or decode JSON from /home/frank/data/code/scala-steward/core/workspace/store/repo_cache/v1/github/scala-steward-org/scala-steward/repo_cache.json
[info] io.circe.DecodingFailure$$anon$2: Attempt to decode value on failed cursor: DownField(headers),DownField(MavenRepository),DownArray,DownField(resolvers),DownArray,DownField(dependencyInfos)
```
Another alternative would be to bump the version `repoCacheStore` [here](https://github.com/scala-steward-org/scala-steward/blob/2cb32ef795f482a6854e3e3734e0e3c159283f2b/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala#L137-L138). But for this small change I would prefer making the fields optional.

Other than that, this PR looks good to me.

markvandertol:
Thanks @fthomas for having a look at the PR.

I fixed the issue by using a default value of `Nil` for the headers, and using deriveConfiguredCodec as shown below.
```scala
  @annotation.nowarn("cat=unused-locals")
  implicit val resolverCodec: Codec[Resolver] = {
    implicit val customConfig: Configuration = Configuration.default.withDefaults
    deriveConfiguredCodec
  }
```
I had to suppress the warning, as the implicit is only used by a macro. I could also have used the compiler flag `-Ywarn-macros:after`, but that would be inconsistent with other `@annotation.nowarn` usages.

exoego:
Sorry for the late response.

jchapuis:
Any change this can get shipped in a new release? 😃 🙏 thanks!

fthomas:
Just noticed that this change broke `StewardPlugin.scala` for old sbt versions. See #2847 for details.

PR #2536, opened 2022-02-23
Give advise to slow down update frequency

exoego:
Follow-up of #2515


exoego:
Merging since this changes only PR description and I believe it helps user to find new settings.

fthomas:
I'm a bit late but just wanted to say that I like this feature!

PR #2538, opened 2022-02-25
SBT plugin: Find credentials by realms if possible

NeQuissimus:
Some repository providers use the same host URL for multiple repositories.
Repositories are differentiated by sub-paths.
In the current implementation, Scala Steward uses the first set of credentials that fits the host name.
This leads to all other repositories under the same base path to have invalid credentials.

This PR attempts to solve this issue by matching realm names first and falling back onto the original logic if such a match cannot be made.

Example credentials file
```
credentials += Credentials("Private Repository: org/repo1", "pkgs.org.io", "token", "ABC")
credentials += Credentials("Private Repository: org/repo2, "pkgs.org.io", "token", "DEF")
```
Example build.sbt
```
resolvers += "Private repository: org/repo1" at "https://pkgs.org.io/basic/repo1/maven/"
resolvers += "Private repository: org/repo2" at "https://pkgs.org.io/basic/repo2/maven/"
```

Currently, the first credentials line will be used exclusively, leading to HTTP 401 errors for `repo2`.

With this PR, realm names will be matched first, assigning the correct credentials to `repo2` during `sbt stewardDependencies`

### DIFF
--- a/modules/sbt-plugin/src/main/scala/org/scalasteward/sbt/plugin/StewardPlugin.scala
+++ b/modules/sbt-plugin/src/main/scala/org/scalasteward/sbt/plugin/StewardPlugin.scala
@@ -52,17 +52,22 @@ object StewardPlugin extends AutoPlugin {
           )
         val dependencies = libraryDeps ++ scalafixDeps
 
-        def getCredentials(url: URL): Option[Resolver.Credentials] =
-          Try(Credentials.forHost(sbtCredentials, url.getHost)).toOption.flatten
+        def getCredentials(url: URL, name: String): Option[Resolver.Credentials] =
+          (for {
+            allDirect <- Try(Credentials.allDirect(sbtCredentials)).toOption
+            maybeRealmAndHost = allDirect.find(c => c.realm == name && c.host == url.getHost)
+            maybeRealm = allDirect.find(_.realm == name)
+            maybeHost = allDirect.find(_.host == url.getHost)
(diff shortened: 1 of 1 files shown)


fthomas (dismissed):
LGTM.

https://github.com/sbt/librarymanagement/blob/c571b466eed2094fa0ffb29409c849eab1c81560/ivy/src/main/scala/sbt/librarymanagement/ivy/Credentials.scala#L30 shows that `Credentials.forHost` is the same as we have now but without the realm matching.

fthomas:
I just noticed that the [sbt docs](https://www.scala-sbt.org/1.x/docs/Publishing.html#Credentials) say this about credentials matching:

> NOTE: Credentials matching is done using both: realm and host keys.

Maybe we should do this here too. Instead of looking for credentials that match only the realm or only the host, we only use credentials that match both.

fthomas:
I interpret that as both must match at the same time. How about trying to match realm and host both at the same time first, then trying only the realm, and only the host last:
```
for {
  allDirect <- Try(Credentials.allDirect(sbtCredentials)).toOption
  maybeRealmAndHost = allDirect.find(c => c.realm == name && c.host == url.getHost)
  maybeRealm = allDirect.find(_.realm == name)
  maybeHost = allDirect.find(_.host == url.getHost)
} yield maybeRealmAndHost.orElse(maybeRealm).orElse(maybeHost)
```
This should find the correct credentials in case realm or host is ambiguous and keep the lenience of your current proposal.

NeQuissimus:
I like that, I'll add the additional check.

fthomas (approved):

PR #2543, opened 2022-03-04
Fix ignored test assertion

htmldoug:
First assertion was ignored previously. https://github.com/typelevel/munit-cats-effect#getting-started suggests using a for comprehension.


exoego (approved):
Nice catch

htmldoug:
Thanks for the quick merge!

PR #2598, opened 2022-05-09
Add sttp 3.6.1 artifact migrations

hugo-vrijswijk:
In [sttp 3.6.1](https://github.com/softwaremill/sttp/releases/tag/v3.6.1) some artifacts are no longer published as they're included in their respective 'core' modules (`httpclient-backend-fs2` ➡ `fs2`). Example of the rename that needs to be done: https://github.com/stryker-mutator/stryker4s/pull/1154/files

This PR adds artifact migrations for the changed modules.

I haven't found a setting for what version the artifact is migrated, so I assume scala-steward uses the latest version when upgrading?


hugo-vrijswijk:
@fthomas could you have a look at this 🙏?

mzuehlke (approved):

PR #2612, opened 2022-05-22
Adding IteratorToVector rule for sbt-guardrail 0.72.0

blast-hardcheese:
Attempt to reflect a change in defaults for generated code to reduce upgrade burden


mzuehlke (approved):

blast-hardcheese:
Thank you!

Issue #2616, opened 2022-06-01
Failed to load artifact migration (artifact-migrations.v2.conf)


xerial:
Recently, scala-steward starts failing when reading artifact-migrations.v2.conf file (an example [GitHub Action run](https://github.com/xerial/scala-steward-repos/runs/6698874751?check_suite_focus=true)): 
```
  Error: java.lang.Throwable: Failed to load artifact migrations from https://raw.githubusercontent.com/scala-steward-org/scala-steward/master/modules/core/src/main/resources/artifact-migrations.v2.conf
  Error: 	at org.scalasteward.core.update.artifact.ArtifactMigrationsLoader$$anonfun$decodeMigrations$1.applyOrElse(ArtifactMigrationsLoader.scala:50)
  Error: 	at org.scalasteward.core.update.artifact.ArtifactMigrationsLoader$$anonfun$decodeMigrations$1.applyOrElse(ArtifactMigrationsLoader.scala:50)
  Error: 	at scala.PartialFunction$AndThen.applyOrElse(PartialFunction.scala:288)
  Error: 	at cats.ApplicativeError.$anonfun$recoverWith$1(ApplicativeError.scala:130)
  Error: 	at recoverWith$extension @ org.scalasteward.core.io.process$.$anonfun$slurp$3(process.scala:63)
  Error: 	at blocking @ org.scalasteward.core.io.FileAlg$$anon$1.ensureExists(FileAlg.scala:100)
  Error: 	at blocking @ org.scalasteward.core.io.FileAlg$$anon$1.ensureExists(FileAlg.scala:100)
  Error: 	at blocking @ org.scalasteward.core.io.FileAlg$$anon$1.ensureExists(FileAlg.scala:100)
  Error: 	at make @ org.http4s.okhttp.client.OkHttpBuilder$.defaultOkHttpClient(OkHttpBuilder.scala:202)
  Error: 	at make @ org.http4s.okhttp.client.OkHttpBuilder$.defaultOkHttpClient(OkHttpBuilder.scala:202)
  Error: 	at flatMap @ org.scalasteward.core.update.artifact.ArtifactMigrationsLoader.$anonfun$loadMigrations$2(ArtifactMigrationsLoader.scala:46)
  Error: 	at map @ org.scalasteward.core.update.artifact.ArtifactMigrationsLoader.$anonfun$loadMigrations$2(ArtifactMigrationsLoader.scala:46)
[...]

PR #2619, opened 2022-06-02
Check files frequently updated by community

exoego:
To prevent issues like #2616

Maybe it is better to have a dedicated sbt sub-project for this test, but I want to add this ASAP.
Let's consider such refactoring later.

### DIFF
--- a/modules/core/src/test/scala/org/scalasteward/core/BuiltinConfigFilesTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/BuiltinConfigFilesTest.scala
@@ -1,0 +1,76 @@
+package org.scalasteward.core
+
+import better.files._
+import cats.effect.unsafe.implicits.global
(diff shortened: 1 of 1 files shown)


fthomas:
I thought that the lines
https://github.com/scala-steward-org/scala-steward/blob/4eeb2894c0a31b5bec4270c2895a8b951994587d/modules/core/src/test/scala/org/scalasteward/core/mock/MockContext.scala#L24-L33
would also catch errors in these files. It is probably worth investigating why these didn't caught the issue mentioned above.

fthomas:
> It is probably worth investigating why these didn't caught the issue mentioned above.

It would have caught the issue if it would load the current config:
```diff
- ioFileAlg.readResource("artifact-migrations.conf").unsafeRunSync(),
+ ioFileAlg.readResource("artifact-migrations.v2.conf").unsafeRunSync(),
```

PR #2630, opened 2022-06-10
Add Nebula/Gradle support

jvican:


jvican:
Hi folks, I opened this PR by mistake 😅 This is a preliminary PR for Nebula support (a Gradle flavor that we have internally). I would like to make it available for stock Gradle at some point, so I'm going to battle test that it works for Nebula first and if it does I'll port it to stock gradle (which requires to parse the dependencies from a more difficult output format). Let me know if there's interest in it.

fthomas:
Support for Gradle sounds good to me!
IMO adding support for JVM build systems is uncontroversial since it is just another implementation of `BuildSystemAlg`.

PR #2635, opened 2022-06-15
Allow pre-release updates to specific dependencies

### LINKED ISSUE #2634: Allow pre-release updates to configured dependencies
It could be that one wants to update the specific artifact to a pre-released version ignoring same series limitation of the current implementation.

Proposed solution is to have it configured separately, with the list being empty by default (disallow all).

arixmkii:
Fixes #2634 

Added dedicated configuration, which alters pre-release version checks for specified dependencies.

Similar features has been mentioned there https://github.com/scala-steward-org/scala-steward/issues/1023#issuecomment-551942307

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -52,6 +52,12 @@ updates.pin  = [ { groupId = "com.example", artifactId="foo", version = "1.1." }
 # Defaults to empty `[]` which mean Scala Steward will not ignore dependencies.
 updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
+# The dependencies which match the given patterns are allowed to be updated to pre-release from stable.
+#
+# Each pattern must have `groupId`, and may have `artifactId` and `version`.
+# Defaults to empty `[]` which mean Scala Steward will ignore all stable to pre-release update options.
+updates.allowPreReleases  = [ { groupId = "com.example", artifactId="foo" } ]
+
 # If set, Scala Steward will only create or update `n` PRs each time it runs (see `pullRequests.frequency` above).
 # Useful if running frequently and/or CI build are costly
 # Default: None
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -53,6 +53,7 @@ updates.pin  = [ { groupId = "com.example", artifactId="foo", version = "1.1." }
 updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
 # The dependencies which match the given patterns are allowed to be updated to pre-release from stable.
+# This also implies, that it will be allowed for snapshot versions to be updated to snapshots of different series.
 #
 # Each pattern must have `groupId`, and may have `artifactId` and `version`.
 # Defaults to empty `[]` which mean Scala Steward will ignore all stable to pre-release update options.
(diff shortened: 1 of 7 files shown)


arixmkii:
I plan to update docs, but posting this early here to see if additional changes will be requested.

arixmkii:
I though it was, but adding more tests it didn't play nice with snapshots. Added support for snapshots and tests in a separate commit.

exoego (changes requested):
Thanks!
Code looks great to me.
Could you update document as well🙇
https://github.com/scala-steward-org/scala-steward/blob/main/docs/repo-specific-configuration.md

arixmkii:
Changes:
* Fixed compatibility with https://github.com/sbt/sbt-dynver versioning scheme for snapshots.
* Updated sample repo config with new feature documentation
* Added new config settings to all config settings test.

arixmkii:
I can squash commits and rebase if necessary. Kept commits separate to ease the review process.

docs/repo-specific-configuration.md
```diff
@@ -52,6 +52,12 @@ updates.pin  = [ { groupId = "com.example", artifactId="foo", version = "1.1." }
 # Defaults to empty `[]` which mean Scala Steward will not ignore dependencies.
 updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
+# The dependencies which match the given patterns are allowed to be updated to pre-release from stable.
+#
+# Each pattern must have `groupId`, and may have `artifactId` and `version`.
+# Defaults to empty `[]` which mean Scala Steward will ignore all stable to pre-release update options.
+updates.allowPreRelease  = [ { groupId = "com.example", artifactId="foo" } ]
```
armanbilge, line 59: Question: if I add this, do I get Ms and RCs or do I get snapshots? What if only want Ms and RCs?

arixmkii: For now both. Could you help me understand your requirements a bit better. Instead of boolean flag we would like to have pre-release modes like:

1. none - previous behavior
2. only Ms and RCs
3. only snapshots
4. all kind of pre-releases (2 + 3)

Is this correct? Or should it be 

1. none - previous behavior
2. only Ms and RCs
3. Ms, RCs and snapshots

It will end up having 2 lists in seconds anyway:
* updates.allowPreRelease
* updates.allowSnapshots

The question is "Should allowSnapshots include allowPreRelease?"

armanbilge: Yes, I think `allowSnapshots` should imply `allowPreRelease`. Similarly, it should allow updates to ordinary stable releases! :) For example, if a pre-release or snapshot is succeeded by a stable release, it should update to that (at least until the next snapshot/pre-release is available).

exoego: @arixmkii 
Could you add a test case for below?

> if a pre-release or snapshot is succeeded by a stable release, it should update to that

fthomas: I agree with what @armanbilge said in https://github.com/scala-steward-org/scala-steward/pull/2635#discussion_r900295436. But I would prefer implementing `allowSnapshots` in a later PR so that this PR does not become too big. I think there is no harm if `allowPreRelease` implies `allowSnapshots` for now.

Btw, if we pluralize snapshots in `allowSnapshots` shouldn't we also pluralize pre-releases in `allowPreReleases`?

arixmkii: Implemented requested changes:
* used plural version of preReleases, where appropriate
* added more tests and grouped them logically
* added filter to prohibit updates from regular pre-release to snapshot

Rebased to latest main and squashed commits.

exoego (approved):
LGTM 💯 
Sorry for late response.

PR #2523, opened 2022-02-12
Add support for PR labels

ybasket:
Takes a stab at #1471 to make the PR/MR labels appear not only in the textual description, but also using the git hoster's native labelling feature. Implemented for GitHub and GitLab, Bitbucket doesn't have a native label feature, only a plugin (for which support could be added later, now the implementation just logs a warning). It's opt-in for now as it's a change in behaviour not every user might appreciate, but could be promoted to a default later on.

Note: GitHub creates the labels with a standard grey colour (first screenshot), but the user can update that (see second).

PS: Scala Steward is not only a great help, but also a very nice codebase to work with, thank you!

Closes #1471 

Screenshot 2022-02-12 at 11 50 44
Screenshot 2022-02-12 at 12 04 08
Screenshot 2022-02-12 at 12 25 18

PR #2636, opened 2022-06-16
Add regex filtering for PR labels

ybasket:
Filter labels Scala Steward adds to PRs it creates by a regex to avoid visual clutter or conflicts with existing label conventions. Follows up on discussion that happened in #2523.


docs/repo-specific-configuration.md
```diff
@@ -32,6 +32,11 @@ You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward upd
 #pullRequests.frequency = "0 0 ? * 3" # every thursday on midnight
 pullRequests.frequency = "7 days"
 
+# pullRequests.labelRegex allows to control which labels are added to PRs 
+# via a regex check each label is checked against.
+# Defaults to no regex (all labels are added) which is equivalent to ".*".
+pullRequests.labelRegex = "(.*semver.*)|(commit-count:n:.*)"
```
exoego, line 38: I think it would be better naming if the name indicate its behavior or use case.
Could you consider different nam like `pullRequests.excludeMatchedLabels` or such 🙇

ybasket: 👍 It's inclusive, not exclusive though, so I went for `pullRequests.includeMatchedLabels`

exoego (approved):
Sorry for late response.
LGTM 💯

PR #2639, opened 2022-06-22
Tag latest docker image with latest-release

### LINKED ISSUE #2613: suggestion - setting "latest" Docker tag for stable releases only
I've noticed that every time a `-SNAPSHOT` gets pushed to DockerHub, it is tagged as the latest version. I wonder if it would make more sense to only set the `latest` tag when it is a stable release and not a snapshot.

bcarter97:
Closes #2613 

Tag the latest stable docker image with `latest-release`, see https://www.scala-sbt.org/sbt-native-packager/formats/docker.html#publishing-settings

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -325,6 +325,9 @@ lazy val dockerSettings = Def.settings(
   },
   Docker / packageName := s"fthomas/${name.value}",
   dockerUpdateLatest := true,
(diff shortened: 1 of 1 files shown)


fthomas (approved):
Thank you @bcarter97!

PR #2645, opened 2022-06-25
Let Scala Steward send upgrades to Scala.js 1.10.1.

sjrd:
Not sure this config file is actually used by any running Steward at the moment, but I'm still following the protocol. ;)


exoego (approved):

fthomas:
> Not sure this config file is actually used by any running Steward at the moment, but I'm still following the protocol. ;)

Since I shut down my public instance there are now more public instances that use this file and I guess that most of the private instances use this file too (since this is the default). So IMO it still makes sense to keep this file up-to-date.

PR #2665, opened 2022-07-11
Advertize VirtusLab's public Scala Steward instance

fthomas:
This points users to https://github.com/VirtusLab/scala-steward-repos
instead of the now archived https://github.com/scala-steward-org/repos.

Many thanks again to VirtusLab for hosting the public instance!

/cc @romanowski @Kordyjan @pikinier20

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,12 +18,10 @@ See also the announcement blog post:
 
 ## Quick start guide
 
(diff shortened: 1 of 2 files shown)


Kordyjan:
If you can add that information, we will be really grateful.

exoego (approved):

PR #2668, opened 2022-07-13
Scalafix rewrites via dependency when target="build"

russellremple:
All our Scalafix migrations are in private repos. By using the "dependency:" prefix in rewrite rule definitions, Scala Steward leverages Coursier to resolve these from our private artifactory. But this only seems to work when the `target="sources"` (default). 

I'm attempting to create a new rewrite for SBT files by using `target="build"` and I am unable to fetch these rules from our artifactory. It appears the issue is that, where `"sources"` uses the Scalafix SBT plugin, `"build"` uses the Scalafix CLI, and the CLI does not support the "dependency:" prefix for rules. The [Scalafix docs](https://scalacenter.github.io/scalafix/docs/developers/tutorial.html#publish-the-rule-to-maven-central) indicate that "Users of the Scalafix command-line interface can use the `--tool-classpath` flag" to use published artifacts.

If there is a better way to achieve this through configuration, environment variables, etc., that would be amazing. But I haven't been able to work out how to do that, and without it, I'm thinking a change like this might solve the problem. Looking for feedback and guidance on this approach. Basically it would keep existing `target="build"` rules using `https`, `github`, `file`, etc. unchanged, but would reinterpret rules specified in the form `dependency:ruleName@repo::artifact:version` to be `ruleName --tool-classpath $(cs fetch repo::artifact:version -p)`, consistent with the Scalafix documentation linked above.

If I'm on the right track here, I'd be happy to update documentation and write a couple tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,7 @@ Consider creating PR to add your company to the list and join the community.
 * [Teads](https://medium.com/teads-engineering)
 * [Tegonal GmbH](https://tegonal.com)
 * [Tupl](https://www.tupl.com)
+* [VirtusLab](https://virtuslab.com/)
 * [wehkamp](https://www.wehkamp.nl/)
 * [Zalando](https://en.zalando.de/)
 
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/scalafix/ScalafixCli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/scalafix/ScalafixCli.scala
@@ -29,7 +29,14 @@ final class ScalafixCli[F[_]](implicit
     F: Monad[F]
 ) {
   def runMigration(workingDir: File, files: Nel[File], migration: ScalafixMigration): F[Unit] = {
-    val rules = migration.rewriteRules.map("--rules=" + _)
+    val rules = migration.rewriteRules.map { r =>
+      r.split("[:@]").toList match {
+        case "dependency" :: ruleName :: depParts =>
+          s"--rules=$ruleName --tool-classpath $$(cs fetch ${depParts.mkString(":")} -p)"
+        case _ =>
+          s"--rules=$r"
+      }
+    }
     val cmd = scalafixBinary :: rules ::: files.map(_.pathAsString)
     processAlg.exec(cmd, workingDir).void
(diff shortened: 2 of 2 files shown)


fthomas:
The approach look good to me or at least I don't have a better idea how to use published rules with the Scalafix CLI. But I don't think that the proposed solution would work:
```scala
          s"--rules=$ruleName --tool-classpath $$(cs fetch ${depParts.mkString(":")} -p)"
```
Scala Steward will pass this argument verbatim to `java.lang.ProcessBuilder` and not a shell so that `$(cs fetch ... -p)` will not be expanded unless Scalafix will do the expansion (which I doubt). A solution that could work in this case is to add a new `toolClasspathCommand` (or `toolClasspathCommands` if Scalafix allows multiple `--tool-classpaths` arguments) to [`ScalafixMigration`](https://github.com/scala-steward-org/scala-steward/blob/2cb32ef795f482a6854e3e3734e0e3c159283f2b/modules/core/src/main/scala/org/scalasteward/core/edit/scalafix/ScalafixMigration.scala) which is(are) executed by Scala Steward and whose output(s) is(are) then used as value(s) of `--tool-classpath` argument(s).

WDYT?

fthomas:
> A solution that could work in this case is to add a new `toolClasspathCommand`

We could of course keep the proposed syntax and let Scala Steward execute the `cs fetch ...` command but adding a new field to `ScalafixMigration` is IMO more general, easier to implement and easier to maintain.

russellremple:
Thanks for the thoughtful input. You are definitely correct, that this will not work as it is. We tried simulating the change locally, and the `$(cs fetch ... -p)` part does not expanded. I've tested the Scalafix CLI locally and it does syntactically allow for multiple `--tool-classpath` arguments s on the same line (e.g., `--rules=a --tool-classpath pathsForA- -rules=b --tool-classpath pathsForB`), but I don't know if it actually processes them all. I really like the idea of a new `toolClasspathCommands`. I'll struggle with that for a bit and update here if/when I get stuck. (Putting this in draft for now.) Thanks again!

russellremple:
Not sure when I'll get a chance to get back to this.

PR #2671, opened 2022-07-15
Add post-update hook for OrganizeImports

fthomas:
https://github.com/VirtusLab/akka-serialization-helper/issues/199#issuecomment-1183024264


PawelLipski:
+1... only seen now

Issue #2659, opened 2022-07-08
Support for Azure Repos


vidalgp:
Helloo,
I was wondering if there is any intention to support azure in the future (particularly azure DevOps git repos).
I tried [Renovate](https://docs.renovatebot.com/setup-azure-devops/), but it's nothing compared to what scala-steward accomplishes in a Scala project.

Cheers

exoego:
PR welcome

robinraju:
@exoego @vidalgp I created a PR to add support for Azure repos over the weekend. I tested it on my personal account, and it works as expected.

fthomas:
Fixed by #2837.

PR #2677, opened 2022-07-24
Add support for Azure Repos

robinraju:
This PR adds support for Azure Repos as requested on #2659 

Azure APi Doce: [here](https://docs.microsoft.com/en-us/rest/api/azure/devops/git/?view=azure-devops-rest-7.1)

WIth this feature, scala-steward can be run on Azure Repos using docker as follows

1. Create and add an access token to `pass.sh` file:  [See how](https://docs.microsoft.com/en-us/azure/devops/organizations/accounts/use-personal-access-tokens-to-authenticate?view=azure-devops&tabs=Windows)

```bash
echo "#!/bin/sh"  >> pass.sh
echo "echo '$AZURE_REPO_ACCESS_TOKEN'" >> pass.sh
```

2. Run steward
```bash
docker run -v $PWD:/opt/scala-steward \
    -it fthomas/scala-steward:latest \
    -DLOG_LEVEL=TRACE \
    --do-not-fork \
    --workspace "/opt/scala-steward/workspace" \
    --repos-file "/opt/scala-steward/repos.md" \
    --git-author-email "email@mycompany.com" \
    --vcs-type "azure-repos" \
    --vcs-api-host "https://dev.azure.com" \
    --vcs-login "email@mycompany.com" \
    --azure-repos-organization "mycompany" \
    --git-ask-pass "/opt/scala-steward/pass.sh"
```

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -238,6 +238,15 @@ object Cli {
   private val gitHubApp: Opts[Option[GitHubApp]] =
     (githubAppId, githubAppKeyFile).mapN(GitHubApp.apply).orNone
 
+  private val azureReposOrganization: Opts[Option[String]] =
+    option[String](
+      "azure-repos-organization",
+      "The Azure organization (required when vcs type is azure-repos)"
+    ).orNone
+
+  private val azureReposConfig: Opts[AzureReposConfig] =
+    azureReposOrganization.map(AzureReposConfig.apply)
+
   private val refreshBackoffPeriod: Opts[FiniteDuration] = {
     val default = 0.days
     val help = "Period of time a failed build won't be triggered again" +
@@ -270,6 +279,7 @@ object Cli {
     cacheTtl,
     bitbucketServerCfg,
     gitLabCfg,
+    azureReposConfig,
     gitHubApp,
     urlCheckerTestUrl,
     defaultMavenRepo,
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ Thanks goes to these wonderful people for contributing to Scala Steward:
 * [Renato Cavalcanti](https://github.com/renatocaval)
 * [Rikito Taniguchi](https://github.com/tanishiking)
 * [Robert Stoll](https://github.com/robstoll)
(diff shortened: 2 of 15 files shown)


modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
```diff
@@ -238,6 +238,15 @@ object Cli {
   private val gitHubApp: Opts[Option[GitHubApp]] =
     (githubAppId, githubAppKeyFile).mapN(GitHubApp.apply).orNone
 
+  private val azureReposOrganization: Opts[Option[String]] =
+    option[String](
+      "azure-repos-organization",
+      "The Azure organization (required when vcs type is azure-repos)"
+    ).orNone
```
exoego, line 245: I think it is better to validate `azure-repos-organization` so that user can understand what is missing.
Could you try to implement validation?
https://ben.kirw.in/decline/usage.html#transforming-and-validating

robinraju: @exoego Thanks for taking a look. I'll implement the validation.

sapka12: @robinraju tried your changes on azure and they work. thanks for it.

I have this proposal for the validation
```
private val azureReposOrganization: Opts[Option[String]] =
    option[String](
      "azure-repos-organization",
      "The Azure organization (required when vcs type is azure-repos)"
    ).validate("required when vcs type is azure-repos")(organization =>
      if (vcsTypeArgument.toString() === "azure-repos") organization.nonEmpty else true
    ).orNone
```

@exoego is this the validation that is missing to complete this PR?

exoego: @sapka12 👍 
You can create a new PR to complete this PR if you need this feature.

sapka12: https://github.com/scala-steward-org/scala-steward/pull/2837

sapka12: @exoego 
PR: https://github.com/scala-steward-org/scala-steward/pull/2837

Issue #2692, opened 2022-08-11
Issues parsing mill modules headers


daddykotex:
Hi, thanks for the project.

We're using the scala-steward on the bleeding edge. Right now, we're seeing issues with mill repositories where parsing `headers` from the json produced by Mill plugin. The JSON looked like this:

```json
{
  "modules": [
    {
      "name": "dslib-traits",
      "repositories": [
        {
          "url": "https://repo1.maven.org/maven2",
          "type": "maven",
          "auth": null
        }
      ],
      "dependencies": [
        {
          "groupId": "software.amazon.smithy",
          "artifactId": {
            "name": "smithy-model",
            "maybeCrossName": null
          },
          "version": "1.21.0"
        }
      ]
    }
  ]
}
```

Now it looks more like this:

```json
{
  "modules": [
    {
      "name": "dslib-traits",
      "repositories": [
        {
          "url": "https://repo1.maven.org/maven2",
          "type": "maven",
          "auth": null,
          "headers": [
            [
              
            ]
          ]
        }
      ],
      "dependencies": [
        {
          "groupId": "software.amazon.smithy",
          "artifactId": {
            "name": "smithy-model",
            "maybeCrossName": null
          },
          "version": "1.21.0"
        }
      ]
    },
  ]
}
```

A new field is introduced `headers`. The code makes it optional, its type being: `Option[List[Header]]`, but looking at the serialization, it looks like the option is rendered as a list. It leaves us with a `Array` of `Array` and this is not what the parser expects.

IMO, the fix is to render null if `authentication` is undefined, or to render a `List[Header]` if `authentication` is defined.

I'm preparing a patch for that.

After the patch, I expect the json to look like:
```
      "repositories": [
        {
          "url": "https://repo1.maven.org/maven2",
          "type": "maven",
          "auth": null,
          "headers": null // or "headers": [...]
        }
      ],
```

exoego:
This was overlooked in https://github.com/scala-steward-org/scala-steward/pull/2533
I think it was fixed in https://github.com/scala-steward-org/scala-steward/pull/2693

exoego:
I believe the SBT plugin does not have a similar issue.
Because it print `[]` if `headers` is empty.
https://github.com/scala-steward-org/scala-steward/blob/a2954c3d09f9ed01aaed377a0674aba5a60f90e5/modules/sbt-plugin/src/main/scala/org/scalasteward/sbt/plugin/StewardPlugin.scala#L203
[...]

PR #2693, opened 2022-08-11
Fix mill `headers` serialization

daddykotex:
Address: #2692


daddykotex:
ping @exoego since they're the maintainers that reviewed the original changes, but really, anyone can review this

exoego (approved):
Thanks for looking at this.

PR #2697, opened 2022-08-17
Add sbt-typelevel-github-actions to sbtGitHubWorkflowGenerateModules

rossabaker:
If the `sbt-typelevel-github-actions` plugin is used directly, we want pull requests to run `githubWorkflowGenerate`.


rossabaker:
Does this work if a plugin that Scala Steward is updating transitively adds a dependency on something in `sbtGitHubWorkflowGenerateModules`, or must it be directly mentioned?

I have a private plugin that is introducing sbt-typelevel-github-actions, and I want to make sure it generates the workflow before sending the entire organization a PR that fails the `githubWorkflowCheck` step it introduces.  Thanks!

rossabaker:
I think I want to configure `postUpdateHooks` in our private installation for my use case, but that this is still a useful default for the public.

fthomas:
Scala Steward does not know anything about transitive library or plugin dependencies. So `githubWorkflowGenerate` is only executed if a dependency in `sbtGitHubWorkflowGenerateModules` is updated or if a dependency in `conditionalSbtGitHubWorkflowGenerateModules` is updated and the project depends on one of the plugins in `sbtGitHubWorkflowGenerateModules`.

> I think I want to configure postUpdateHooks in our private installation for my use case

:+1:

fthomas (approved):

PR #2701, opened 2022-08-29
feat: add ability to detect Mill plugins

### LINKED ISSUE #2694: [feature request] Support for Mill plugins
From looking at the code here:

https://github.com/scala-steward-org/scala-steward/blob/a2954c3d09f9ed01aaed377a0674aba5a60f90e5/modules/mill-plugin/src/main/scala/org/scalasteward/mill/plugin/StewardPlugin.scala#L44-L45

I see that Steward only considers dependencies in modules, and also the Mill version itself from [here](https://github.com/scala-steward-org/scala-steward/blob/a2954c3d09f9ed01aaed377a0674aba5a60f90e5/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala#L87-L88). However, this misses out on the Mill plugins that are included in the actual build itself like this:

```
import $ivy.`com.goyeau::mill-scalafix::0.2.10`
import $ivy.`de.tototec::de.tobiasroeser.mill.integrationtest::0.6.1`
import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version::0.2.0`
...
```

There really isn't a way to pull this information from Mill as far as I'm aware, so it'd need to manually be put together. It'd also have to consider the Mill version because take `com.goyeau::mill-scalafix::0.2.10` for example. The artifact is actually `mill-scalafix_mill0.10_2.13` to support the current `0.10.x` series of Mill.

One potential way to grab these would be to just regex them since they're almost always in this format at the to of the file. You could do this after grabbing the Mill version, and then put the artifact name together. This would then allow for PRs to also be sent in for these, which are currently missed.

ckipp01:
This pr adds in the ability to detect Mill plugins that are defined in
the following way, which is new as of 0.10.x of Mill.

```
import $ivy.`com.goyeau::mill-scalafix::0.2.10`
```

The artifact for this will end up being:

```
mill-scalafix_mill0.10_2.13
```
There is no API currently for Mill to give you the plugins being used in
the build file, so instead of this logic being in the actual Mill plugin
we parse the build file.

Closes #2694

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/parser.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/parser.scala
@@ -19,6 +19,11 @@ package org.scalasteward.core.buildtool.mill
 import cats.syntax.all._
 import io.circe.{Decoder, DecodingFailure}
 import org.scalasteward.core.data.{Dependency, Resolver, Version}
+import cats.parse.Parser
+import cats.parse.Rfc5234.sp
+import org.scalasteward.core.data.GroupId
+import org.scalasteward.core.data.ArtifactId
+import scala.util.Try
 
 object parser {
   sealed trait ParseError extends RuntimeException {
@@ -43,6 +48,67 @@ object parser {
   def parseMillVersion(s: String): Option[Version] =
     Option(s.trim).filter(_.nonEmpty).map(Version.apply)
 
+  /**
+    * Used to correctly format the Mill plugin artifacts will when included look like:
+    * - import $ivy.`com.goyeau::mill-scalafix::0.2.10`
+    * However for the actual artifact if the user is on 0.10.x will look like:
+    * - mill-scalafix_mill0.10_.2.13
+    *
+    * @param artifactName name of the artifact parsed from the build file
+    * @param millVerion the current Mill version being used
+    * @return the newly put together ArtifactId
+    */
+  private def millPluginArtifact(artifactName: String, millVersion: Version): ArtifactId = {
+    def format(major: String, minor: String) = s"${major}.${minor}_2.13"
+    val millSuffix = millVersion.value.split('.') match {
+      // Basically for right now we only accept "0.9.12", which is when this syntax for Mill plugins
+      // was introduced, and all other pre v1 versions. Once it's for sure verified that v1 will also
+      // follow this pattern we can include that.
+      case Array(major, minor, patch) if major == "0" && minor == "9" && patch == "12" =>
+        format(major, minor)
+      case Array(major, minor, _) if major == "0" && Try(minor.toInt).map(_ > 9).getOrElse(false) =>
+        format(major, minor)
+      case _ => ""
+    }
+
+    ArtifactId(artifactName, Some(s"${artifactName}_mill${millSuffix}"))
+  }
+
+  def parseMillPluginDeps(s: String, millVersion: Version): List[Dependency] = {
+
+    val importParser = Parser.string("import")
+    val ivyParser = Parser.string("$ivy")
+    val backtickParser = Parser.char('`')
+    val doubleColonParser = Parser.string("::")
+    val dotParser = Parser.char('.')
+    val grabUntilColonParser = Parser.until(doubleColonParser)
+    val grabntilBacktickParser = Parser.until(backtickParser)
+
+    val parser =
+      (importParser ~
+        sp.rep ~
+        ivyParser ~
+        dotParser ~
+        backtickParser) *>
+        (grabUntilColonParser.string <*
+          doubleColonParser) ~
+        grabUntilColonParser.string ~
+        (doubleColonParser *>
+          grabntilBacktickParser.string <*
+          backtickParser)
+
+    val pluginDependencies = s
+      .split("\n")
+      .toList
+      .map { line =>
+        parser.parse(line)
+      }
+      .collect { case Right((_, ((org, artifact), version))) =>
+        Dependency(GroupId(org), millPluginArtifact(artifact, millVersion), Version(version))
+      }
+
+    pluginDependencies
+  }
 }
 
 case class Modules(modules: List[MillModule])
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/parser.scala
```diff
@@ -43,6 +47,62 @@ object parser {
...
+    * - mill-scalafix_mill0.10_.2.13
+    *
+    * @param artifactName name of the artifact parsed from the build file
+    * @param millVerion the current Mill version being used
+    * @return the newly put together ArtifactId
+    */
+  private def millPluginArtifact(artifactName: String, millVersion: Version): ArtifactId = {
+    val millSuffix = millVersion.value.split('.') match {
+      // Basically for right now we only accept "0.10.x" which is when this syntax for Mill plugins
+      // was introduced. As the versions increase we'll need to readdress this.
```
lolgab, line 63: This was backported to Mill 0.9.12 and will be used most likely in future 0.x versions of Mill.
Why don't you accept any major and minor instead?

ckipp01: Mainly because if I accept any version, there is no guarantee that it will work. I've seen a lot of 0.9.x repos that were using the old format before the whole `_mill*` thing was introduced, so this wouldn't work on those. Same with 1.x stuff, there is no guarantee that this will be the same patter as mentioned [here](https://github.com/com-lihaoyi/mill/pull/1994#discussion_r951594263). So I guess I was just being a bit conservative here. I can switch it, but then it should probably only match on `0.9.12` if it's a `0.9.x` and only consider minors until we're positive this will be the format for `1.x`.

wdyt?

ckipp01: I went ahead and did that, I included 0.9.12, expanded it to any minor above that, but still don't consider 1.x until we are 100% sure it'll be this way.

lolgab: You should only check the dependencies that have the second `::`.
These are accepted only in versions of Mill that support the binary suffix so it is safe in older versions since mill plugins were imported with:
```scala
import $ivy.`group_id::artifact_id:version`
```
And after the mill binary syntax was introduced you can have:
```scala
import $ivy.`group_id::artifact_id::version`
```

Note also that even in current mill, some plugins didn't update yet to the new `_mill0.10` binary suffix and need to be imported without the second `::`

lolgab: About `1.x` we can't see the future, but most likely this will not change.
I don't see other ways `group_id::artifact::version` could be used if not for mill binary version.
If you make it general, it will work out of the box in case Mill doesn't change.
If you fix it to `0.x` you have 100% probability of future maintenance to implement it for 1.x.

ckipp01: > You should only check the dependencies that have the second ::.

Yes, this was intended actually but I had leftover stuff in the parser things. I've updated this now to look for the `::` instead of just `:` so it will only parse those that use the double `::`.

ckipp01: Actually re-reading this, sorry I don't understand what you mean by:

> Note also that even in current mill, some plugins didn't update yet to the new _mill0.10 binary suffix and need to be imported without the second ::

If they haven't updated to this format, then it'd be `:`, not `::` right?

lolgab: > Actually re-reading this, sorry I don't understand what you mean by:
> 
> > Note also that even in current mill, some plugins didn't update yet to the new _mill0.10 binary suffix and need to be imported without the second ::
> 
> If they haven't updated to this format, then it'd be `:`, not `::` right?

Right!

```scala
import $ivy.`org.example::plugin:1.2.3`
```
fetches `org.example:plugin_2.13:1.2.3`

```scala
import $ivy.`org.example::plugin::1.2.3`
```
fetches `org.example:plugin_mill0.10_2.13:1.2.3`

ckipp01:
> I've made an ammonite script to do this and found that the plugins could be declared in many ways like:
> 
> ```scala
> // Separate lines:
> import $ivy.`io.github.davidgregory084::mill-tpolecat::0.3.1`
> import $ivy.`de.tototec::de.tobiasroeser.mill.vcs.version::0.1.3`
> // Ivy import and plugin import in the same line
> import $ivy.`com.goyeau::mill-scalafix::0.2.9`, com.goyeau.mill.scalafix.ScalafixModule
> // Multiline import
> import $ivy.{
>   `com.lihaoyi::requests:0.2.0 compat`,
>   `com.lihaoyi::ujson:0.7.5 compat`
> }
> // Multiple imports in the same line:
> import $ivy.{`com.lihaoyi::requests:0.2.0 compat`, `com.lihaoyi::ujson:0.7.5 compat`}
> ```
> 
> You can check my implementation at https://gist.github.com/carlosedp/c3ea2814ac5392051a562e95c1298239

Thanks for including the extra info @carlosedp. It's honestly a shame that they can be included in so many ways. My goal wasn't to capture every way, but capture what I thought was the most common way I see them being used in modern Mill builds. Hopefully just having this in here will be a good base to start off, and if we find that users are commonly including plugins in another way, then we can probably address it then.

carlosedp:
Yea, I solved this by filtering the plugins using some regexes. First get something that is between backticks, then making sure this string contains a pattern with three fields separated by any number of ":".
Dunno if it's totally correct but worked fine and didn't get false-positives on literals.

lefou:
I haven't looked into this PR, but I just wanted to mention, that we probably should solve this issue rather in the Mill plugin which currently extracts the other updateable dependencies. That way, we could better use Mill internals (or even future API) to get access to these details.

Also, I have some vague plans to support alternative ways of declaring plugins and other dependencies in Mill (keyword: using directives), and handling these in the plugin seems to be more easy and correct.

ckipp01:
> I haven't looked into this PR, but I just wanted to mention, that we probably should solve this issue rather in the Mill plugin which currently extracts the other updateable dependencies. That way, we could better use Mill internals (or even future API) to get access to these details.

Agreed, but for now this was sort of a way to at least get some automated prs for plugins. Once an API is exposed I'd be happy to update this to that.

exoego (approved):
LGTM thanks 🏆

PR #2707, opened 2022-09-06
Ignore common prefix weight for snapshots ordering

arixmkii:
This is a bug fix to match the documented behavior in #2635 

> This also implies, that it will be allowed for snapshot versions to be updated to snapshots of different series.

For this to work reliably we need to eliminate weight of common prefix in allowPrereleases mode upgrading from Snapshot to Snapshot.

This is crucial, because in many projects (which publishes only snapshots for some reasons like continuous builds) there is no point to merge snapshots one by one sequentially.

Tests for these cases added.


exoego (approved):
LGTM

PR #2709, opened 2022-09-08
chore: Exclude Akka artifacts due to change of license

tgodzik:
Akka starting from version 2.7.0 will use the BSL license, which will require from large companies to pay for license. Due to that any library that uses Akka and is included in a project would also need to pay for licenses.

To avoid unexpected costs for companies, let's add Akka 2.7.0 as ignored for Scala Steward, so that users will need to update it manually and consiously accept the new license.

Not sure how to properly allow for the updates if the user wants it, but if it proves to be difficult we can add some additional settings. I was thinking about maybe `allowedLicenses` that could get overriden in a local repo.

Let me know what everyone thinks about this change.

### DIFF
--- a/modules/core/src/main/resources/default.scala-steward.conf
+++ b/modules/core/src/main/resources/default.scala-steward.conf
@@ -134,5 +134,14 @@ updates.ignore = [
   { groupId = "org.http4s", version = "0.19.0" },
 
   // https://github.com/scala-js/scala-js/issues/3865
-  { groupId = "org.scala-js", version = "0.6.30" }
+  { groupId = "org.scala-js", version = "0.6.30" },
+  
+  // https://www.lightbend.com/blog/why-we-are-changing-the-license-for-akka
+  // https://github.com/scala-steward-org/scala-steward/issues/2708
+  { groupId = "com.typesafe.akka"},
+  { groupId = "com.lightbend.akka" },
+  { groupId = "com.lightbend.akka.grpc" },
+  { groupId = "com.lightbend.akka.management" },
+  { groupId = "com.lightbend.akka.discovery" },
+  
 ]


modules/core/src/main/resources/default.scala-steward.conf
```diff
@@ -134,5 +134,8 @@ updates.ignore = [
   { groupId = "org.http4s", version = "0.19.0" },
 
   // https://github.com/scala-js/scala-js/issues/3865
-  { groupId = "org.scala-js", version = "0.6.30" }
+  { groupId = "org.scala-js", version = "0.6.30" },
+  
+  // https://github.com/scala-steward-org/scala-steward/issues/2708
+  { groupId = "com.typesafe.akka", version = { prefix = "2.7." } }
```
fthomas, line 140: FYI: @ckipp01 ignored two more groupIds in https://github.com/akka/akka/pull/31561#issuecomment-1239052679. Maybe it is safer to add them too.

fthomas: Btw, if Akka is ignored like the above, Scala Steward will prevent updates to 2.7 but will propose updates to 2.8 or later when they are available.

tgodzik: Looks like the other ones have a lot of different versions :thinking: so we would need to ignore them per artifact or ignore them altogether for now and work on license checking. What do you think?

tgodzik: > Btw, if Akka is ignored like above, Scala Steward will prevent updates to 2.7 but will propose updates to 2.8 or later when they are available.

ach, that's right. So maybe ignore all for now?

tgodzik: I changed it to all the groupIds I found. I will try to work on the licenses next week.

tgodzik:
> > I was thinking about maybe `allowedLicenses` that could get overriden in a local repo.
> 
> This would be awesome, not just for the Akka case, but also for libraries that change license to GPL/LGPL, for example.

I think this would be a better long term solution, though I am not currently that familiar with the codebase.

modules/core/src/main/resources/default.scala-steward.conf
armanbilge: Instead of ignoring, why not pinning to `2.6`? This is what we do in Typelevel for Scala 3 updates.
- https://github.com/typelevel/steward/pull/10

tgodzik: Wasn't sure what pinning would entail. Would it only update it up until the said version?

Also I would need to go through all the libraries since they have different versions where they start being BSL :/

Probably best approach would be to ignore it for now and then add the configuration for licenses.

tgodzik: @fthomas what do you think? Is it ok to ignore for now and fix the license later? Alternatively, we can go trough the akka libs one by one and pin all of them.

We could also only change the config for the public instance of Scala Steward, however, we would need to make sure in that case that any changes to the default config are upstreamed.

fthomas: > @fthomas what do you think? Is it ok to ignore for now and fix the license later?

I'm OK with that.

tgodzik: Awesome! I will make sure that this is only temporary and work on the license as soon as possible.

PR #2715, opened 2022-09-25
prepare Scala 3

xuwei-k:


modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -17,7 +17,7 @@
 package org.scalasteward.core.coursier
 
 import cats.effect._
-import cats.implicits._
+import cats.syntax.all._
```
xuwei-k, line 20: - https://github.com/typelevel/cats/issues/4138

modules/core/src/main/scala/org/scalasteward/core/io/FileAlg.scala
```diff
@@ -23,7 +23,7 @@ import cats.{ApplicativeError, MonadThrow, Traverse}
 import fs2.Stream
 import org.apache.commons.io.FileUtils
 import org.http4s.Uri
-import org.http4s.implicits.http4sLiteralsSyntax
+import org.http4s.implicits._
```
xuwei-k, line 26: - https://github.com/http4s/http4s/blob/6b0a05eb7f17fc923dafa46507639893cbe6fcb7/core/shared/src/main/scala-2/org/http4s/syntax/LiteralsSyntax.scala#L21
- https://github.com/http4s/http4s/blob/6b0a05eb7f17fc923dafa46507639893cbe6fcb7/core/shared/src/main/scala-3/org/http4s/syntax/LiteralsSyntax.scala#L25

exoego:
Thanks!

PR #2717, opened 2022-09-30
Fail when Mill can't extract dependencies JSON

missingfaktor:
If the output from the sandboxed Mill process exceed the configured buffer size, we lose the start of the JSON, leading to `extracted.dropWhile(!_.startsWith("{")).mkString("\n")` evaluating to an empty string, which causes other failures down the line. Those errors don't indicate the actual cause. As of this change, we fail early, right when the actual problem occurs.

Joint work with @IdiosApps.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
@@ -30,6 +30,7 @@ import org.scalasteward.core.vcs.data.BuildRoot
 
 final class MillAlg[F[_]](implicit
     fileAlg: FileAlg[F],
+    logger: Logger[F],
     processAlg: ProcessAlg[F],
     workspaceAlg: WorkspaceAlg[F],
     F: MonadCancelThrow[F]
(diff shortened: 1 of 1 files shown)


modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
```diff
@@ -47,9 +48,14 @@ final class MillAlg[F[_]](implicit
         val command = Nel("mill", List("-i", "-p", predef.toString, "show", extractDeps))
         processAlg.execSandboxed(command, buildRootDir)
       }
-      parsed <- F.fromEither(
-        parser.parseModules(extracted.dropWhile(!_.startsWith("{")).mkString("\n"))
-      )
+      jsonString = extracted.dropWhile(!_.startsWith("{")).mkString("\n")
+      _ <- F.whenA(jsonString.isBlank) {
```
exoego, line 52: `.isBlank` is added in JDK 11.
Since we still support JDK 8, it may be replaced with `jsonString.trim.isEmpty`.

modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
```diff
@@ -30,6 +30,7 @@ import org.scalasteward.core.vcs.data.BuildRoot
 
 final class MillAlg[F[_]](implicit
     fileAlg: FileAlg[F],
+    logger: Logger[F],
```
exoego, line 33: Not imported.

modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
exoego: Also please add test cases.

fthomas:
Superseded by #2940.

PR #2745, opened 2022-10-20
Update logback-classic only on 1.3.x series that supports JDK 8

exoego:
Supersedes #2719 

[logback-classic no longer supports JDK 8 starting from 1.4.0](https://logback.qos.ch/dependencies.html).
image

Let's pin it to 1.3.x series until we drop JDK 8 support.

### DIFF
--- a/.scala-steward.conf
+++ b/.scala-steward.conf
@@ -1,2 +1,6 @@
 updatePullRequests = "always"
 
+updates.pin = [
+  # We may drop this when we drop JDK 8 since minimum JDK is 11 from logback-classic 1.4.0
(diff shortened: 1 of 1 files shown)


alejandrohdezma (approved):
LGTM. Just curious, what's holding us back from dropping support for JDK 8?

exoego:
Good question.
Since Scala Steward is an app and not a library to be used in "enterprise" app, perhaps we can drop Java 8 soon.
It might be easy for users, who hosted their own Steward, to run Steward on Java 11.

exoego:
I've opened https://github.com/scala-steward-org/scala-steward/issues/2746 to consider dropping Java 8

PR #2714, opened 2022-09-23
Add support for grouping dependency updates

alejandrohdezma:
## 💻 How to review this PR?

This PR was created with the idea of being reviewed commit by commit.

Each commit contains an incremental change (and its tests, if applicable) that makes it easier to review. Also some of the commits contain additional information in their description to help understand why the change was made.

<details><summary><b>I also recommend checking "Hide whitespace" when reviewing this PR!</b></summary></br>

</details>

## 🎯 What's the goal of this PR?

This PR adds support for a long-requested feature of Scala Steward, the ability to group pull-request updates.

## ❓ How this will be done?

A new configuration value (`pullRequests.grouping`) can be added to the `.scala-steward.conf` file. This config will represent the desired PR groups.

A group will be an object with:

- `name` (mandatory): the name of the group, can be used for things like naming the branch.
- `title` (optional): if provided it will be used as the title for the PR.
- `filter` (mandatory): a non-empty list containing the filters to use to know if an update falls into this group. Filters would have this format:
  + `{version = "major" | "minor" | "patch" | "pre-release" | "build-metadata"}, group = "{group}", artifact = "{artifact}"}`.

For filters every component is optional but at least one must be provided. For grouping every update a filter like `{group = "*"}` can be provided.

<details><summary><b>Examples</b></summary></br>

```conf
pullRequests.grouping = [
  { name = "patches", title = "Patch updates", filter = [{version = "patch"}] },
  { name = "minor_major", title = "Minor/major updates", filter = [{version = "minor"}, {version = "major"}] }
]
```

☝🏼 This configuration will group updates in two PRs: one titled "Patch updates" that will contain patch version updates and another one titled "Minor/major updates" with the rest of them.

```conf
pullRequests.grouping = [
  { name = "typelevel", title = "Typelevel updates", filter = [{group = "org.typelevel"}, {group = "org.http4s"}] }
]
```

☝🏼 This configuration will group Typelevel and Http4s updates into one PR.  Since there is no other group, any update not falling into this group will follow normal behavior.

```conf
pullRequests.grouping = [
  { name = "my_libraries", filter = [{"artifact" = "my-library-*"}, {"artifact" = "my-other-library", "group" = "my-org.*"}] },
  { name = "all", title = "Not my libraries, filter = [{group = "*"] }
]
```

[...]

PR #2748, opened 2022-10-23
Rename the `AnUpdate` hierarchy

alejandrohdezma:
This new ADT was introduced in #2714 for enabling supporting grouping updates, but the naming was not very good. This PR fixes the hierarchy names.

## Renames in this PR
- `Update.Single` => `Update.ForArtifactId`
- `Update.Group` => `Update.ForGroupId`
- `Update` => `Update.Single`
- `GroupedUpdate` => `Update.Grouped`
- `AnUpdate` => `Update`

## Functional changes in this PR

Instead of deriving circe's `Codec` for the whole `Update` ADT, individual `Decoder` are created to ensure JSON created using older versions of Scala Steward will still be decoded.

```scala
implicit val ForGroupIdDecoder: Decoder[ForGroupId] = {
  val derived = deriveDecoder[ForGroupId]
  derived
    .prepare(_.downField("ForGroupId"))
    .or(derived.prepare(_.downField("Group")))
}

implicit val UpdateEncoder: Encoder[Update] = deriveEncoder

implicit val UpdateDecoder: Decoder[Update] =
    ForArtifactIdDecoder
      .widen[Update]
      .or(ForGroupIdDecoder.widen[Update])
      .or(Decoder[Grouped].widen[Update])
```

vs

```scala
implicit val UpdateCodec: Codec[Update] = deriveCodec
```

### DIFF
--- a/modules/benchmark/src/main/scala/org/scalasteward/benchmark/UpdatesConfigBenchmark.scala
+++ b/modules/benchmark/src/main/scala/org/scalasteward/benchmark/UpdatesConfigBenchmark.scala
@@ -33,7 +33,7 @@ class UpdatesConfigBenchmark {
     val newerVersions = Nel
       .of("2.0.0", "2.1.0", "2.1.1", "2.2.0", "3.0.0", "3.1.0", "3.2.1", "3.3.3", "4.0", "5.0")
       .map(Version.apply)
-    val update = Update.Single(dependency, newerVersions)
+    val update = Update.ForArtifactId(dependency, newerVersions)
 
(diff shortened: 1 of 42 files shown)


alejandrohdezma:
@fthomas can we instruct Codacy here to ignore the `_1`/`_2` usages on tuples? I don't think I have permission in Codacy to do so

exoego (approved):
Sorry for late response.
I was trying to find better naming/hierarchy for those Update-classes, but ended in smoke.

For this PR, findings by CodeCov and Codacy are trivial and negligible, I think.

PR #2762, opened 2022-11-11
Allow `--url-checker-test-url` option multiple times

fthomas:
This allows the `--url-checker-test-url` option to be used multiple times. Each argument of this option is checked with the `UrlChecker` in `SelfCheckAlg` at program start-up.

This is useful when Scala Steward is used in corporate environments where one wants to ensure that it can check internal and externals URLs.


exoego (approved):
LGTM

PR #2779, opened 2022-11-18
Replace Java 8 with newer LTS Java 17

### LINKED ISSUE #2746: Drop Java 8 Support
Since Scala Steward is an app and not a library to be used in "enterprise" app, perhaps we can drop Java 8 soon.
This makes maintenance a bit easier.

We can un-pin logback-classic updates when dropping Java 8. https://github.com/scala-steward-org/scala-steward/pull/2745#issuecomment-1285084527

Newer scalacache-caffeine also dropped Java 8. https://github.com/scala-steward-org/scala-steward/pull/2340

exoego:
Closes #2746

Unfortunately, not a small number of libraries used in Scala Steward dropped Java 8.
Since scala-steward is not a library, I think long-term Java 8 support is not needed fundamentally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         scala: [2.12.17, 2.13.10]
-        java: [adopt-hotspot@8, adopt-hotspot@11]
+        java: [adopt-hotspot@17, adopt-hotspot@11]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout current branch (full)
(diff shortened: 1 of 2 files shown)


exoego:
> Since scala-steward is not a library, I think long-term Java 8 support is not needed fundamentally.

Hmm, `mill-plugin` and `sbt-plugin` may be affected?

alejandrohdezma (approved):
Shouldn't we do the same with Java 11 and only support 17?

exoego:
That is a good point.
I want to propose "last two LTS" strategy so it gives some time to prepare to users.

alejandrohdezma:
Yeah, sounds like a sensible idea 😸

Issue #2608, opened 2022-05-19
GitHub actions workflow `.yml` file not updated


mkurz:
See https://github.com/playframework/play-json/pull/759
Shouldn't also the Scala version be updated in a GHA workflow file? At the time the pull request got opened by scala-steward the exact same version (`3.1.3-RC2`) could be found in https://github.com/playframework/play-json/blob/9bd7d5ec6b6f3f42eeb63fbedd3efa7ac30005e5/.github/workflows/build-test.yml#L41. However, scala steward did *not* update the version in that file...

According to the default configs
* https://github.com/scala-steward-org/scala-steward/blob/0a4d1a09b47760b51c89870db3ba25430aae8ad2/docs/repo-specific-configuration.md?plain=1#L61 and 
* https://github.com/scala-steward-org/scala-steward/blob/d43fad147af294abd64f745cc6120e4993075ae1/modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala#L76-L77

shouldn't those yml files be updated as well?

exoego:
I’ve tried to reproduce in https://github.com/scala-steward-org/scala-steward/pull/2780 but GitHub action file was updated in the test case at least 🤔

PR #2780, opened 2022-11-18
Check files under .github updated

exoego:
Closes https://github.com/scala-steward-org/scala-steward/issues/2608


fthomas:
In the last commit I used the same artifactIds as in the play-json PR and also added the file where Scala Steward did actually update the version. It is failing now as reported in #2608.

fthomas:
I also tried this case with #2816 but it is also failing there.

mzuehlke:
I converted the test to a `RewriteTest` and found that the `.github/` folder is not the problem, but the key `scala` in the `ci.yml` prevents the update. If I change the key to `scala3`it works.

The following tests passes:
```Scala
  test("Files under .github yml should be updated".only) {
    val artifactIds = Nel.of("scala3-compiler".a, "scala3-library".a, "scala3-library_sjs1".a)
    val update = ("org.scala-lang".g % artifactIds % "3.1.3-RC2" %> "3.1.3-RC3").group
    val original = Map(
      "project/Dependencies.scala" -> """val Scala3 = "3.1.3-RC2" """,
      ".github/workflows/ci.yml" -> """scala3: 2.12.15, 2.13.8, 3.1.3-RC2"""
    )
    val expected = Map(
      "project/Dependencies.scala" -> """val Scala3 = "3.1.3-RC3" """,
      ".github/workflows/ci.yml" -> """scala3: 2.12.15, 2.13.8, 3.1.3-RC3"""
    )
    runApplyUpdate(update, original, expected)
  }	
```

fthomas:
> [...] but the key `scala` in the `ci.yml` prevents the update. If I change the key to `scala3`it works.

I think we can make it work, if we add `scala` as a search term if one the search terms is `scala3`. Or we strip trailing digits from all search terms.

Issue #2770, opened 2022-11-17
Run `scala-collection-compat` on update to Scala 2.13


EwoutH:
When the Scala is bumped to 2.13, it would be great if the [scala-collection-compat](https://github.com/scala/scala-collection-compat) migration rules could be ran.

```scala
dependency:Collection213Upgrade@org.scala-lang.modules:scala-collection-migrations:2.8.1
```

exoego:
For volunteer, https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/resources/scalafix-migrations.conf should be updated

alifirat:
Hey @EwoutH
The migration rule should be apply when the scala version is bumped from "2.12.X" to "2.13.X" right?

EwoutH:
Hi @alifirat! Yes, that should be the case.

fthomas:
@EwoutH, can this be closed?

EwoutH:
Thanks for this effort. I correctly attempted the run when updating to Scala 2.13, but in our setup it unfortunately failed. But that could also be due to the rule itself. See the [log here](https://github.com/EwoutH/NetLogo/actions/runs/3752758458/jobs/6375284107#step:6:213).

(CC @SethTisue)

fthomas:
The errors in the logs are:
```
2022-12-21 21:01:39,318 INFO  Running migration ScalafixMigration(org.scala-lang,NonEmptyList(scala-library, scala-compiler, scala-reflect, scalap),2.13.0,NonEmptyList(dependency:Collection213Upgrade@org.scala-lang.modules:scala-collection-migrations:2.8.1),Some(https://github.com/scala/scala-collection-compat#collection213upgrade),Some(NonEmptyList(-P:semanticdb:synthetics:on)),None,None,None)
2022-12-21 21:02:12,628 WARN  Scalafix migration failed
java.io.IOException: '"SBT_OPTS=-Xmx2048m -Xss8m -XX:MaxMetaspaceSize=512m" sbt -Dsbt.color=false -Dsbt.log.noformat=true -Dsbt.supershell=false ;scalafixEnable;scalafixAll dependency:Collection213Upgrade@org.scala-lang.modules:scala-collection-migrations:2.8.1' exited with code 1
...
[...]

PR #2799, opened 2022-11-28
Add scala-collection-compact Scalafile rule when bump from 2.12 to 2.13

alifirat:
Address #2770 

I'm not sure if my rule is too specific to the `scala-compact-library` and should be more general to the scala bumping version.

### DIFF
--- a/modules/core/src/main/resources/scalafix-migrations.conf
+++ b/modules/core/src/main/resources/scalafix-migrations.conf
@@ -205,5 +205,13 @@ migrations = [
     newVersion: "0.14.0",
     rewriteRules: ["github:trace4cats/trace4cats-scalafix/v0_14?sha=v0.14.0"],
     doc: "https://github.com/trace4cats/trace4cats-docs/blob/master/docs/migrating.md"
+  },
+  {
+    groupId: "org.scala-lang",
+    artifactIds: ["scala-library", "scala-compiler", "scala-reflect", "scalap"],
+    newVersion: "2.13.0",
+    doc: "https://github.com/scala/scala-collection-compat#collection213upgrade",
+    rewriteRules: ["dependency:Collection213Upgrade@org.scala-lang.modules:scala-collection-migrations:2.8.1"], 
+    scalacOptions: ["-P:semanticdb:synthetics:on"]
   }
 ]


modules/core/src/main/resources/scalafix-migrations.conf
```diff
@@ -205,5 +205,11 @@ migrations = [
     newVersion: "0.14.0",
     rewriteRules: ["github:trace4cats/trace4cats-scalafix/v0_14?sha=v0.14.0"],
     doc: "https://github.com/trace4cats/trace4cats-docs/blob/master/docs/migrating.md"
-  }
+  },
+    {
+    groupId: "org.scala-lang.modules",
```
alejandrohdezma, line 210: **question (blocking):** Shouldn't this be run when updating Scala itself (instead of `scala-collection-compat`)?

alifirat: That was my remark: I have the feeling that the rule is too specific to the collection itself while it should be more oriented to the scala version. 
In this case, what should be the value of `newVersion` ? "2.13.0" or we case use a expression like 'newVersion = 2.13. ' to match any update to Scala 2.13 ?

alejandrohdezma: Yeah, I believe this should be:

```conf
  {
    groupId: "org.scala-lang",
    artifactIds: ["scala-library", "scala-compiler", "scala-reflect", "scalap"],
    newVersion: "2.13.0",
    doc: "https://github.com/scala/scala-collection-compat#collection213upgrade",
    rewriteRules: ["dependency:Collection213Upgrade@org.scala-lang.modules:scala-collection-migrations:2.8.1"]
  },
```

alejandrohdezma: WDYT @exoego?

fthomas: Yeah, I think this makes more sense.

The docs of that rewrite also mention that this Scalafix requires `scalacOptions: ["-P:semanticdb:synthetics:on"]`.

> In this case, what should be the value of `newVersion` ? "2.13.0" or we case use a expression like 'newVersion = 2.13. ' to match any update to Scala 2.13 ?

`2.13.0` is fine because Scala Steward will use that the rule when it updates from a version `< 2.13.0` to a version `>= 2.13.0`.

alifirat: Then, if we all agree I can update the PR with your feedbacks?

alifirat: @fthomas @alejandrohdezma thanks for the feedback, I've rebased and updated the PR. 
Tell me if it's good to you.

exoego (approved):

alejandrohdezma (approved):

EwoutH:
Thanks a lot for this! Any way that I can test it before the next release is tagged?

fthomas:
You don't need to use the latest Scala Steward version to test this. Scalafix migrations are available to all Scala Steward instances as soon as they are merged here (unless that is disabled by the `--disable-default-scalafix-migrations` command-line option).

EwoutH:
Thank you both!

PR #2801, opened 2022-11-29
Revert "Update logback-classic only on 1.3.x series that supports JDK 8"

fthomas:
Reverts scala-steward-org/scala-steward#2745

Unpin logback-classic as mentioned in https://github.com/scala-steward-org/scala-steward/issues/2746#issue-1416131445.


exoego (approved):
Thanks!

PR #2807, opened 2022-12-01
Add note to default config files that are downloaded at start-up

fthomas:
This should clarify that the latest versions of these default config files are downloaded by all Scala Steward instances.


alejandrohdezma (approved):

exoego (approved):
nice

PR #2810, opened 2022-12-06
Add Bitbucket default reviewers option

rolang:
Hello,
this PR adds default reviewers option for the `bitbucket` vcs type the same way it's working with `bitbucket-server`. Would be great to get it supported. We have this implementation running in our CI and it seems to work well.


rolang:
will have a look on code coverage later

exoego:
Thanks

PR #2816, opened 2022-12-12
Untangle `UpdateHeuristic` into separate tasks

### LINKED ISSUE #1184: Nonexistent cats-effect version

### LINKED ISSUE #1534: Support custom heuristics

### LINKED ISSUE #1699: Incorrect library update
Hi,

I've noticed that Scala Steward is trying to bump incorrect library. It generated MR in Gitlab with title `Update async-http-client-backend-cats, ... to 2.2.9` from branch `update/async-http-client-backend-cats-2.2.9`, so it means that it will update `async-http-client-backend-cats` to version `2.2.9` from version `2.2.0`

But it generated following diff:

```diff
   val cats = Seq(
-      "org.typelevel" %% "cats-core" % "2.2.0",
-      "org.typelevel" %% "cats-effect" % "2.2.0",
+      "org.typelevel" %% "cats-core" % "2.2.9",
+      "org.typelevel" %% "cats-effect" % "2.2.9",
       "com.olegpy" %% "meow-mtl-core" % "0.4.1" % "test"
    )
```

which is incorrect, because it is going to update packages `cats-core` and `cats-effect` to version `2.2.9` (which doesn't exist)

### LINKED ISSUE #1977: Artifact migration fails if version and groupId/artifactId are in different files
See #1978.

### LINKED ISSUE #2099: Maven pom.xml update does not happen
If we mention a dependency in pom.xml in the normal format (shown below) scala steward finds the update but fails to process the update and create a PR
```
        <dependency>
            <groupId>commons-io</groupId>
            <artifactId>commons-io</artifactId>
            <version>2.6</version>
        </dependency>
```
Log
```
18:41:49 2021-04-26 01:41:49,335 INFO  Process update commons-io:commons-io : 2.6 -> 2.8.0
18:41:49 2021-04-26 01:41:49,406 INFO  Trying heuristic 'moduleId'
18:41:49 2021-04-26 01:41:49,411 INFO  Trying heuristic 'strict'
18:41:49 2021-04-26 01:41:49,415 INFO  Trying heuristic 'original'
18:41:49 2021-04-26 01:41:49,419 INFO  Trying heuristic 'relaxed'
18:41:49 2021-04-26 01:41:49,424 INFO  Trying heuristic 'sliding'
18:41:49 2021-04-26 01:41:49,435 INFO  Trying heuristic 'completeGroupId'
18:41:49 2021-04-26 01:41:49,440 INFO  Trying heuristic 'groupId'
18:41:49 2021-04-26 01:41:49,440 INFO  Trying heuristic 'specific'
18:41:49 2021-04-26 01:41:49,440 WARN  Unable to bump version
18:41:49 2021-04-26 01:41:49,440 WARN  No commits created 
```

But same works if the version is specified in common properties
```
  <properties>
    <commons-io.version>2.6</commons-io.version>
  </properties>
...
        <dependency>
            <groupId>commons-io</groupId>
            <artifactId>commons-io</artifactId>
            <version>${commons-io.version}</version>
        </dependency>
```

### LINKED ISSUE #2162: Updating random values in the codebase that match the updated version.
The repo that this happened in is private, but I'll do my best to outline what happened below. If there is anything else you need to better identify what may have happened, let me know.

```scala
lazy val library = new Object {
  object Version {
        val akkaAlpakka                = "3.0.0"
   }

  val akkaAlpakkaS3          = "com.lightbend.akka"     %% "akka-stream-alpakka-s3"    % version.akkaAlpakka
  val akkaStreamAlpakkaFile  = "com.lightbend.akka"     %% "akka-stream-alpakka-file"  % version.akkaAlpakka
}
```
The pr sent in changed this `3.0.0` to `3.0.1` correctly in the `build.sbt`, but it included a warning that it wasn't updated in a few places:

```
The following files still refer to the old version number (3.0.0).
You might want to review and update them manually.
package-lock.json
scalableTest-compose/package-lock.json
server/src/main/resources/<redacted>/<redacted>_OD_v3.0.0_<redacted>-1.0.2.xml
server/src/test/scala/<redacted>/repositories/<redacted>Spec.scala
```
That was fine, however there were a couple places in the codebase that `3.0.0` was updated to `3.1.1`.

```diff
  lazy val `Module Object Recognition v3.0`: OurType =
-    load(resourceName = "<redacted>/<redacted>_OD_v3.0.0_<redacted>-1.0.2.xml")
+    load(resourceName = "<redacted>/<redacted>_OD_v3.0.1_<redacted>-1.0.2.xml")
```

Another was in a spec file:

```diff
 "some test name" in { f =>
      for {
        _ <- f.<redacted>.persist(
          Seq(
            ...
-            f.stuff.copy(<redacted>Version    = "3.0.0")
+            f.stuff.copy(<redacted>Version    = "3.0.1")
          )
        )
...
```

I'm assuming this has to do with the first one being a `lazy val` maybe and the second included `Version =`?

### LINKED ISSUE #2412: False matching in yaml file.
At work I noticed today a bunch of updates were sent in that ended up breaking builds due to an incorrect matching in a yaml file. The example that it happened in can be minimized to:

```scala
lazy val library = new {
  object version {
    val dep = "1.3.0"
  }

  val fullDep = "org" %% "name" % version.dep
}
```

A pr was sent in to update this dep from `1.3.0` -> `1.3.1` which was correct. However at the root of our repo we also have a `build.yml` file which is used in CI to set up our env. In that `build.yml` there was the following:

```yaml
image: our-registry/name:tagname-v1.3.0-1
```

Steward also tried changing this.

```diff
- image: our-registry/name:tagname-v1.3.0-1
+ image: our-registry/name:tagname-v1.3.1-1
```

I'd expect this not to update the yaml file especially since the version imo doesn't match. `1.3.0` should be matched with `1.3.0-1`.

### LINKED ISSUE #2664: Some dependencies not updated
I didn't see another issue like this one, so I'm opening this.

I saw a few dependencies were outdated in a project despite daily scala-steward runs (on latest image). I confirmed there were updates with `mill mill.scalalib.Dependency/showUpdates`

Here's some examples:
- log4j api/core/slf4j-impl (2.16.0 -> 2.17.2)
- refined (0.9.14 -> 0.9.28)
- os-lib (0.7.8 -> 0.8.1)
- wiremock-jre8 (2.31.0 -> 2.33.2)

In the case of slf4j, 2.16.0 was Dec 2021 and 2.17.2 is Feb 2022. I manually updated it in late June, so Scala Steward wasn't picking up those updates for about half a year.

Some of these in an dependencies.sc file have the version as part of the value. Some have it interpolated into the ivy dependency. I don't see an obvious pattern for why some things get updates, but others don't.

Posssible causes
- Does scala steward cache failures from `fail-fast`, so if you have a failed CI check then you won't ever get e.g. 0.0.1 -> 0.0.2 of a dependency? This is a total guess. I don't know how scala-steward works.

### LINKED ISSUE #2675: Erroneously updating `http4s-okhttp-client` to non-existent versions
Since [The Great Schism](https://github.com/http4s/http4s/releases/tag/v0.23.12) of http4s, Scala Steward has been trying to update `http4s-okhttp-client` to versions that do not exist. Unfortunately, this is in a private repository but I'll try and include the full details here.

My dependencies are defined in a separate file, and look a bit like the following. Note that `http4s-okhttp-client` is defined separately with its own version number. `lazy val http4sOkHttp = "0.23.11"` is only used in one place, as shown:
```
object Dependencies {
  object Versions {
    lazy val http4s = "0.23.11"

    lazy val http4sOkHttp = "0.23.11"

    lazy val http4sPrometheusMetrics = "0.24.0"
    // ... etc
  }

  object Libs {  
    lazy val http4sDsl                 = "org.http4s"            %% "http4s-dsl"                % Versions.http4s
    lazy val http4sOkHttpClient        = "org.http4s"            %% "http4s-okhttp-client"      % Versions.http4sOkHttp
    lazy val http4sBlazeServer         = "org.http4s"            %% "http4s-blaze-server"       % Versions.http4s
    lazy val http4sCirce               = "org.http4s"            %% "http4s-circe"              % Versions.http4s
    lazy val http4sMetrics             = "org.http4s"            %% "http4s-prometheus-metrics" % Versions.http4sPrometheusMetrics
    // ... etc
  }

  object Module {  
    lazy val `my-module-deps` = Seq(
      http4sDsl,
      http4sOkHttpClient,
      http4sBlazeServer,
      http4sCirce,
      http4sMetrics,
      // ... etc
    )
  }
}
```

Whenever Scala Steward I get PRs raised with titles like "Upgrade http4s-blaze-server from 0.23.11 to 0.23.12" but with the following diff:

```
-    lazy val http4s = "0.23.11"
+    lazy val http4s = "0.23.12"

     lazy val http4sPrometheusMetrics = "0.24.0"

-    lazy val http4sOkHttp = "0.23.11"
+    lazy val http4sOkHttp = "0.23.12"
```

While it's true this does update http4s-blaze-server from 0.23.11 to 0.23.12, it has also updated `http4s-okhttp-client` to a non-existent version.

Running `dependencyUpdates` shows all the correct versions, and without any mention of updating the `http4s-okhttp-client` dependency:

```
[info] Found 3 dependency updates for my-module
[info]   org.http4s:http4s-blaze-server : 0.23.11 -> 0.23.12
[info]   org.http4s:http4s-circe        : 0.23.11 -> 0.23.13
[info]   org.http4s:http4s-dsl          : 0.23.11 -> 0.23.13
```

This suggests the problem is either with how Scala Steward is parsing the dependencies, or how it updates the files afterwards.

It isn't the end of the world, I can manually patch the PRs that Scala Steward is raising. This is a bit of a papercut though, and It does also mean that we have to be careful to make sure we aren't automatically merging any of these changes.

Apologies I can't link you to a repository, hopefully this is enough information to be helpful and thank you!

### LINKED ISSUE #2716: Group update works depending on names of updated dependencies
At our company, we release two libraries together with the same version number:
`com.example:foo-lastmile` and
`com.example:foo-scorer`.

Scala Steward does not update one project that relies on both dependencies. Through debugging, I was able to figure out that the names of the dependencies plays a role. Scala Steward [sorts updates by name](https://github.com/scala-steward-org/scala-steward/blob/fa6f7f0505e964b599a32c023cc4fe491c2f613b/modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala#L45); therefore, the `Update.Group` will contain `foo-lastmile` first and `foo-scorer` second.

This update fails, as the following test proves:
```scala
  test("fails") {
    val original =
      """private val version = "1.0.0"
        |
        |lazy val foo = "com.example" % "foo-lastmile" % version
        |lazy val foo = "com.example" % "foo-scorer" % version
        |""".stripMargin
    val expected =
      """private val version = "1.1.0"
        |
        |lazy val foo = "com.example" % "foo-lastmile" % version
        |lazy val foo = "com.example" % "foo-scorer" % version
        |""".stripMargin

    val update = ("com.example".g % Nel.of(
      "foo-lastmile".a,
      "foo-scorer".a,
    ) % "1.0.0" %> "1.1.0").group
    assertEquals(update.replaceVersionIn(original)._1, Some(expected))
  }
```

However, if we change the name of `foo-lastmile` to `foo-zzz` so that it comes _after_ `foo-scorer`, the test suddenly passes:
```scala
  test("passes") {
    val original =
      """private val version = "1.0.0"
        |
        |lazy val foo = "com.example" % "foo-zzz" % version
        |lazy val foo = "com.example" % "foo-scorer" % version
        |""".stripMargin
    val expected =
      """private val version = "1.1.0"
        |
        |lazy val foo = "com.example" % "foo-zzz" % version
        |lazy val foo = "com.example" % "foo-scorer" % version
        |""".stripMargin

    val update = ("com.example".g % Nel.of(
      "foo-scorer".a,
      "foo-zzz".a,
    ) % "1.0.0" %> "1.1.0").group
    assertEquals(update.replaceVersionIn(original)._1, Some(expected))
  }
```

In other words, the alphabetical order of the dependencies' names being updated plays a role in whether Scala Steward can update them or not, and I think this should not be the case.

This was tested on version 0.15.0.

I'd be happy to assist in investigating the problem. I tried to understand the update heuristics' code, but found the logic quite hard to follow, so I would need to pair on that with someone.

### LINKED ISSUE #2811: Use a more principled approach for replacing version numbers in files
One of the oldest and opaque parts of Scala Steward is [`UpdateHeuristic`](https://github.com/scala-steward-org/scala-steward/blob/48b8d97b82a1a082b7a6960d3f717b1e7dc3fed0/modules/core/src/main/scala/org/scalasteward/core/edit/UpdateHeuristic.scala). It is responsible for locating the correct version number in files and for updating them. It is hard to understand and to extend because it does three things at the same time per file:
 1. locating the positions of the old version
 2. deciding which version number is the correct one (i.e. corresponds to the artifacts that are updated)
 3. replacing the old version number with the new one

I would like to retire `UpdateHeuristic` with an approach where the three steps above are completely separated and done sequentially. I envision that it works more like this:
1. locate all positions of the old version number in all files and classify them (e.g. if they are part of a sbt ModuleID, a Maven dependency definition, or the value of a `val`)
2. look at all of the above version positions and decide which one of those should be updated
3. replace the version in the files and positions selected by step 2

I think the reason why this is easier to understand and to extend is that we would have in step 2 all positions of the old version numbers at hand and then can make an informed decision based on the classification that happened in step 1.

I'm working on a prototype of this approach in https://github.com/scala-steward-org/scala-steward/compare/wip/version-position and it looks promising so far. For example, problems as described in https://github.com/scala-steward-org/scala-steward/issues/1184 are easy to fix since matching the `val` used in the ModuleID with the `val` used for the version number is easy. This approach would also eliminate current limitations of `UpdateHeurisitc` like multiline `val`s or Maven dependency definitions where groupId, artifactId, and version are on different lines.

Comments about this approach or other ideas how to do this even better are very welcome!

### LINKED ISSUE #2836: scala-steward detects that it should update a version but then fails to bump it
I'm working on setting up scala-steward for some internal projects. There is one in particular that is showing some very strange behavior:

The version is defined as a single `private val FooVersion = "22.12.2"` in a `project/Dependencies.scala` file. This val is then re-used to define nine different dependencies within the same file.

Scala steward recognizes that it should update the version:

```
2022-12-17 00:53:01,826 INFO  Found 9 updates:
com.foo:(aaa, aaa_2.13) : 22.12.2 -> 22.12.4
com.foo:(bbb, bbb_2.13) : 22.12.2 -> 22.12.4
com.foo:(ccc, ccc_2.13) : 22.12.2 -> 22.12.4
com.foo:(ddd, ddd_2.13) : 22.12.2 -> 22.12.4
com.foo:(eee, eee_2.12) : 22.12.2 -> 22.12.4
com.foo:(fff, fff_2.13) : 22.12.2 -> 22.12.4
com.foo:(ggg, ggg_2.13) : 22.12.2 -> 22.12.4
com.foo:(hhh, hhh_2.13) : 22.12.2 -> 22.12.4
com.foo:(iii, iii_2.13) : 22.12.2 -> 22.12.4
```

But then it fails to do so, all nine times:

```
2022-12-17 00:53:04,617 INFO  Trying heuristic 'moduleId'
2022-12-17 00:53:04,628 INFO  Trying heuristic 'strict'
2022-12-17 00:53:04,632 INFO  Trying heuristic 'original'
2022-12-17 00:53:04,637 INFO  Trying heuristic 'relaxed'
2022-12-17 00:53:04,642 INFO  Trying heuristic 'sliding'
2022-12-17 00:53:04,648 INFO  Trying heuristic 'completeGroupId'
2022-12-17 00:53:04,653 INFO  Trying heuristic 'groupId'
2022-12-17 00:53:04,713 INFO  Trying heuristic 'specific'
2022-12-17 00:53:04,714 WARN  Unable to bump version
```

(The above is repeated nine times exactly)

So it seems like the UpdateHeuristics are failing to replace the version. But I'm very confused how that's possible given it was able to find the versions in the first place. It should only have to update that single `private val`.

This has happened on 0.16.1 and 0.17.0. I haven't tried others.

LMK if there's a suggested way to diagnose and/or reproduce. Obviously I'm not going to post the full project, but maybe I can provide some approximation of it. But I'm not exactly sure what's the best way to package that up.

fthomas:
This PR implements the idea outlined in #2811.

# The big picture

* `UpdateHeuristic` is gone.
* `VersionScanner` finds all positions of a version in a string and classifies them.
* `ModuleScanner` finds all positions of a artifactId and groupId pair in a string and classifies them.
* `ScanAlg` uses `VersionScanner` and `ModuleScanner` to find all version positions and module positions in all relevant files.
* `Selector` takes an update and corresponding version and module positions to produce a list of `Substring.Replacement`s that describe what needs to be replaced in which file. The heuristics that were in `UpdateHeuristic` are now here.
* `EditAlg` uses `ScanAlg` and `Selector` to find the list of replacements and then later performs the actual editing of files and creates the update commit as before.

# Benefits of this approach

* Finding versions, updating them, and writing the results to disk are separated tasks that can be understood, changed, and improved independently.
* Classification of version and module positions and retaining relevant context in the `*Position` subclasses helps a lot with finding the correct positions where to update versions and artifacts. This makes, for example, updating Maven dependencies or matching version `val`s with the corresponding dependency definition easy and allows to fix a lot of reported bugs.
* `EditAlg.applyUpdate` can be simplified because trying an update heuristic does not write to disk and so we don't need to discard previous changes before we apply Scalafix migrations.

# How to review?

This PR started as a proof of concept and most additions underwent significant changes later. It is probably easier to review the whole diff at once. 

closes #1184, closes #1534, closes #1699, closes #1977, closes #2099, closes #2162, closes #2412, closes #2664, closes #2675, closes #2716, closes #2811, closes #2836


exoego (approved):
This is awesome 🥇 
I checked the existing test cases were consolidated or moved to new files.

alejandrohdezma:
I did a first diagonal reading and...



I would like to do a more thorough review during the weekend. @fthomas do you mind holding off merging until then?

fthomas:
> I checked the existing test cases were consolidated or moved to new files.

Yeah, I should have mentioned that. Every test in `UpdateHeuristicTest` and some of the `EditAlgTest` tests are now in `RewriteTest`. I also added a bunch of new tests that show that the linked issues are actually fixed. One benefit of `RewriteTest` over `UpdateHeuristicTest` is that it calls `EditAlg#applyUpdate` and therefore covers more production code then `UpdateHeuristicTest`. It is also possible to test scenarios that involve multiple files.

fthomas:
> @fthomas do you mind holding off merging until then?

Sure, there is no need to rush this.

I'm also happy to document or explain the parts that are unclear or intricate.

modules/core/src/main/scala/org/scalasteward/core/edit/update/data/ModulePosition.scala
```diff
@@ -0,0 +1,46 @@
...
+      groupId: Substring.Position,
+      artifactId: Substring.Position,
+      version: Substring.Position
+  ) extends ModulePosition
+
+  final case class MavenDependency(
+      groupId: Substring.Position,
+      artifactId: Substring.Position,
+      version: Substring.Position
+  ) extends ModulePosition
```
alejandrohdezma, line 45: **question:** why do we need to differentiate between these three? are we expecting projects to use more than one build system?

alejandrohdezma:
@fthomas, would you be able to add some scaladocs to some of the new classes? It will sure help future maintenance

modules/core/src/main/scala/org/scalasteward/core/edit/update/data/ModulePosition.scala
fthomas: We currently don't need these three. `ModulePosition` could just be a case class with the three `Position` fields. I guess I added these because `ModuleScanner` has different `find` methods for each of them. But we currently don't differentiate between the types of `ModulePosition` later.

I'll change them to a single case class.

fthomas: Done in [30e1cd4](https://github.com/scala-steward-org/scala-steward/pull/2816/commits/30e1cd4c1c556b7ce23af535a18727bea846938f).

fthomas: > are we expecting projects to use more than one build system?

Yes, projects that use more than one build system exist. One example is https://github.com/com-lihaoyi/scalatags which has a sbt and Mill build. And Scala Steward uses both to extract dependencies.

alejandrohdezma: Oh! Great to know that 😊 thanks!

fthomas:
@alejandrohdezma I added Scaladocs to the new classes. Feel free to point me to any of them, if they should be extended, are unclear, or confusing. :-)

mzuehlke (approved):
That is great simplification for the heart of `scala-steward`.
The fact this enables to fix and verify 12 issues speaks for itself. 👍 

One tiny suggestion would be to make 2 names more explicit:
*  `ModuleScanner` => `ModulePositionScanner`
* `VersionScanner`=> `VersionPositionScanner`

fthomas:
Thank you @mzuehlke!

> One tiny suggestion would be to make 2 names more explicit:

These are good suggestions. I'll change them.

alejandrohdezma (approved):
Sorry for the delay 😅, between the festivities and the flu I hadn't touch a computer in some time.

One thing I see missing that could benefit both maintaining all these as well as showing what Scala Steward can do is a set of integration tests with "real" files on `resources` like `build.sbt_before`/`build.sbt_after`. Nevertheless this PR is perfect, if you like the idea I can work on it in a subsequent PR.

fthomas:
> Sorry for the delay sweat_smile, between the festivities and the flu I hadn't touch a computer in some time.

No worries! And don't feel bad about it.

> One thing I see missing that could benefit both maintaining all these as well as showing what Scala Steward can do is a set of integration tests with "real" files on `resources` like `build.sbt_before`/`build.sbt_after`. Nevertheless this PR is perfect, if you like the idea I can work on it in a subsequent PR.

Sounds like a good idea to me! `RewriteTest` is similar but the files there only contain the parts that are interesting to Scala Steward. So I guess the focus of these tests would be different.

Regarding your suggestions about performance optimizations: these are good suggestions and I'd like to do them in a subsequent PR after this new approach has been proven itself in the wild.

fthomas:
I know there is a test case for this but seeing diffs like this in the wild still makes me happy: https://github.com/7mind/izumi/pull/1849/files

PR #2820, opened 2022-12-13
fix(maven): exclude transitive deps for Maven

### LINKED ISSUE #2819: Incorrect pr sent in for an artifact found in test code
I have a small Maven project that as part of the tests checks to see what jars exist in a module. Steward today sent in a PR to update one of those jar versions even though it's not defined anywhere that I'm using this. Plus this is just code inside of one of my Scala files that I don't believe Steward should be touching unless it was somehow linked to my build definition. You can see the example of the pull request here in [scalacenter/bloop-maven-plugin](https://github.com/scalacenter/bloop-maven-plugin/pull/33).

- For context the job that created the pr can be found [here](https://github.com/scalacenter/steward/actions/runs/3681007721/jobs/6227224990). 
- Steward version - `v0.16.1`

ckipp01:
Fixes #2819

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/MavenAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/MavenAlg.scala
@@ -42,7 +42,7 @@ final class MavenAlg[F[_]](config: Config)(implicit
     for {
       buildRootDir <- workspaceAlg.buildRootDir(buildRoot)
       dependenciesRaw <- exec(
-        mvnCmd(command.listDependencies),
+        mvnCmd(command.listDependencies, args.excludeTransitive),
         buildRootDir
       )
       repositoriesRaw <- exec(
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/command.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/command.scala
@@ -20,3 +20,7 @@ object command {
   val listDependencies = "dependency:list"
   val listRepositories = "dependency:list-repositories"
 }
+
+object args {
+  val excludeTransitive = "-DexcludeTransitive=true"
+}
--- a/modules/core/src/test/scala/org/scalasteward/core/buildtool/maven/MavenAlgTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/buildtool/maven/MavenAlgTest.scala
@@ -25,7 +25,8 @@ class MavenAlgTest extends FunSuite {
           "--env=VAR2=val2",
           "mvn",
           "--batch-mode",
-          command.listDependencies
+          command.listDependencies,
+          args.excludeTransitive
         ),
         Cmd(
           repoDir.toString,


modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/command.scala
```diff
@@ -17,6 +17,6 @@
 package org.scalasteward.core.buildtool.maven
 
 object command {
-  val listDependencies = "dependency:list"
+  val listDependencies = "dependency:list -DexcludeTransitive=true"
```
fthomas, line 20: This needs to be a separate argument in https://github.com/scala-steward-org/scala-steward/blob/d3e22af810fa71b8917fd47e5ba128b461b8b197/modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/MavenAlg.scala#L45
Maven will fail otherwise with `[info] [ERROR] Could not find goal 'list -DexcludeTransitive=true'`.

ckipp01: Ah, how do the `MavenAlgTest` pass then? I guess I would have expected that to blow up in this case.

fthomas: `MavenAlgTest` unfortunately does not actually run Maven but only checks that it would call `mvn` with the expected args.

modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/MavenAlg.scala
```diff
@@ -42,7 +42,7 @@ final class MavenAlg[F[_]](config: Config)(implicit
     for {
       buildRootDir <- workspaceAlg.buildRootDir(buildRoot)
       dependenciesRaw <- exec(
-        mvnCmd(command.listDependencies),
+        mvnCmd(command.listDependencies ++ args.excludeTransitive),
```
fthomas, line 45:
```suggestion
        mvnCmd(command.listDependencies, args.excludeTransitive),
```

modules/core/src/main/scala/org/scalasteward/core/buildtool/maven/command.scala
```diff
@@ -20,3 +20,7 @@ object command {
   val listDependencies = "dependency:list"
   val listRepositories = "dependency:list-repositories"
 }
+
+object args {
+  val excludeTransitive = " -DexcludeTransitive=true"
```
fthomas, line 25:
```suggestion
  val excludeTransitive = "-DexcludeTransitive=true"
```

modules/core/src/test/scala/org/scalasteward/core/buildtool/maven/MavenAlgTest.scala
```diff
@@ -25,7 +25,7 @@ class MavenAlgTest extends FunSuite {
           "--env=VAR2=val2",
           "mvn",
           "--batch-mode",
-          command.listDependencies
+          command.listDependencies ++ args.excludeTransitive
```
fthomas, line 28:
```suggestion
          command.listDependencies,
          args.excludeTransitive
```

fthomas (approved):
Thanks @ckipp01!

PR #2833, opened 2022-12-16
Use `Sync[F].blocking` for creating system processes

### LINKED ISSUE #2790: Your CPU is probably starving... logged
We are seeing below errors when running scala steward on github actions. Mainly at the start of the run, I guess it improves with jvm warmup.

```
  Warning:  Your CPU is probably starving. Consider increasing the granularity
  Error: of your delays or adding more cedes. This may also be a sign that you are
  Error: unintentionally running blocking I/O operations (such as File or InetAddress)
  Error: without the blocking combinator.
```

This is coming from cats-effect
https://github.com/typelevel/cats-effect/blob/72fc25482fe01413a598432c522cb92dc3dcff3f/core/shared/src/main/scala/cats/effect/CpuStarvationCheck.scala#L43

Just wanted to mention this, if this is a non-issue feel free to close.

fthomas:
Maybe fixes #2790.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/io/process.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/io/process.scala
@@ -55,7 +55,7 @@ object process {
           }
         }
 
-        val fallback = F.delay(process.destroyForcibly()) >> {
+        val fallback = F.blocking(process.destroyForcibly()) >> {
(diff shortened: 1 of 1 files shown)


alejandrohdezma (approved):

exoego (approved):
Let's see how it works

PR #2844, opened 2022-12-21
CI: Introduce link checker job

exoego:
Follow-up of #2842 

- Checker typically ends within 30 seconds which sounds acceptable. Faster than Scala jobs anyway.
- Check is skipped if no `~.md` are changed, so false alerts due to tentative network failure are less likely to happen 

https://github.com/gaurav-nelson/github-action-markdown-link-check is used since it is highly starred in this area and its README are well-written.

Also fixed some "broken" links found in test runs.

### Failure

https://github.com/exoego/scala-steward/actions/runs/3746779243/jobs/6362385292

image

image

### Succcess
https://github.com/exoego/scala-steward/actions/runs/3746849923/jobs/6362522900
image

### DIFF
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -1,0 +1,19 @@
+name: Link checker
+
+on: push
+
+jobs:
+  link-check:
+    runs-on: ubuntu-latest
+    name: Link checker
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: gaurav-nelson/github-action-markdown-link-check@v1
+        with:
+          config-file: 'mlc_config.json'
+          folder-path: 'docs' # comma separated
+          max-depth: 2
+          check-modified-files-only: ${{ github.ref_name == 'main' && 'no' || 'yes' }}
+          base-branch: main
--- a/.github/workflows/linkcheck.yml
+++ b/.github/workflows/linkcheck.yml
@@ -12,8 +12,8 @@ jobs:
           fetch-depth: 0
       - uses: gaurav-nelson/github-action-markdown-link-check@v1
         with:
-          config-file: 'mlc_config.json'
           folder-path: 'docs' # comma separated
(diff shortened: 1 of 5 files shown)


docs/running.md
```diff
@@ -191,7 +191,6 @@ pipelines:
 There is multiple articles on how to run Scala Steward on-premise:
 
 * [Running Scala Steward On-premise](https://engineering.avast.io/running-scala-steward-on-premise)
-* [Running scala-steward periodically on AWS Fargate](https://medium.com/@tanishiking/running-scala-steward-periodically-on-aws-fargate-3d3d202f0f7)
```
exoego, line 194: This blog appears to be gone.

fthomas:
This is great! Could we also run this on a schedule like once per week or so? It would then notify us about broken links (like those replaced in #2842) in a timely manner without the need that somebody touches these files in a PR.

docs/scalafix-migrations.md
```diff
@@ -13,10 +13,13 @@ Here are two pull requests that demonstrate this feature:
 And here are all pull requests where Scala Steward applied Scalafix
 migrations:
 
+<!-- These links exists but returns 404 often so skip check -->
```
fthomas, line 16: The `pulls` URLs require that the user is logged into GitHub. I guess this is why they are reported as dead.

exoego: Aha, I see

.github/workflows/linkcheck.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  link-check:
+    runs-on: ubuntu-latest
+    name: Link checker
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - uses: gaurav-nelson/github-action-markdown-link-check@v1
```
exoego, line 17: Note) I am waiting https://github.com/gaurav-nelson/github-action-markdown-link-check/pull/168 to be released
Otherwise, the GH action checks not only files in `docs` but also in `module/docs/mdoc/~` which contains vairables in links.

PR #2851, opened 2022-12-22
Add node/yarn to the docker image

AlexITC:
These tools are required to deal with some scala.js projects.

See https://github.com/wiringbits/scala-webapp-template/issues/298

NOTE:
- I have tested that the docker image can be published locally.
- I'm yet to test that this change fixes https://github.com/wiringbits/scala-webapp-template/issues/298
- A potential improvement could be to install nvm instead but that would require to install yarn per repository.


fthomas (approved):
I don't know much about Docker, but this looks good to me.

Issue #2847, opened 2022-12-21
StewardPlugin.scala cannot be loaded by sbt < 1.3.0


fthomas:
Since https://github.com/scala-steward-org/scala-steward/pull/2533 `StewardPlugin.scala` cannot be loaded by sbt versions older than 1.3.0 (i.e sbt without Coursier integration) as can be seen in com-lihaoyi/scalatags which still uses sbt 1.2.8:
```
[info] 2022-12-21 20:06:37,126 INFO  ──────────── Steward com-lihaoyi/scalatags ────────────
[info] 2022-12-21 20:06:37,127 INFO  Check cache of com-lihaoyi/scalatags
[info] 2022-12-21 20:06:38,065 INFO  Clone scala-steward/scalatags
[info] 2022-12-21 20:06:39,254 INFO  Synchronize with com-lihaoyi/scalatags
[info] 2022-12-21 20:06:40,215 INFO  Refresh cache of com-lihaoyi/scalatags
[info] 2022-12-21 20:07:01,284 ERROR Steward com-lihaoyi/scalatags failed
[info] java.io.IOException: 'sbt -Dsbt.color=false -Dsbt.log.noformat=true -Dsbt.supershell=false ;+ stewardDependencies;reload plugins;stewardDependencies' exited with code 1
[info] [info] Loading global plugins from /home/frank/.sbt/1.0/plugins
[info] [info] Updating ProjectRef(uri("file:/home/frank/.sbt/1.0/plugins/"), "global-plugins")...
[info] [info] Done updating.
[info] [info] Compiling 1 Scala source to /home/frank/.sbt/1.0/plugins/target/scala-2.12/sbt-1.0/classes ...
[info] [error] /home/frank/.sbt/1.0/plugins/StewardPlugin.scala:67:15: not found: value csrConfiguration
[info] [error]               csrConfiguration.value.authenticationByRepositoryId.toList
[info] [error]               ^
[info] [error] /home/frank/.sbt/1.0/plugins/StewardPlugin.scala:69:56: value headers is not a member of Any
[info] [error]             (headerKey, headerValue) <- authentication.headers
[info] [error]                                                        ^
[info] [error] /home/frank/.sbt/1.0/plugins/StewardPlugin.scala:70:35: type mismatch;
[info] [error]  found   : Any
[info] [error]  required: String
[info] [error]           } yield Resolver.Header(headerKey, headerValue)
[info] [error]                                   ^
[...]

PR #2852, opened 2022-12-22
Add sbt plugins locally and not globally

fthomas:
`StewardPlugin.scala` is the sbt plugin that outputs all relevant information from a build for consumption by Scala Steward. That plugin has been installed so far globally at the start of Scala Steward into sbt's home directory. With this PR, that plugin is now installed into the local `project/` directories as long as it is needed to run `sbt stewardDependencies`.

This change is motivated by #2847 - i.e. only sbt >= 1.3.0 can currently load our sbt plugin. With this change we would be able to use different plugins dependending on a repo's sbt version.

closes: #1624


exoego (approved):

modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
```diff
@@ -117,9 +112,6 @@ final class SbtAlg[F[_]](config: Config)(implicit
       }
     } yield ()
 
-  private val sbtDir: F[File] =
-    fileAlg.home.map(_ / ".sbt")
```
fthomas, line 121: I should check if FileAlg#home is still needed.

alejandrohdezma (approved):

PR #2859, opened 2022-12-26
Use a version dependent sbt plugin

fthomas:
This PR injects a sbt plugin into a repo's build that depends on the used sbt version. The main motivation of this change is to fix #2847.

The changes here are:
- the `sbt-plugin` project in the build is renamed to `sbt-plugin-1_3_11`. `1_3_11` is used as suffix because the plugin requires at least sbt 1.3.11, see #2858.
- a new `sbt-plugin-1_0_0` is added that contains a `StewardPlugin_1_0_0.scala` that is compatible with sbt 1.0.0 and above. It should also work with sbt 0.13.x but we don't test it here because I don't want to add Scala 2.10 to `crossScalaVersions`.
- `SbtAlg#getDependencies` first reads the sbt version from `build.properties` and then selects the appropriate plugin based on that version.

closes: #2847

PR #2860, opened 2022-12-26
Use sbt plugins from scala-steward-org/sbt-plugin repo

fthomas:
Instead of including the sbt plugins for extracting dependencies in this repo we now download them from https://github.com/scala-steward-org/sbt-plugin and add them as resources to the `core` project as before. This allows to trim the build a little bit and remove 2.12 from `crossScalaVersions` which halves the number of CI builds.


fthomas:
Follow-up to #2859.

mzuehlke (approved):

exoego (approved):
👍

Issue #2541, opened 2022-03-04
Versioning scheme of dependencies is not taken into account


julienrf:
When Scala Steward makes a PR, it decorates it with a set of labels such as “semver-spec-patch”. I believe such labels might be misleading because the actual versioning scheme of the dependency is not taken into account.

For instance, a bump from version `1.2.3` to version `1.3.0` would be labeled as a minor update (`semver-spec-minor`), but it could really be a major update if the library actually uses “package-versioning-policy” (PVP).

That could be an issue if someone wants to setup a mergify rule that automatically merges patch updates, for instance, since there is no proper way to detect what is a patch update (the label `semver-spec-patch` would be set by Scala Steward for a minor update, if the dependency uses PVP).

I believe this issue could be mitigated by taking into account the underlying [versioning scheme](https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme) of the library. The versioning scheme is manually set by the library authors, and there is nothing that forces them to apply the versioning scheme they declared (except some third-party plugins such that sbt-version-policy), so this is why this solution is not perfect, but I do believe that it would be more accurate than the status quo.

A possible way forward would be to add a label “version-scheme-pvp”, “version-scheme-semver-spec”, or “unknown-version-scheme” to the PR’s body without changing the labels that are currently created. Another approach would be to change the existing labels to use the actual versioning scheme (e.g., “semver-spec-patch”, or “pvp-minor”).

Retrieving the underlying version scheme of an artifact is possible with Coursier. It requires querying the `properties` of a `coursier.core.Project` to look for a key `info.versionScheme`.

PR #2863, opened 2022-12-28
Show release notes URL from POM in PR body

fthomas:
This PR shows a release notes URL from the POM in the body of PRs.

Currently, Scala Steward uses the [project.scm.url](https://maven.apache.org/pom.html#scm) or [project.url](https://maven.apache.org/pom.html#More_Project_Information) and tries to find release notes / change logs / version diffs from these URLs by constructing common URLs and checking if they exist. These are then added to the PRs.

With this change, Scala Steward also reads a release notes URL from a [`project.properties.releaseNotesUrl`](https://maven.apache.org/pom.html#Properties) element in the POM and shows it in the PR body. It allows library maintainers to have their release notes anywhere they want them to be and still have Scala Steward show them in PRs. Note that I just made up the name `releaseNotesUrl` for this feature.

This has been partly motivated by https://github.com/com-lihaoyi/requests-scala/pull/122. The changes in `CoursierAlg` also allow to extract more metadata in the future to maybe implement the idea @lefou mentioned in https://github.com/scala-steward-org/scala-steward/issues/2708#issuecomment-1240653195.

/ping @ckipp01 and @lefou since I mentioned this the other day on [Discord](https://discord.com/channels/632150470000902164/940067748103487558/1056877426929573909).

Open questions:
 - Is there prior art of adding release notes or changelog URLs to the POM? Maybe we can reuse some convention of adding URLs to the POM that other projects already use.
 - If not, is it okay to use `releaseNotesUrl` or should the element name be prefixed with `scalasteward.` or something like that?

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
@@ -16,15 +16,14 @@
 
 package org.scalasteward.core.coursier
 
+import cats.Parallel
 import cats.effect._
 import cats.implicits._
-import cats.{Applicative, Parallel}
 import coursier.cache.{CachePolicy, FileCache}
 import coursier.core.{Authentication, Project}
-import coursier.{Fetch, Info, Module, ModuleName, Organization}
-import org.http4s.Uri
+import coursier.{Fetch, Module, ModuleName, Organization}
 import org.scalasteward.core.data.Resolver.Credentials
-import org.scalasteward.core.data.{Dependency, Resolver, Scope, Version}
+import org.scalasteward.core.data.{Dependency, Resolver, Version}
 import org.scalasteward.core.util.uri
 import org.typelevel.log4cats.Logger
 
@@ -32,16 +31,9 @@ import org.typelevel.log4cats.Logger
   * metadata.
   */
 trait CoursierAlg[F[_]] {
-  def getArtifactUrl(dependency: Scope.Dependency): F[Option[Uri]]
+  def getMetadata(dependency: Dependency, resolvers: List[Resolver]): F[Option[DependencyMetadata]]
 
   def getVersions(dependency: Dependency, resolver: Resolver): F[List[Version]]
-
-  final def getArtifactIdUrlMapping(dependencies: Scope.Dependencies)(implicit
-      F: Applicative[F]
-  ): F[Map[String, Uri]] =
-    dependencies.sequence
-      .traverseFilter(dep => getArtifactUrl(dep).map(_.map(dep.value.artifactId.name -> _)))
-      .map(_.toMap)
 }
 
 object CoursierAlg {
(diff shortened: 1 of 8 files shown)


ckipp01:
> Is there prior art of adding release notes or changelog URLs to the POM? Maybe we can reuse some convention of adding URLs to the POM that other projects already use.

I dug around a bit to see if I could find anything and the only thing I found is that some release plugins will use this same convention. For example you can see this in a couple apache Maven projects that use `properties.commons.releaseNotesLocation`. Here are a couple examples:

- https://github.com/apache/commons-vfs/blob/61e51fbf806da05d82b3207cf4d5ddd6dfa1136a/pom.xml#L169
- https://github.com/apache/commons-math/blob/585b04c22edee12b193fc3e5099537af4127080b/dist-archive/pom.xml#L40

> If not, is it okay to use releaseNotesUrl or should the element name be prefixed with scalasteward. or something like that?

I like `releaseNotesUrl`. It's clear. I actually don't think we should prefix it with `scalasteward`. This might be a good opportunity to introduce something that various tools in theory could use. So I'm in favor of leaving in generic.

Thanks for diving into this @fthomas!

fthomas:
> I like `releaseNotesUrl`. It's clear. I actually don't think we should prefix it with `scalasteward`. This might be a good opportunity to introduce something that various tools in theory could use. So I'm in favor of leaving in generic.

I agree with that and also think that it could be less likely that people adopt that property if it is tied to Scala Steward. At least I wouldn't want to sprinkle my POMs with tool specific properties.

fthomas:
Btw, I'd like to note that the idea of adding a release notes URL to the POM has been discussed before in the [
"Coordinate compiler and tooling releases?"](https://contributors.scala-lang.org/t/coordinate-compiler-and-tooling-releases/5312) thread on the Scala Contributors forum although it was discussed for the purpose of delaying PRs until a library is announced.

fthomas:
Just noticed that with this change #2541 proposed by @julienrf should be almost trivial to implement by adding a new `versionScheme` field to `DependencyMetadata` and using that in `NewPullRequestData`.

modules/core/src/test/scala/org/scalasteward/core/nurture/NurtureAlgTest.scala
```diff
@@ -0,0 +1,73 @@
...
+import org.scalasteward.core.TestSyntax._
+import org.scalasteward.core.data.{DependencyInfo, RepoData, UpdateData}
+import org.scalasteward.core.edit.EditAttempt.UpdateEdit
+import org.scalasteward.core.git.{Branch, Commit}
+import org.scalasteward.core.mock.MockContext.context.nurtureAlg
+import org.scalasteward.core.mock.{MockEff, MockState}
+import org.scalasteward.core.repoconfig.RepoConfig
+import org.scalasteward.core.vcs.data.{NewPullRequestData, Repo}
+
+class NurtureAlgTest extends CatsEffectSuite with Http4sDsl[MockEff] {
```
mzuehlke, line 16: I always bothered me that the heart of steward the `NurtureAlg` was not tested at all. Well done.

fthomas: We're slowly getting there. Unfortunately there is a lot of stuff to set up until we can meaningfully test `NurtureAlg`.

fthomas:
I just looked at the POM of [Cats 2.9.0](https://repo1.maven.org/maven2/org/typelevel/cats-core_2.13/2.9.0/cats-core_2.13-2.9.0.pom). It contains these properties:
```xml
<properties>
  <info.apiURL>https://www.javadoc.io/doc/org.typelevel/cats-docs_2.13/2.9.0/</info.apiURL>
  <info.versionScheme>early-semver</info.versionScheme>
</properties>
```
Both come from the corresponding sbt settings [`apiURL`](https://www.scala-sbt.org/1.x/docs/Howto-Scaladoc.html#Define+the+location+of+API+documentation+for+a+library) and [`versionScheme`](https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme). I don't know if there is something special about the `info` prefix but sbt is using that for the settings it writes into the POM. I'm now tempted to imitate sbt here and to require the `info` prefix for `releaseNotesUrl`.

fthomas:
The last commit changed the key to `info.releaseNotesUrl`. Matching the key is done case-insensitive, so it will also match `info.releaseNotesURL` or any other capitalisation. The added `NurtureAlgTest` demonstrates that URLs are still displayed correctly for a simple update. As far as I'm concerned this is ready to be merged.

ckipp01:
> The last commit changed the key to `info.releaseNotesUrl`. Matching the key is done case-insensitive, so it will also match `info.releaseNotesURL` or any other capitalisation. The added `NurtureAlgTest` demonstrates that URLs are still displayed correctly for a simple update. As far as I'm concerned this is ready to be merged.

Awesome.  I don't want to steal your thunder so I didn't do it, but you might want to announce this new feature in https://contributors.scala-lang.org/ so that more people are aware of it. Basically all over the place, ha.

alejandrohdezma (approved):
👏🏼 👏🏼

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
```diff
@@ -50,64 +42,72 @@ object CoursierAlg {
...
+        }
 
-      private def getArtifactUrlImpl(
+      private def getMetadataImpl(
           dependency: coursier.Dependency,
-          repositories: List[coursier.Repository]
-      ): F[Option[Uri]] = {
+          repositories: List[coursier.Repository],
+          acc: Option[DependencyMetadata]
+      ): F[Option[DependencyMetadata]] = {
```
mzuehlke, line 66: just an idea:
you could simplify the `acc` to `acc: DependencyMetadata = DependencyMetadata(None,None,None)`
And the return type would be `F[DependencyMetadata]`

The further down the `fold` gets simpler, too:
instead of 
```Scala
val metadata = {
    val current = metadataFrom(project)
    acc.fold(current)(_.enrichWith(current))
}
```

you could write:
```Scala
val metadata = acc.enrichWith(metadataFrom(project))
```

mzuehlke (approved):
Would you update the reference to the new property in the PR description, there you are still referring to `project.properties.releaseNotesUrl`

modules/core/src/main/scala/org/scalasteward/core/coursier/CoursierAlg.scala
fthomas: That is a good suggestion. We currently do not care about the difference of an empty `Option[DependencyMetadata]` and an empty `DependencyMetadata`. And dealing with the latter is easier in a few places.

fthomas: Done in [7e0bec6](https://github.com/scala-steward-org/scala-steward/pull/2863/commits/7e0bec6a3cb42f5af754f6a1b2e9ceb1e93c9464).

fthomas:
Btw, in sbt the `info.releaseNotesUrl` property can be added to the POM vie [`pomPostProcess`](https://www.scala-sbt.org/1.x/docs/Publishing.html#Modifying+the+generated+POM):
```sbt
pomPostProcess := { (node: xml.Node) =>
  val releaseNotesUrl =
    <info.releaseNotesUrl>https://www.scala-lang.org/blog/2022/11/07/scala-3.2.1-released.html</info.releaseNotesUrl>
  var updated = false
  val updatedChild = node.child.map {
    case e: xml.Elem if e.label == "properties" =>
      updated = true; e.copy(child = e.child :+ releaseNotesUrl)
    case n => n
  }
  node match {
    case e: xml.Elem if updated => e.copy(child = updatedChild)
    case e: xml.Elem => e.copy(child = e.child :+ <properties>{releaseNotesUrl}</properties>)
    case n           => n
  }
}
```
This adds the `info.releaseNotesUrl` to an existing `properties` element or adds the `properties` element with a sole `info.releaseNotesUrl` element if the former does not exist.

Not sure if there is an easier way to do it. It would be certainly nice if this wouldn't be so complicated.

lefou:
In Mill, you can add the property as follows:

```scala
// in object foo extends PublishModule
  def publishProperties = super.publishProperties() ++ Map(
    "info.releaseNotesURL" -> https://www.scala-lang.org/blog/2022/11/07/scala-3.2.1-released.html
  )

```

fthomas:
Awesome, thanks @julienrf!

fthomas:
Just announced this on Scala Contributors: https://contributors.scala-lang.org/t/add-release-notes-urls-to-your-poms/6059

PR #2864, opened 2022-12-30
Optimize regexes in VersionPositionScanner

fthomas:
I played with using `VersionPositionScanner` instead of `GitAlg#findFilesContaining` in `RepoCacheAlg` and noticed that the regex used in `findMillDependency` is quite slow since it allows whitespace (including newlines) in the groupId and artifactId identifiers.

This change disallows whitespace in those identifiers and requires that they are not empty. It increases performance significantly even with the toy example in the added benchmark.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/update/VersionPositionScanner.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/update/VersionPositionScanner.scala
@@ -44,8 +44,9 @@ object VersionPositionScanner {
     }
 
   private def sbtModuleIdRegex(version: Version): Regex = {
+    val ident = """[^\s]+"""
     val v = Regex.quote(version.value)
-    raw"""(.*)"(.*)"\s*%+\s*"(.*)"\s*%+\s*"$v"""".r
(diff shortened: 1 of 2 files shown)


modules/core/src/main/scala/org/scalasteward/core/edit/update/VersionPositionScanner.scala
```diff
@@ -44,8 +44,9 @@ object VersionPositionScanner {
     }
 
   private def sbtModuleIdRegex(version: Version): Regex = {
+    val ident = """[^\s]+"""
```
mzuehlke, line 47: I locally tried to simplify
```suggestion
    val ident = """\S+"""
```
but it is slower

mzuehlke (approved):

modules/core/src/main/scala/org/scalasteward/core/edit/update/VersionPositionScanner.scala
fthomas: Oh, that's interesting and strange given that `[^\s]` looks like the exact definition of `\S`.

mzuehlke: Some more runs lead to no measurable difference. Which is what to expect 😄

PR #2875, opened 2023-01-03
Use `clientResponses` from `MockState` in vcs tests

mzuehlke:
* Use the newly added #2872  `clientResponses` field in `MockState` throughout the test in vcs.
* Construct the different vcsAlg instances using `VCSSelection`
* add tests for the `Authorization` headers
* convert tests to `CatsEffectSuite`

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
@@ -62,6 +62,7 @@ final class Context[F[_]](implicit
     val fileAlg: FileAlg[F],
     val filterAlg: FilterAlg[F],
     val gitAlg: GitAlg[F],
+    val httpJsonClient: HttpJsonClient[F],
     val hookExecutor: HookExecutor[F],
     val logger: Logger[F],
     val mavenAlg: MavenAlg[F],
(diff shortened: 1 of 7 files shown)


fthomas (approved):
That is an impressive PR!

> * Use the newly added Add pre-determined HTTP responses to MockState #2872 clientResponses field in MockState throughout the test in vcs.

It looks like this was mostly a straightforward change, or did you encounter any quirks with using `clientResponses` ? I wasn't sure how well this would work in all other tests.

mzuehlke:
The only thing that was really tricky was: https://github.com/scala-steward-org/scala-steward/blob/vcs-mockstate/modules/core/src/test/scala/org/scalasteward/core/vcs/gitlab/GitLabApiAlgTest.scala#L241-L249
in `GitLabApiAlgTest`  "createPullRequest -- don't fail on error code"
locally extending the `HttpApp` was not that simple. Especially silencing the `incomplete match error`.

fthomas:
> locally extending the `HttpApp` was not that simple

Yes, that looks tricky indeed. Maybe a method on `MockState` would have been handy:
* `def addResponses(httpApp: HttpApp[MockEff]): MockState` or maybe
* `def addResponses(httpRoutes: HttpRoutes[MockEff]): MockState` since `HttpRoutes` can be created from a partial function.

PR #2878, opened 2023-01-04
Log the update when scala-steward is unable to bump a version

alexklibisz:
Sometimes scala-steward fails to bump a version.

When this happens, the logs just show:

```
WARN Unable to bump version
```

This PR adds the name of the update to that warning. This will look something like:

```
WARN Unable to bump version for update com.foo:library : 1.2.3 -> 1.3.0
```


alejandrohdezma (approved):
Great improvement! Thanks!

PR #2881, opened 2023-01-04
Rename `WorkspaceAlg#cleanWorkspace`

fthomas:
This renames `WorkspaceAlg#cleanWorkspace` to `cleanReposDir` since this is what this method actually does. The log message is also changed from "Clean workspace" to "Clean $reposDir". The former message is misleading since the workspace needs to be persisted between runs for certain features to work, but Scala Steward says on every run that it is cleaning this workspace.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
@@ -24,7 +24,7 @@ import org.scalasteward.core.vcs.data.{BuildRoot, Repo}
 import org.typelevel.log4cats.Logger
 
 trait WorkspaceAlg[F[_]] {
-  def cleanWorkspace: F[Unit]
+  def cleanReposDir: F[Unit]
 
   def rootDir: F[File]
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
```diff
@@ -40,24 +40,22 @@ object WorkspaceAlg {
       F: FlatMap[F]
   ): WorkspaceAlg[F] =
     new WorkspaceAlg[F] {
-      private[this] val reposDir = config.workspace / "repos"
+      private val reposDir: File =
+        config.workspace / "repos"
 
-      private[this] def repoDirUnsafe(repo: Repo) = reposDir / repo.owner / repo.repo
+      private def mkRepoDir(repo: Repo): File =
```
mzuehlke, line 46: only a comment about naming :-)
I first thought that `mkRepoDir` creates the directory.
Suggestion:
* `val reposDir` => `val reposRootDir`
* `def mkRepoDir(repo: Repo)` => `def repoDir(repo: Repo)`

If I got the meaning right

mzuehlke (approved):

modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
fthomas: > I first thought that mkRepoDir creates the directory.

Right, this is a bad name. But we can't use `repoDir` since this already exists. I tend to rename `mkRepoDir` to `toFile`.

mzuehlke: looks fine, too

Issue #2877, opened 2023-01-03
"Unable to bump version" when version is defined as both a string and a val


alexklibisz:
The great work in #2816 resolved most of the "Unable to bump version" errors that I've seen on internal projects. However, I did find one new one.

The build setup looks like this:

```scala
val FooLibrariesVersion = "1.2.3"

libraryDependencies ++= Seq(
 "com.foo" %% "library-1" % FooLibrariesVersion,
 "com.foo" %% "library-2" % FooLibrariesVersion,
 "com.foo" %% "library-3" % FooLibrariesVersion,
 "com.foo" %% "library-4" % "0.1.2"
)
```

The latest version of all four dependencies is "2.0.0".

Here's what scala-steward does:

1. recognizes that `"library-1", "library-2", "library-3", "library-4"` all need to be bumped to `"2.0.0"` as well ("Found 4 updates")
2. fails to bump `"library-1", "library-2", "library-3"` ("Unable to bump version")
3. correctly bumps `"library-4"` to `"2.0.0"`

I've changed the library names and versions, but everything should still map to reality.

I suspect that consolidating the versions to a single val will fix things, but this still seems like it's worth reporting.

Can someone point me to a way to test out this specific scenario w/ the current heuristics implementation? Maybe I can take a pass at fixing this.

fthomas:
> [...] but this still seems like it's worth reporting.

Definitely, thanks for reporting! 

> Can someone point me to a way to test out this specific scenario w/ the current heuristics implementation? Maybe I can take a pass at fixing this.

The first step would be to add a reproduction to [`RewriteTest`](https://github.com/scala-steward-org/scala-steward/blob/bdd73b888d26b36373abd5bf868e32fd4b9d7e0e/modules/core/src/test/scala/org/scalasteward/core/edit/RewriteTest.scala). Once we have that, we can provide better guidance how to fix it. Feel free to open a PR with just the reproduction.

mzuehlke:
You can add a "failing" test to the test suite: https://scalameta.org/munit/docs/tests.html#declare-test-that-should-always-fail

fthomas:
Fixed by #2884.

PR #2884, opened 2023-01-05
Issue #2877: Fix "Unable to bump version" when using a val version in libraryDependencies ++= Seq(...)

alexklibisz:
This is a fix and tests for the issue I raised in #2877.

Prior to this fix, scala-steward was failing to properly update this example:

```scala
val ScalaTestVersion = "3.2.13"
libraryDependencies ++= Seq(
  "org.scalatest" %% "scalatest" % ScalaTestVersion,
  "org.scalatest" %% "scalactic" % "3.2.13"
)
````

due to the trailing comma in `ScalaTestVersion,`.

### DIFF
--- a/modules/core/src/test/scala/org/scalasteward/core/edit/RewriteTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/edit/RewriteTest.scala
@@ -840,7 +840,7 @@ class RewriteTest extends FunSuite {
     runApplyUpdate(update, original, expected)
   }
 
-  test("issue-2877: sbt using same version in a val and a literal using a Seq addition".fail) {
+  test("issue-2877: sbt using same version in a val and a literal using a Seq addition") {
     val update = ("org.scalatest".g % Nel.of(
       "scalatest".a,
       "scalactic".a
--- a/modules/core/src/test/scala/org/scalasteward/core/edit/RewriteTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/edit/RewriteTest.scala
@@ -840,7 +840,7 @@ class RewriteTest extends FunSuite {
     runApplyUpdate(update, original, expected)
   }
 
-  test("issue-2877: sbt using same version in a val and a literal".fail) {
+  test("issue-2877: sbt using same version in a val and a literal using a Seq addition".fail) {
     val update = ("org.scalatest".g % Nel.of(
       "scalatest".a,
       "scalactic".a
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: modules/core/src/test/scala/org/scalasteward/core/edit/RewriteTest.scala [hunk 1]
```diff
@@ -840,6 +840,31 @@ class RewriteTest extends FunSuite {
...
+          |)
+          |""".stripMargin
+    )
+    val expected = Map(
+      "build.sbt" ->
+        """
+          |val ScalaTestVersion = "3.2.14"
+          |libraryDependencies ++= Seq(
+          |  "org.scalatest" %% "scalatest" % ScalaTestVersion,
+          |  "org.scalatest" %% "scalactic" % "3.2.14"
```
fthomas, line 861: Why should the scalactic version be bumped if the update only contains scalatest? I'd say it is correct to not bump the scalactic version in this case and Scala Steward is already doing the right thing.

Also this looks like Scala Steward bumped the version of the `val` but you mentioned in #2877 that it is unable to so. Something is missing here.

alexklibisz: You're right, I messed up my own example while translating it to code :)

I pushed up a fix in  https://github.com/scala-steward-org/scala-steward/pull/2884/commits/102a8e47de8de36c70c8ce6c1622bfaf53dc6808.

I'll update the PR description to match that fix.

fthomas: That looks much better, thanks! :-) My guess is that the comma in `ScalaTestVersion,` is tripping Scala Steward up. You could try removing it and see if that fixes the test. If you want to dig into the code, I think that [`scalaValInDependencyDefPositions`](https://github.com/scala-steward-org/scala-steward/blob/12362b512ad12bab690f2d12554595bc84ef1166/modules/core/src/main/scala/org/scalasteward/core/edit/update/Selector.scala#L46) should have returned a `VersionPosition`  of `val ScalaTestVersion = "3.2.13"` but it returned an empty list instead.

fthomas: Btw, if it is the comma that is causing trouble, the issue can be fixed by removing trailing commas in https://github.com/scala-steward-org/scala-steward/blob/89c66e36a42425857250b17be0b1be55aa0fad02/modules/core/src/main/scala/org/scalasteward/core/edit/update/data/ModulePosition.scala#L31-L32

mzuehlke: Gotcha:
```Scala
  def unwrappedVersion: String =
    version.value.dropWhile(Set('"', '$', '{')).reverse.dropWhile(Set('"', '}', ',')).reverse
```
Fixed the issue 👍

alexklibisz: Ah, yep, it's the comma. I added a test case that uses 2x `libraryDependencies +=` instead of 1x `libraryDependencies ++=`, and the 2x case passes: https://github.com/scala-steward-org/scala-steward/pull/2884/commits/102a8e47de8de36c70c8ce6c1622bfaf53dc6808

alexklibisz:
Maybe we can merge this and then @mzuehlke can follow-up w/ the fix? Or I can add the fix to this branch? LMK, or feel free to merge.

mzuehlke:
> Maybe we can merge this and then @mzuehlke can follow-up w/ the fix? Or I can add the fix to this branch? LMK, or feel free to merge.

Feel free to just apply the my online fix in your branch

alexklibisz:
> > Maybe we can merge this and then @mzuehlke can follow-up w/ the fix? Or I can add the fix to this branch? LMK, or feel free to merge.
> 
> Feel free to just apply the my online fix in your branch

Will do now, thanks!

alexklibisz:
Cool, fixed the issue in the last commit and updated the PR description. Should be ready to merge.

fthomas (approved):
Many thanks, @alexklibisz!

mzuehlke (approved):

PR #2888, opened 2023-01-05
Replacing deprecated http4s-okhttp-client with http4s-jdk-http-client

TonioGela:
As per title, I swapped the client implementation.
There are a few things to note by the way:
- As you can see [here](https://github.com/http4s/http4s-jdk-http-client/blob/main/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala#L257) JdkHttpClient doesn't expose a builder that let you operate over the `java.net.HttpClient.Builder` so I had to copy some of the default logic from the lib (this might be a nice addition to the upstream lib btw)
- the default `BuilderMiddleware` is no more an identity as the default for HttpClient is to do not follow redirects
- There's no way to "destroy" the HttpClient instance so the Resource.release is a F.unit

Closes https://github.com/scala-steward-org/scala-steward/issues/2887

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/client/ClientConfiguration.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/client/ClientConfiguration.scala
@@ -41,7 +41,7 @@ object ClientConfiguration {
     Resource
       .eval(defaultHttpClient[F](bmw).map(JdkHttpClient[F](_)))
       .map(cmw)
-
+  // Copied from https://github.com/http4s/http4s-jdk-http-client/blob/b9655b90549319fbe999069e7c95ab1752efecb9/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala#L257-L275 to support BuilderMiddleware
   private def defaultHttpClient[F[_]: Async](bmw: BuilderMiddleware): F[HttpClient] =
     Async[F].executor.flatMap { exec =>
       Async[F].delay {
--- a/modules/core/src/main/scala/org/scalasteward/core/client/ClientConfiguration.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/client/ClientConfiguration.scala
@@ -17,27 +17,49 @@
 package org.scalasteward.core.client
 
 import cats.effect._
+import cats.syntax.all._
 import eu.timepit.refined.auto._
 import eu.timepit.refined.types.numeric.PosInt
 import org.http4s.Response
 import org.http4s.client._
 import org.http4s.headers.`User-Agent`
-import org.http4s.okhttp.client.OkHttpBuilder
+import org.http4s.jdkhttpclient.JdkHttpClient
+import java.net.http.HttpClient
+import java.net.http.HttpClient.Builder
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/org/scalasteward/core/client/ClientConfiguration.scala
```diff
@@ -17,28 +17,50 @@
...
   }
-  def build[F[_]: Async](bmw: BuilderMiddleware[F], cmw: Middleware[F]): Resource[F, Client[F]] =
-    OkHttpBuilder
-      .withDefaultClient[F]
-      .map(bmw)
-      .flatMap(_.resource)
+
+  def build[F[_]: Async](bmw: BuilderMiddleware, cmw: Middleware[F]): Resource[F, Client[F]] =
+    Resource
+      .make(defaultHttpClient[F](bmw).map(JdkHttpClient[F](_)))(_ => Async[F].unit)
```
armanbilge, line 42:
```suggestion
    Resource
      .eval(defaultHttpClient[F](bmw).map(JdkHttpClient[F](_)))
```

fthomas:
Thank you for the PR, @TonioGela! Just checked this behind our corporate proxy and it works! :tada:

modules/core/src/main/scala/org/scalasteward/core/client/ClientConfiguration.scala
```diff
@@ -17,28 +17,50 @@
...
-  def build[F[_]: Async](bmw: BuilderMiddleware[F], cmw: Middleware[F]): Resource[F, Client[F]] =
-    OkHttpBuilder
-      .withDefaultClient[F]
-      .map(bmw)
-      .flatMap(_.resource)
+
+  def build[F[_]: Async](bmw: BuilderMiddleware, cmw: Middleware[F]): Resource[F, Client[F]] =
+    Resource
+      .eval(defaultHttpClient[F](bmw).map(JdkHttpClient[F](_)))
       .map(cmw)
```
fthomas, line 44:
```suggestion
  // Copied from https://github.com/http4s/http4s-jdk-http-client/blob/b9655b90549319fbe999069e7c95ab1752efecb9/core/src/main/scala/org/http4s/jdkhttpclient/JdkHttpClient.scala#L257-L275 to support BuilderMiddleware
```

fthomas (approved):
> so I had to copy some of the default logic from the lib (this might be a nice addition to the upstream lib btw)

I agree, it would be nice if we wouldn't need a copy of this http4s-jdk-http-client function here, but this is not a blocker for me.

TonioGela:
> Thank you for the PR, @TonioGela! Just checked this behind our corporate proxy and it works! :tada: 

So glad that it worked!

> I agree, it would be nice if we wouldn't need a copy of this http4s-jdk-http-client function here, but this is not a blocker for me.

I promess I'll open a PR upstream with a refactor proposal in order to expose the builder. If it gets accepted, once released, I'll update this implementation 😇

alejandrohdezma (approved):

PR #2891, opened 2023-01-06
Rename VCSExtraAlg to UpdateInfoUrlFinder

fthomas:
This is a purely stylistic change. No behaviour has been changed.

It
 - renames `VCSExtraAlg` to `UpdateInfoUrlFinder`
 - renames `ReleaseRelatedUrl` to `UpdateInfoUrl`
 - moves `UpdateInfoUrlFinder` and `UpdateInfoUrl` to the nurture package since it is mainly used there
 - moves the functions from the `vcs` package object that created `ReleaseRelatedUrl`s into the `UpdateUrlFinder` companion since these functions are more related to the `UpdateUrlFinder` than to the different VCS integrations. For example, it would be perfectly fine to have a VCS intgeration without updating the `UpdateInfoUrlFinder` or to update it for VCSs for which we don't have integrations.

I hope the new names and the reorganization makes it clearer what `UpdateInfoUrlFinder` is about.


mzuehlke (approved):
Better naming 👍

exoego (approved):

Issue #1974, opened 2021-02-27
Artifact migrations do not work for group updates


fthomas:
Which is not surprising because only `Update.Single` has the [`newerGroupId` and `newerArtifactId`](https://github.com/scala-steward-org/scala-steward/blob/299cdfbdeaa6099912d0085b6a26a710ce5ddf7e/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala#L69-L70) fields.

PR #2898, opened 2023-01-07
Support artifact migrations in `Update.ForGroupId`

fthomas:
This supports artifact migrations of multiple artifacts in one update. Prior to this change, this was not possible because the `newerGroupId` and `newerArtifactId` fields were discarded while `ForArtifactId` updates were grouped into `ForGroupId`. With this change, `ForGroupId` is just a non-empty list of `ForArtifactId`s, so that these fields are not lost.

Closes: #1974

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala
@@ -18,11 +18,9 @@ package org.scalasteward.core.data
 
 import cats.Order
 import cats.implicits._
-import io.circe.Decoder
-import io.circe.Encoder
-import io.circe.Json
 import io.circe.generic.semiauto._
 import io.circe.syntax._
+import io.circe.{Decoder, Encoder, HCursor, Json}
 import org.scalasteward.core.repoconfig.PullRequestGroup
 import org.scalasteward.core.util
 import org.scalasteward.core.util.Nel
@@ -52,6 +50,7 @@ object Update {
   }
 
   sealed trait Single extends Product with Serializable with Update {
+    def forArtifactIds: Nel[ForArtifactId]
     def crossDependencies: Nel[CrossDependency]
     def dependencies: Nel[Dependency]
     def groupId: GroupId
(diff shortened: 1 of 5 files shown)


mzuehlke (approved):

exoego (approved):

alejandrohdezma (approved):
This is an awesome change, it also simplifies a lot understanding the update hierarchy!

fthomas:
> it also simplifies a lot understanding the update hierarchy!

I agree. Maybe we can simplify it even further in the future by removing `CrossDependency` and adding a new `Update.ForCrossArtifactId`. I could imagine that it looks like this: 
- if we find an update for a `Dependency` (for example `cats-core_2.12`), we wrap it into `ForCrossArtifactId`
- multiple `ForCrossArtifactId`s are grouped into `ForArtifactId` if they have the same groupId, the same newer versions, the same artifactId name, but different artifactId cross names (for example `cats-core_2.12` and `cats-core_2.13`)
- multiple `ForArtifactId` are grouped into `ForGroupId` if they have the same groupId, the same newer versions, but different artifactIds (as it is today)

IMO `CrossDependency` is a bit strange today, since we artificially wrap only one `Dependency` in https://github.com/scala-steward-org/scala-steward/blob/028a0e1cbc0a1c03f7fc9a461e462fb5cfc3148a/modules/core/src/main/scala/org/scalasteward/core/update/UpdateAlg.scala#L59 and then later extend the `ForArtifactId.crossDependency` in https://github.com/scala-steward-org/scala-steward/blob/028a0e1cbc0a1c03f7fc9a461e462fb5cfc3148a/modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala#L104

Maybe it becomes simpler if we just use distinct types in these places.

mzuehlke:
> Maybe it becomes simpler if we just use distinct types in these places.

Your proposal sounds really promising 👍

fthomas:
> Maybe it becomes simpler if we just use distinct types in these places.

Ok, I tried this now twice and it always becomes ugly in `PruningAlg` and non-trivial to get rid of `CrossDependency` there. This is probably an indication that `PruningAlg` is itself a mess that would benefit from some clean-up.

PR #2908, opened 2023-01-10
sbt-site is changing orgs

SethTisue:
fyi @ennru (note that Frank assures me it's not too early to do this, even though there isn't a 1.5.0 final release yet)

@fthomas thanks for the suggestion on Mastodon

### DIFF
--- a/modules/core/src/main/resources/artifact-migrations.v2.conf
+++ b/modules/core/src/main/resources/artifact-migrations.v2.conf
@@ -127,6 +127,11 @@ changes = [
   },
   {
     groupIdBefore = com.typesafe.sbt
+    groupIdAfter = com.github.sbt
+    artifactIdAfter = sbt-site
+  },
+  {
+    groupIdBefore = com.typesafe.sbt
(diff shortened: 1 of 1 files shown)


exoego (approved):
Context: https://github.com/sbt/sbt-site/pull/193

mzuehlke (approved):

fthomas:
> note that Frank assures me it's not too early to do this, even though there isn't a 1.5.0 final release yet

Let me repeat what I said on Mastodon: Scala Steward does not propose updates from a stable version to a pre-release version and adding an artifact migration doesn't change this. So we won't see updates from com.typesafe.sbt:sbt-site 1.4.1 to com.github.sbt:sbt-site 1.5.0-RC1. Only 1.5.0 or a later stable version is considered a suitable next version from 1.4.1. So Scala Steward will only start proposing updates and change the groupId, once 1.5.0 is released.

fthomas:
Btw, this was the 100th artifact migration added to Scala Steward! :tada:

PR #2911, opened 2023-01-11
Use ember server for tests

armanbilge:
Since blaze is schismed, it may lag on the milestone releases, but ember is in sync with core.


fthomas (approved):
Perfect, thanks @armanbilge!

Issue #2912, opened 2023-01-11
Replace the VCS prefix?


fthomas:
Scala Steward currently works with these systems where Git repositores are hosted and PRs are created: GitHub, GitLab, Bitbucket, Bitbucket Server, and Azure Repos. In our code these are called a VCS which is an initialism for version control system. I'd like to change the VCS prefix in Scala Steward since I think there is mismatch between what is commonly understood as a VCS and for what it is used in Scala Steward. The Wikipedia page for VCS refers to https://en.wikipedia.org/wiki/Version_control which talks about systems like Git, CVS, Subversion, Mercurial, Bazaar etc. of which we only support Git. So a VCS there is the software that does the version control but not necessarily the software that hosts the repositories or supports pull requests.

The Wikipedia also has this [comparison of source-code-hosting facilities](https://en.wikipedia.org/wiki/Comparison_of_source-code-hosting_facilities) which lists all the above hosting providers that Scala Steward supports and refers to them as [forge](https://en.wikipedia.org/wiki/Forge_(software)). I like the name forge as replacement for VCS in Scala Steward since it is succinct and more accurate.

Any objections to replacing `VCS` with `Forge` or other suggestions for a better name that is less misleading?

mzuehlke:
Sounds like a good idea

PR #2916, opened 2023-01-13
Replace VCS with Forge

fthomas:
This replaces `VCS` with `Forge`, see #2912. The old command-line options `--vcs-type`, `--vcs-api-host`, and `--vcs-login` are retained so that users don't have to switch to the new options immediately.


mzuehlke (approved):
I only found a single leftover
https://github.com/scala-steward-org/scala-steward/blob/topic/forge/build.sbt#L401

alejandrohdezma (approved):

exoego (approved):

fthomas:
> I only found a single leftover https://github.com/scala-steward-org/scala-steward/blob/topic/forge/build.sbt#L401

Good catch, thanks!

Issue #1468, opened 2020-06-04
Add release notes/changelog/version diff for dependencies in private GitHub repos


MPV:
These nice features don't work for dependencies whose git repos are private (thus need auth):
- Release Notes
- Changelog
- Version Diff

This is probably because the http client used for those doesn't seem to use the github token:

https://github.com/fthomas/scala-steward/blob/master/modules/core/src/main/scala/org/scalasteward/core/util/HttpExistenceClient.scala

fthomas:
`HttpExistenceClient` has been renamed to [`UrlChecker`](https://github.com/scala-steward-org/scala-steward/blob/0315d316f753696948befa2db4aff03f52e1c2d6/modules/core/src/main/scala/org/scalasteward/core/util/UrlChecker.scala). I think the `Client` used by `UrlChecker` needs do authenticated calls for anything that goes to `VCSCfg#apiHost`. I once tried this in this old branch (https://github.com/scala-steward-org/scala-steward/compare/topic/VCSClient) but then ran out of steam.

fthomas:
There is now a PR at #2920 that fixes this if someone wants to give it a try.

laughedelic:
Great news! Thanks!

Can I try this version with the [scala-steward-action](https://github.com/scala-steward-org/scala-steward-action)?

fthomas:
Yes, just set the `scala-steward-version` option of the action to `0.20.0-61-1f0282a6-SNAPSHOT` for example. Available snapshot versions are here: https://oss.sonatype.org/content/repositories/snapshots/org/scala-steward/scala-steward-core_2.13/

lenguyenthanh:
This currently doesn't work for private github repositories. because normal url like `https://`github.com/org/private-repository` will not accept any token. So both basic auth and jwt token don't work.

I'm not sure this is a regression or it has never worked 🤔 

Can anyone confirm?

laughedelic:
In my experience it never worked

PR #2919, opened 2023-01-14
Do forge authentication in one function

fthomas:
This adds `ForgeSelection.authenticate` which modifies requests for authentication for a given forge type. This is nicer than to scatter this logic over multiple files and I think it will also be useful in fixing #1468.


exoego (approved):

modules/core/src/main/scala/org/scalasteward/core/forge/ForgeSelection.scala
```diff
@@ -30,57 +32,43 @@ import org.scalasteward.core.util.HttpJsonClient
...
+      case Bitbucket       => _.putHeaders(basicAuth(user)).pure[F]
+      case BitbucketServer =>
+        // Bypass the server-side XSRF check, see
+        // https://github.com/scala-steward-org/scala-steward/pull/1863#issuecomment-754538364
+        val xAtlassianToken = Header.Raw(ci"X-Atlassian-Token", "no-check")
+        _.putHeaders(basicAuth(user), xAtlassianToken).pure[F]
+      case GitHub => _.putHeaders(basicAuth(user)).pure[F]
+      case GitLab => _.putHeaders(Header.Raw(ci"Private-Token", user.accessToken)).pure[F]
     }
+
```
mzuehlke, line 71: what about:
```Scala
case BitbucketServer => _.putHeaders(basicAuth(user), xAtlassianToken).pure[F]

...

// Bypass the server-side XSRF check, see
// https://github.com/scala-steward-org/scala-steward/pull/1863#issuecomment-754538364
private val xAtlassianToken = Header.Raw(ci"X-Atlassian-Token", "no-check")
```

to make the patter matching simpler

mzuehlke: And I like the overall code reduction 😄 👍

fthomas: Good suggestion! That nicely aligns all cases.

fthomas: Done in https://github.com/scala-steward-org/scala-steward/pull/2920/commits/da4ec126317ab27885cd7dfb0e32f8b0605241c5.

PR #2920, opened 2023-01-14
Authenticate UrlChecker calls to API host

### LINKED ISSUE #1468: Add release notes/changelog/version diff for dependencies in private GitHub repos

fthomas:
This authenticates `UrlChecker` calls if the scheme and host of the request URL match `ForgeCfg#apiHost`.

Hopefully fixes #1468.


mzuehlke (approved):
interesting that both braches are already covered with the existing tests

fthomas:
> interesting that both braches are already covered with the existing tests

Right. I'll add a test that checks that auth headers are actually added to the request. :-)

fthomas:
Just tested this with a private Bitbucket Server and it works for me. The repo URL, release notes, and version diff are now included in the PR body.

exoego (approved):
Nice

lenguyenthanh:
This currently doesn't work for private github repositories. because normal url like `https://`github.com/org/private-repository` will not accept any token. So both basic auth and jwt token don't work.

I'm not sure this is a regression or it has never worked?

Can anyone confirm?

PR #2921, opened 2023-01-14
Change Context.printBanner to a field

fthomas:
This is a cheap trick to increase coverage.


mzuehlke:
> This is a cheap trick to increase coverage.

😉 maybe cheap, but finally 🎉 **90.0 % coverage** 🎉

PR #2922, opened 2023-01-14
Add `ForgeSpecificCfg` trait

fthomas:
This adds the `Config.ForgeSpecificCfg` trait which allows us to only pass config options to `ForgeSelection` that are relevant for the different `ForgeApiAlg`s instead of the complete `Config` object.

This simplifies the tests a little bit because we don't need create copies of `Config` to change forge specific configuration.


exoego (approved):
Makes sense

mzuehlke (approved):
Wouldn't it be possible to just store the `forgeSpecificCfg: ForgeSpecificCfg` as part of the `ForgeCfg` in the `Config` ?
Because at any time only one can be present.
Then the `tpe` would be redundant, too.

But this is already a good step.

fthomas:
> Wouldn't it be possible to just store the `forgeSpecificCfg: ForgeSpecificCfg` as part of the `ForgeCfg` in the `Config` ?

Sure, but there are a few places where only the `ForgeCfg` is required but not the `ForgeSpecificCfg`. I therefore would like to keep them separate.

PR #2923, opened 2023-01-15
Log a warning, when the workspace directory doesn't exist

### LINKED ISSUE #2889: Log a warning at start-up if the workspace is empty
See https://github.com/scala-steward-org/scala-steward/pull/2886#issuecomment-1372712263.

mzuehlke:
See https://github.com/scala-steward-org/scala-steward/pull/2886#issuecomment-1372712263.

Fixes #2889

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/SelfCheckAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/SelfCheckAlg.scala
@@ -21,12 +21,14 @@ import cats.syntax.all._
 import org.scalasteward.core.edit.scalafix.ScalafixCli
 import org.scalasteward.core.edit.scalafix.ScalafixCli.scalafixBinary
 import org.scalasteward.core.git.GitAlg
+import org.scalasteward.core.io.FileAlg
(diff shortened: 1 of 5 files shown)


exoego (approved):

modules/core/src/main/scala/org/scalasteward/core/application/SelfCheckAlg.scala
```diff
@@ -37,12 +37,25 @@ final class SelfCheckAlg[F[_]](config: Config)(implicit
...
       _ <- logger.info("Run self checks")
+      _ <- checkWorkspaceDirectory
       _ <- checkGitBinary
       _ <- checkScalafixBinary
       _ <- checkScalafmtBinary
       _ <- checkUrlChecker
     } yield ()
 
+  private def checkWorkspaceDirectory: F[Unit] =
+    F.whenA(!config.workspace.exists)(
```
fthomas, line 48: I've two comments:
 * `config.workspace.exists` has side-effects, we need to use `FileAlg` here
 * If the workspace exists, we should also check if it is empty.

fthomas (approved):
The `StewardAlgTest` failed. Other than that this LGTM.

PR #2924, opened 2023-01-15
add test for GitHubAuthAlg

mzuehlke:


modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubAuthAlg.scala
```diff
@@ -32,7 +32,11 @@ trait GitHubAuthAlg[F[_]] {
 
   /** [[https://docs.github.com/en/free-pro-team@latest/developers/apps/authenticating-with-github-apps#authenticating-as-a-github-app]]
     */
-  def createJWT(app: GitHubApp, ttl: FiniteDuration): F[String]
+  def createJWT(
+      app: GitHubApp,
+      ttl: FiniteDuration,
+      nowMillis: Long = System.currentTimeMillis()
```
fthomas, line 38: `System.currentTimeMillis()` is a side effect that needs to be wrapped in `F.delay`. You could remove the default value here and add a second `createJWT` with two parameters that calls this method and is implemented like this: `F.delay(System.currentTimeMillis()).flatMap(createJWT(app, ttl, _))`.

mzuehlke: Thanks for the hint, I will change it.

fthomas (approved):

PR #2925, opened 2023-01-15
Add "versionScheme" labels to pull-requests.

### LINKED ISSUE #2541: Versioning scheme of dependencies is not taken into account

mzuehlke:
An additional label is added that reflect the `versionScheme` of the updated artifact. This is a new label. The existing labels have not been changed to ensure compatibility.

- `versionScheme`reference: https://www.scala-sbt.org/1.x/docs/Publishing.html#Version+scheme
- As suggested in https://github.com/scala-steward-org/scala-steward/pull/2863#issuecomment-1368220133 getting the data was trivial.

fixes #2541


fthomas:
I like this! The only thing I would change is how the label is rendered. I would prefer `version-scheme:$scheme` over `version-scheme-$scheme` to make it clearer that `$scheme` is the value of the `info.versionScheme` property.

I'd also note that this label is purely informational in case of `pvp` with version numbers that have four or more components `A.B.C.D`. With such versions none of the other `semver` labels we currently have are present, so `version-scheme:pvp` would then be the only version related label. But if we look at that label we don't know if the bump is [`breaking`](https://pvp.haskell.org/#version-numbers) (`A.B` is different), `non-breaking` (`C` is different), or `other` (`D` is different). I think this is fine for now, since we can extend this label in the future by appending a change identifier to it: `version-scheme:pvp:non-breaking`.

If we have the `version-scheme` label with the change identifier, maybe we can also deprecate and remove the other `early-semver` and `semver-strict` labels we currently have. For example, if there is no `versionScheme` defined but the version numbers have three components and both start with zeros, Scala Steward could add a `version-scheme:early-semver-inferred:minor`. It is `inferred` because Scala Steward inferred the scheme from the version numbers.

mzuehlke:
> I like this! The only thing I would change is how the label is rendered. I would prefer `version-scheme:$scheme` over `version-scheme-$scheme` to make it clearer that `$scheme` is the value of the `info.versionScheme` property.

Good suggestion, I adapted the label format.
And I like your ideas to enhance the new labels in the future 👍 

/cc @julienrf (because you suggested the feature)

exoego (approved):

fthomas (approved):

fthomas:
@exoego Are you waiting for further changes or can we merge this?

PR #2953, opened 2023-01-31
Add ability to add assignees and reviewers for GitLab and GitHub

danielleontiev:
Notable changes:

**Gitlab**

- Adding reviewers and assignees is possible with "create MR" request by providing lists of **user ids** for reviewers and assignees and it was implemented by
  - getting usernames from config
  - fetching user ids by usernames using /users API

**GitHub**

- GitHub does not support creating PR and add reviewers and assignees in one call - adding was implemented as separate calls.
- Team reviewers are specified with `organization/team`

**Other**

- Azure and BitBucket do not support assignees - warning was added

- Azure and BitBucket support reviewers but it was not implemented at the time, instead warning was added saying that implementation is missing for now

- Gitea supports both reviewers and assignees, warning was added saying that implementation is missing for now


fthomas:
Thanks for the PR, @danielleontiev!

I've a quick question. The last time I looked into this (https://github.com/scala-steward-org/scala-steward/issues/504#issuecomment-495316099) assigning PRs on GitHub required write access to the repo. I guess this hasn't changed, right? If this is still the case, there should be a note in `docs/repo-specific-configuration.md` that this will only work if the Scala Steward user has the appropriate permissions. Maybe also note that it won't work for PRs created by @scala-steward.

danielleontiev:
Hi @fthomas!
Yes, you are right, I've checked and it only works when account has write access (for GitHub) or Developer role (for GitLab). Added a note to documentation

fthomas:
I just tried this branch locally and it tried to label a PR and failed:
```
[info] 2023-02-04 10:52:41,816 ERROR Steward scala-steward-org/scala-steward failed
[info] org.scalasteward.core.util.UnexpectedResponse: uri: https://api.github.com/repos/scala-steward-org/scala-steward/issues/2956/labels
[info] method: POST
[info] status: 403 Forbidden
[info] headers:
[info]   access-control-allow-origin: *
[info]   access-control-expose-headers: ETag, Link, Location, Retry-After, X-GitHub-OTP, X-RateLimit-Limit, X-RateLimit-Remaining, X-RateLimit-Used, X-RateLimit-Resource, X-RateLimit-Reset, X-OAuth-Scopes, X-Accepted-OAuth-Scopes, X-Poll-Interval, X-GitHub-Media-Type, X-GitHub-SSO, X-GitHub-Request-Id, Deprecation, Sunset
[info]   content-length: 142
[info]   content-security-policy: default-src 'none'
[info]   content-type: application/json; charset=utf-8
[info]   date: Sat, 04 Feb 2023 09:52:41 GMT
[info]   referrer-policy: origin-when-cross-origin, strict-origin-when-cross-origin
[info]   server: GitHub.com
[info]   strict-transport-security: max-age=31536000; includeSubdomains; preload
[info]   vary: Accept-Encoding, Accept, X-Requested-With
[info]   x-accepted-oauth-scopes: 
[info]   x-content-type-options: nosniff
[info]   x-frame-options: deny
[info]   x-github-api-version-selected: 2022-11-28
[info]   x-github-media-type: github.v3
[info]   x-github-request-id: 9A9F:77A8:1B8CB4E:1C2384D:63DE2AE9
[info]   x-oauth-scopes: delete_repo, notifications, repo, workflow
[info]   x-ratelimit-limit: 5000
[info]   x-ratelimit-remaining: 4990
[info]   x-ratelimit-reset: 1675507865
[info]   x-ratelimit-resource: core
[info]   x-ratelimit-used: 10
[info]   x-xss-protection: 0
[info] body: {"message":"Must have admin rights to Repository.","documentation_url":"https://docs.github.com/rest/reference/issues#add-labels-to-an-issue"}
```
This does not happen on main.

danielleontiev:
> This does not happen on main.

Do you run it with or without `--add-labels` switch?

danielleontiev:
Fixed the issue and added unit test for that case

fthomas:
Hey @danielleontiev, I've one request for future changes. Please don't force push them. That makes it impossible for me to track what has been changed (and thus harder to review the PR) and a lot harder to incorporate them in my local checkout.

danielleontiev:
Ok, no problem! Then - when it will be ready to merge, please, squash all commits into one, preserving the first commit message, if you don't mind

fthomas:
Just tried your latest change, but the labels in the PR body are now gone: https://github.com/scala-steward-org/test-repo-2/pull/179

`--add-labels` should not affect the labels in the PR body but if they are also added via the forge API.

danielleontiev:
> Just tried your latest change, but the labels in the PR body are now gone: [scala-steward-org/test-repo-2#179](https://github.com/scala-steward-org/test-repo-2/pull/179)
> 
> `--add-labels` should not affect the labels in the PR body but if they are also added via the forge API.

Fixed, thank you for pointing out!

danielleontiev:
@fthomas I see that Gitea support was merged into main - I've updated the PR to take in into consideration

fthomas:
@danielleontiev If I understand your changes correctly, the change of how labels are handled is just an optimization that is not necessary for implementing assignees and reviewers. If that is correct, I would ask you to do this optimization in a separate PR. That should make it easier to see which changes are required for assignees and reviewers and which changes are done for the labels optimization.

danielleontiev:
@fthomas I've rolled back all unrelated changes and will do a separate PRs for them after we will finish with the current one

fthomas:
Many thanks @danielleontiev! The diff now looks much easier to review.

modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
```diff
@@ -193,6 +194,9 @@ object Context {
         .create[F, Repo, RepoCache]("repo_cache", "1", kvsPrefix)
       versionsStore <- JsonKeyValueStore
         .create[F, VersionsCache.Key, VersionsCache.Value]("versions", "2")
+      gitlabUserIdCacheStore <- JsonKeyValueStore
+        .create[F, String, Int]("gitlab_user_id_cache", "1", kvsPrefix)
+        .flatMap(CachingKeyValueStore.wrap(_))
```
fthomas, line 199: **question:** Could you please motivate why you added a cache for the user name to ID mapping. I understand that it avoids repeated API calls during PR creation but is a cache for that really necessary? Is it expensive or slow to get the IDs by a username? How much slower would Scala Steward be if these API calls are repeatedly done per PR creation?

I'm asking because adding this cache is not for free and from my perspective this optimization also looks like a complication and I don't know why it is necessary.

fthomas (commented):
@danielleontiev Your PR looks very good overall. :+1:

I've a few small suggestions and nitpicks  and a bigger question about the user name to ID cache.

modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
danielleontiev: When I was playing around this API I have not noticed any delays or slow responses - so, I would say these calls to `GET /users` are lightweight. I've added cache because `username -> user ID` mapping is something that does not change often if changed ever at all, rather than `GET /users/ being expensive in terms of time or something else.

I've checked again now and for my location calls to `GET /users` on public gitlab.com API takes ~400ms on average, same as for `GET /merge_requests`.

I run Scala Steward once in two weeks typically and it creates usually around 20 pull requests. For example, last week it have created 22 PRs and it took approximately 5 minutes (only API interaction stage). If I will add 5 assignees to the config then Steward will issue 5 extra calls for each PR (for example 400ms each call, giving 2 extra seconds for each PR) that will give extra 40 seconds for the whole run.

I am not sure it's enough to actually decide to keep the cache since - (1) it's not dramatically big slowdown for the described workflow, (2) the code gets more complex with the cache, (3) if the feature will introduce critical slow down for some workflows it's always possible to do optimization in the future.

On the other hand, I would say that caching here is not extraordinary complex or tricky or decreases readability dramatically while can bring noticeable optimization. 

So, what do you think at that point, should we keep or remove cache finally? My feelings are that I do not mind to either remove or keep it - for me both options are pretty equal

P.S. another option will be to introduce some in-memory caching to avoid storing the data to the disk, but I've not found any adopted solutions for such caching in the codebase so it will require to bring something new. The reasonable thing to do maybe will be to reuse `KeyValueStore` interface and create in-memory implementation for it.

modules/core/src/main/scala/org/scalasteward/core/forge/data/NewPullRequestData.scala
```diff
@@ -37,6 +37,8 @@ final case class NewPullRequestData(
     head: String,
     base: Branch,
     labels: List[String],
+    assignees: List[String],
+    reviewers: List[String],
     draft: Boolean = false
```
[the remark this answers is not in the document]
danielleontiev: Thank you :heart:

modules/core/src/test/scala/org/scalasteward/core/forge/github/GitHubApiAlgTest.scala
```diff
@@ -237,16 +261,46 @@ class GitHubApiAlgTest extends CatsEffectSuite with Http4sDsl[MockEff] {
...
+      body = "body",
+      head = "aaa",
+      base = Branch("master"),
+      labels = Nil,
+      assignees = List("foo"),
+      reviewers = List("bar")
+    )
+
+    gitHubApiAlg.createPullRequest(repo.copy(repo = "cant-assign-reviewers"), data).runA(state)
+  }
```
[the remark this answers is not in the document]
danielleontiev: Nice catch! Thanks

modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
```diff
@@ -30,27 +30,40 @@ import org.scalasteward.core.git.{Branch, Sha1}
...
-      id,
-      List(if (data.draft) "Draft: " else "", data.title).mkString,
-      data.body,
-      projectId,
-      data.head,
-      data.base
+      id = id,
+      title = List(if (data.draft) "Draft: " else "", data.title).mkString,
+      description = data.body,
+      assignee_ids = if (assignees.nonEmpty) Some(assignees) else None,
```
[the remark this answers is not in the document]
danielleontiev: Changed

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubReviewers.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package org.scalasteward.core.forge.github
+
+import io.circe.Encoder
+import io.circe.generic.semiauto.deriveEncoder
+
+case class GitHubReviewers(reviewers: List[String])
+
+object GitHubReviewers {
+  implicit val encoder: Encoder[GitHubReviewers] = deriveEncoder
```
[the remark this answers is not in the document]
danielleontiev: Renamed

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubAssignees.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package org.scalasteward.core.forge.github
+
+import io.circe.Encoder
+import io.circe.generic.semiauto.deriveEncoder
+
+case class GitHubAssignees(assignees: List[String])
+
+object GitHubAssignees {
+  implicit val encoder: Encoder[GitHubAssignees] = deriveEncoder
```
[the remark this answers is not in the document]
danielleontiev: Renamed

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubApiAlg.scala
```diff
@@ -42,11 +45,20 @@ final class GitHubApiAlg[F[_]](
...
   /** https://developer.github.com/v3/pulls/#create-a-pull-request */
-  override def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut] =
-    client
+  override def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut] = {
+    val createPullRequest = client
       .postWithBody[PullRequestOut, NewPullRequestData](url.pulls(repo), data, modify(repo))
       .adaptErr(SecondaryRateLimitExceeded.fromThrowable)
 
+    for {
+      created <- createPullRequest
```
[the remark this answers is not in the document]
danielleontiev: done

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubApiAlg.scala
```diff
@@ -42,11 +45,20 @@ final class GitHubApiAlg[F[_]](
     }
 
   /** https://developer.github.com/v3/pulls/#create-a-pull-request */
-  override def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut] =
-    client
+  override def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut] = {
+    val createPullRequest = client
```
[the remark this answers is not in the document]
danielleontiev: Agree, adjusted

modules/core/src/main/scala/org/scalasteward/core/forge/azurerepos/AzureReposApiAlg.scala
```diff
@@ -96,4 +103,11 @@ final class AzureReposApiAlg[F[_]](
         modify(repo)
       )
       .void
+
+  private def warnIfAssigneesAreUsed() =
```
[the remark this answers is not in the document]
danielleontiev: Done

modules/core/src/main/scala/org/scalasteward/core/forge/azurerepos/AzureReposApiAlg.scala
```diff
@@ -28,26 +28,33 @@ import org.scalasteward.core.forge.azurerepos.JsonCodec._
...
+  override def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut] = {
+    val create = client.postWithBody[PullRequestOut, PullRequestPayload](
       url.pullRequests(repo),
       PullRequestPayload.from(data),
       modify(repo)
     )
+    for {
+      _ <- if (data.assignees.nonEmpty) warnIfAssigneesAreUsed() else F.unit
+      _ <- if (data.reviewers.nonEmpty) warnIfReviewersAreUsed() else F.unit
+      created <- create
```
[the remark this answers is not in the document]
danielleontiev: Changed here and in other places as well

modules/core/src/main/scala/org/scalasteward/core/forge/azurerepos/AzureReposApiAlg.scala
```diff
@@ -28,26 +28,33 @@ import org.scalasteward.core.forge.azurerepos.JsonCodec._
...
-  override def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut] =
-    client.postWithBody[PullRequestOut, PullRequestPayload](
+  override def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut] = {
+    val create = client.postWithBody[PullRequestOut, PullRequestPayload](
       url.pullRequests(repo),
       PullRequestPayload.from(data),
       modify(repo)
     )
+    for {
+      _ <- if (data.assignees.nonEmpty) warnIfAssigneesAreUsed() else F.unit
```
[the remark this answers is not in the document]
danielleontiev: Changed here and in other places as well

modules/core/src/main/scala/org/scalasteward/core/forge/azurerepos/AzureReposApiAlg.scala
```diff
@@ -28,26 +28,33 @@ import org.scalasteward.core.forge.azurerepos.JsonCodec._
...
 import org.scalasteward.core.git.Branch
 import org.scalasteward.core.util.HttpJsonClient
+import org.typelevel.log4cats.Logger
 
 final class AzureReposApiAlg[F[_]](
     azureAPiHost: Uri,
     config: AzureReposCfg,
     modify: Repo => Request[F] => F[Request[F]]
-)(implicit client: HttpJsonClient[F], monadErrorF: MonadThrow[F])
+)(implicit client: HttpJsonClient[F], F: MonadThrow[F], logger: Logger[F])
```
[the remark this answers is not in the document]
danielleontiev: Swapped the order

danielleontiev:
@fthomas thank you for comprehensive review! I've fixed the issues and written some thoughts about the cache. BTW, should I resolve done threads myself or you would like to resolve them during verification?

modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
alejandrohdezma: Also take into account the rate limits the GitHub API has, this cache will ensure we won't go over the roof with those

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubReviewers.scala
```diff
@@ -0,0 +1,26 @@
...
+
+package org.scalasteward.core.forge.github
+
+import io.circe.Encoder
+import io.circe.generic.semiauto.deriveEncoder
+
+case class GitHubReviewers(reviewers: List[String])
+
+object GitHubReviewers {
+  implicit val gitHubReviewersEncoder: Encoder[GitHubReviewers] = deriveEncoder
```
alejandrohdezma, line 25: **thought:** take into account that this won't work for requesting team reviews, since those are done through a different parameter.

See docs: https://docs.github.com/en/rest/pulls/review-requests?apiVersion=2022-11-28#request-reviewers-for-a-pull-request

alejandrohdezma (commented):
Great addition, and a really nice a clean way to introduce it 👏🏼 👏🏼 

I only added a small comment about the reviewers feature on GitHub, if you feel like adding it to this PR, we could discuss what's could be the best approach 😊

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubReviewers.scala
danielleontiev: Thank you for pointing out. Yes I am ready to add this feature as well - either in the current PR or in new one (if the solution will be an addition that won't require to break existing code for assignees/reviewers I feel that it will be easier to separate it as another PR, if we need to change something in the new code - of course we should do it here).

So, the straightforward solution will be to just add `teamReviewers` parameter to configuration but the downside is that it will not be functional for all implementations but GitHub. It's the same situation as for example AzureRepos does not have "assignee" feature, so, should not be a big deal.

What do you think about implementing `teamReviewers` this way?

alejandrohdezma: Maybe adding it so teams are added like: `team:my-org/my-team` so it's easy to pattern-match. It will also be easy to just strip that from forges where both are added the same way.

alejandrohdezma: Or maybe just add them as `my-org/my-team` and pattern-match on GitHub forge for `$org/$team` and use `teamReviewers` for those that match. WDYT?

fthomas:
> I've fixed the issues and written some thoughts about the cache.

Thanks!

> BTW, should I resolve done threads myself or you would like to resolve them during verification?

Both is fine for me.

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubReviewers.scala
danielleontiev: `my-org/my-team` looks neat, I like it!

I only wonder how to make sure that such entries are filtered for implementations other than GitHub? Should we have some common function that will filter out team reviewers in `ForgeApi` implementations, or we should make it on the level above and pass team reviewers as `ForgeSpecificCfg` for GiHub only?

danielleontiev:
> Both is fine for me.

I've resolved done issues

modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
alejandrohdezma: 😓 That's right! I totally missed that. Thanks for the clarification!

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubReviewers.scala
alejandrohdezma: We should not need to filter for other implementations (given this is a forge-specific), on GitHub we can just pattern-match on the pattern. On other forges where there is no distinction we will just use the same string that is added to the config file.

danielleontiev: Ah sure, you are right! Thank you

The change seems not very complex then, I will add it to the current PR

modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
danielleontiev: Cache dropped :+1:

modules/core/src/main/scala/org/scalasteward/core/forge/github/GitHubReviewers.scala
danielleontiev: I've added team reviewers support with `org/team` construct, please take a look

fthomas (approved):
Looks great to me!

Many thanks for working on this, @danielleontiev!

alejandrohdezma (approved):
Amazing work @danielleontiev 👏🏼 👏🏼

PR #2954, opened 2023-02-02
Update `.scalafmt.conf` and `.mill-version` also in subdirectories

### LINKED ISSUE #2947: `.scalafmt.conf` is only updated in the root directory
Scalafmt updates only change the `.scalafmt.conf` in the root directory since #2816.

before #2816:
https://github.com/fthomas/base.g8/pull/151/files

since #2816:
https://github.com/fthomas/base.g8/pull/160/files

We should restore the previous behavior and update all `.scalafmt.conf` files that use the same version.

fthomas:
Closes #2947


alejandrohdezma (approved):
🎉

mzuehlke (approved):
I locally tested, if the setup from the issue with 2 `.scalafmt.conf` files works.
It does 👍

PR #2980, opened 2023-02-18
github: add dedicated model for create PR body

danielleontiev:
this refactoring replaces `NewPullRequestData` that was serialized into Json and used as a payload for POST /repos/{owner}/{repo}/pulls (creating PR on GitHub) because `NewPullRequestData` contains fields that are not processed by GitHub API (such as labels, assignees, reviewers). New model only has fields that are expected by GitHub API.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/forge/data/NewPullRequestData.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/forge/data/NewPullRequestData.scala
@@ -17,8 +17,6 @@
 package org.scalasteward.core.forge.data
 
 import cats.syntax.all._
-import io.circe.Encoder
-import io.circe.generic.semiauto._
 import org.http4s.Uri
 import org.scalasteward.core.data._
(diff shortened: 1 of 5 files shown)


fthomas:
:+1: This much cleaner.

Could you look into why we've lost coverage here? I believe it is because the test with `Update.Grouped` now only calls `bodyFor` instead of `from`.

Since `NewPullRequestData` is not used as payload for a forge API call anymore, the next step could be to replace the current fields of that class with the parameters of `from` and add `title`, `body`, `head`, and `base` methods where the `body` method is equal to the current `bodyFor` implementation. This could be done in subsequent PR.

danielleontiev:
> Could you look into why we've lost coverage here? I believe it is because the test with `Update.Grouped` now only calls `bodyFor` instead of `from`.

Yes, you are right. This was caused by the code branch inside `CommitMsg` for grouped updates:

image

I've added another unit test for that case

fthomas (approved):

mzuehlke (approved):

PR #2981, opened 2023-02-19
refactoring: get rid of NewPullRequestData.from() method

danielleontiev:


danielleontiev:
Hi @fthomas!

I am working on the follow-up about simplification of `NewPullRequestData`:

> Since NewPullRequestData is not used as payload for a forge API call anymore, the next step could be to replace the current fields of that class with the parameters of from and add title, body, head, and base methods where the body method is equal to the current bodyFor implementation.

I have a question: is it reasonable to remove companion object `NewPullRequestData` and make all methods from there the private members of `NewPullRequestData` class? That will require to change tests to only verify things through public interface? A conservative option is to make these methods `private [data]` to not change test largely

fthomas:
> I have a question: is it reasonable to remove companion object `NewPullRequestData` and make all methods from there the private members of `NewPullRequestData` class? That will require to change tests to only verify things through public interface.

I guess that would mean testing almost everything via `body` which would be very inconvenient, right?

> [...] not change test largely

If my suggestion of changing `NewPullRequestData` requires a lot of changes with no gain, maybe it is better to leave it as is. If we view `NewPullRequestData` as the sum of all the different `PullRequestPayload` / `MergeRequestPayload` / `CreatePullRequestOption` classes that we have for the different forges, it is okay that `title`, `body`, etc. are just string fields and the class itself has no knowledge how it has been constructed.

danielleontiev:
> I guess that would mean testing almost everything via `body` which would be very inconvenient, right?

If all other helper methods will become `private` - then yes, we will need to test almost everything via new `body` method. I won't say it's very inconvenient, because all tests will become vary similar and observable - in terms of how different input will change the body. The only controversial part is that test spec will become very very large because every test case will include the whole expected body
 
> If my suggestion of changing `NewPullRequestData` requires a lot of changes with no gain, maybe it is better to leave it as is.

Actually I like you suggestion for changing it and it would be better to make "computable" parts as methods. It has clear benefit on readability of the code. I think the only question is whether "private" methods of updated class should become truly `private` or `private[data]` for the sake of simplified testing

fthomas:
> The only controversial part is that test spec will become very very large because every test case will include the whole expected body

I think I wouldn't like that. :-)
 
>  I think the only question is whether "private" methods of updated class should become truly `private` or `private[data]` for the sake of simplified testing

I'm fine with `private[data]` if it doesn't mean the above.

danielleontiev:
@fthomas I have done some refactoring so far but I need some feedback early because I am not sure whether I am going to the right direction. Could you please take a look?

So far I've done:
- made `NewPullRequestData.apply` parameters as in old `from` method
- grouped artifacts-related parameters to `ArtifactsMetadata`
- Split `preparePullRequest` into two more short methods
- made methods inside `NewPullRequestData` that depends on some state private members of the class (f.e. `migrationNote` depends on `update` that passed into `NewPullRequestData.apply`, hence we have `NewPullRequestData(...).migrationNote`)
- Kept completely util methods (such as `fromTo`) as companion object's members since they are simple and do not depend on `NewPullRequestData`
- NewPullRequestDataTest adjusted to support changes described above

danielleontiev:
Hi @fthomas! Did you have a chance to take a look at the code?

fthomas:
The more I think about this, the more I become convinced that my suggestion in https://github.com/scala-steward-org/scala-steward/pull/2980#issuecomment-1435742474 was bad. The purpose of `NewPullRequestData` is to group the data needed by the `ForgeApiAlg` implementations to create PRs so that we just need pass this one parameter to `ForgeApiAlg#createPullRequest`. The `ForgeApiAlg` implementations don't need to know how the PR body was constructed or from which data it is constructed. For the purpose of creating the PR via the Forge APIs, a `body: String` is fine and if the `NewPullRequestData` class has only simple fields like `title:String, body: String, reviewers: List[String], ...` that is fine too. I would even argue that it keeps `ForgeApiAlg` simple, because if you look at the interface you don't see anything about `Update`, `UpdateData`, `RepoConfig`, `RepoCache`, or `EditAttempt`s which would all be inside `NewPullRequestData` instead of the simple `String` and `List[String]` fields that are currently there. The `NewPullRequestData` as it is now separates the concerns of `NurtureAlg` from the concerns of `FrogeApiAlg` and that is a good thing.

I'm sorry didn't thought about this more deeply as I wrote https://github.com/scala-steward-org/scala-steward/pull/2980#issuecomment-1435742474. :-(

danielleontiev:
Ok, after some working on it and thinking I agree that the change seems to produce a big code diff without a real benefit. As for now maybe it's better to abandon it

PR #2982, opened 2023-02-21
git check-ignore before adding

### LINKED ISSUE #2942: Creating `.git-blame-ignore-revs` can error the Scala Steward process if that file is covered by `gitignore`
The Guardian's [Scala Steward GitHub Action](https://github.com/guardian/scala-steward-public-repos/actions/workflows/public-repos-scala-steward.yml) (covering all the public repos in our GitHub org) job [started erroring](https://github.com/guardian/support-service-lambdas/pull/1834), and the culprit turned out to be the `git add .git-blame-ignore-revs` added with https://github.com/scala-steward-org/scala-steward/pull/2652 :

```
2023-01-19 10:02:07,438 ERROR Steward guardian/support-service-lambdas failed
  java.io.IOException: 'GIT_ASKPASS=/home/runner/scala-steward/askpass.sh "SBT_OPTS=-Xmx2048m -Xss8m -XX:MaxMetaspaceSize=512m" git -c core.hooksPath=/dev/null add /home/runner/scala-steward/workspace/repos/guardian/support-service-lambdas/.git-blame-ignore-revs' exited with code 1
  The following paths are ignored by one of your .gitignore files:
  .git-blame-ignore-revs
  hint: Use -f if you really want to add them.
  hint: Turn this message off by running
  hint: "git config advice.addIgnoredFile false"
```

It turns out we had [quite broad](https://github.com/guardian/support-service-lambdas/blob/4952dab4c35793b620e82b6f338244628e071962/.gitignore#L1) gitignore configuration in the repo (ignoring all files starting with `.` by default). The file `.git-blame-ignore-revs` starts with a `.`, and `git add` doesn't like adding files that are gitignored.

The gitignore configuration in the troublesome repo is quite easily fixed once the problem's identified, but it took a bit of digging to understand what the issue was, and so I was wondering if there was a possible improvement to Scala Stewards behaviour here? Some possible options:

* **Force `git add`** : Add the `-f` flag to the `git add` command, so Scala Steward just overrides the gitignore configuration, goes ahead and creates the file. https://github.com/scala-steward-org/scala-steward/blob/02acd65ff3243b80254eb44c690eb07ea93ef238/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala#L89
* **Informative warning** : Somehow test if `.git-blame-ignore-revs` can be added, and if not, just provide a friendly warning, rather than having the entire job (which, for a GitHub action, can cover a very large number of repos) being marked as a 'Failure'. image

I'm not sure which approach would be considered best, but I could have a go at a PR for **Force `git add`**, if that would be helpful.

daddykotex:
use git check-ignore before adding a file to the stage in the HookExecutor.

Closes #2942


modules/core/src/main/scala/org/scalasteward/core/git/FileGitAlg.scala
```diff
@@ -47,6 +47,11 @@ final class FileGitAlg[F[_]](config: GitCfg)(implicit
   override def checkoutBranch(repo: File, branch: Branch): F[Unit] =
     git_("checkout", branch.name)(repo).void
 
+  override def checkIgnore(repo: File, file: String): F[Boolean] =
+    git_("check-ignore", file)(repo)
+      .as(true)
+      .recover { case _: ProcessFailedException => false }
```
fthomas, line 53: I suggest to only recover here if the process [exits with 1](https://git-scm.com/docs/git-check-ignore#_exit_status):
```scala
.recover { case ex: ProcessFailedException if ex.exitValue === 1 => false }
```
This would require to make `exitValue` a field of `ProcessFailedException`. It is currently only a constructor parameter.

fthomas (commented):
This looks great, @daddykotex! I've only a small suggestion about checking the exit value of `git check-ignore`.

modules/core/src/main/scala/org/scalasteward/core/git/FileGitAlg.scala
daddykotex: Updated

fthomas: Thanks!

fthomas (approved):

PR #2985, opened 2023-02-22
Expand help strings for github-app-id and github-app-key-file

alexklibisz:
Based on the discussion in https://github.com/scala-steward-org/scala-steward/issues/2973

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -83,9 +83,9 @@ Options and flags:
     --azure-repos-organization <string>
         The Azure organization (required when --forge-type is azure-repos)
     --github-app-id <integer>
-        GitHub application id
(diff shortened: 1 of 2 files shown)


alejandrohdezma (changes requested):
Thanks for this @alexklibisz, one minor clarification though, when using GitHub App credentials, [they are used along with the contents of `repos.md`](https://github.com/scala-steward-org/scala-steward/blob/684e2f370a1d0304b6306ed6161caae06c308707/modules/core/src/main/scala/org/scalasteward/core/application/StewardAlg.scala#L97-L98).

alexklibisz:
Good point, thanks. Fixed in last commit.

alejandrohdezma (approved):

PR #2988, opened 2023-02-26
Support updating `.config/mill-version`

lefou:


modules/core/src/main/scala/org/scalasteward/core/edit/update/Selector.scala
```diff
@@ -169,8 +169,11 @@ object Selector {
       update: Update.Single,
       versionPositions: List[VersionPosition]
   ): List[VersionPosition] =
-    if (isMillMainUpdate(update))
-      versionPositions.filter(_.version.path.endsWith(millVersionName))
+    if (MillAlg.isMillMainUpdate(update))
+      versionPositions.filter(f =>
+        f.version.path.endsWith(MillAlg.millVersionName) ||
```
fthomas, line 174: Since `millVersionName.endsWith(millVersionNameInConfig) == true` it is sufficient to only keep `f.version.path.endsWith(MillAlg.millVersionNameInConfig)` here.

lefou: You are right, I had the same though initially. Yet, it assumes we value of the constants never change. Of course, I can swap the two checks to profit from the stated effect, but suggest to keep both checks for correctness. It's unlikely that in case someone changes the values of these constants, she will detect this specific optimization, as only one of the constants is referenced.

lefou: I swapped the two checks to optimize for the current case.

fthomas (approved):

PR #2992, opened 2023-02-28
docs(repo-specific): fix title for grouping

ropiteaux:
Without the double quotes, it broke my grouping configuration. It seems it was forgotten in this doc


alejandrohdezma (approved):
Hey @ropiteaux, thanks for bringing this up. It's very weird, I would have expect for a `conf` file to be able to not use quotes there. Either way I don't think this change would harm 😊

ropiteaux:
Yeah I wouldn't expect either 😅 , scala steward just updated one by one PR after trying to change the title like this 🤷 (instead of using the grouping)

alejandrohdezma:
Yeah, it is very weird indeed. I will try to add a test checking this 😊

PR #2994, opened 2023-02-28
Ignore file excluded from git

daddykotex:
Git has a .gitignore which can be used to exclude files from getting into the repository. It's pointless to scan files ignored via this mechanism as we'll fail to stage any changes to them.

This approach reads one .gitignore are the top of the file and then use it to check all the paths. This is not _exact_ but I think it's good enough.

I say that this is not exact because Git supports .gitignore file nested at arbitrary depth in the repository. To implement that, we'd need to re-implement the `walk` function which is not worth it in the scope of this issue.

Fixes https://github.com/scala-steward-org/scala-steward/issues/2979


modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala
```diff
@@ -27,21 +27,26 @@ class EditAlgTest extends FunSuite {
     val update = ("org.typelevel".g % "cats-core".a % "1.2.0" %> "1.3.0").single
     val file1 = repoDir / "build.sbt"
     val file2 = repoDir / "project/Dependencies.scala"
+    val gitignore = repoDir / ".gitignore"
 
     val state = MockState.empty
-      .addFiles(file1 -> """val catsVersion = "1.2.0"""", file2 -> "")
+      .addFiles(file1 -> """val catsVersion = "1.2.0"""", file2 -> "", gitignore -> "")
```
fthomas, line 33: **suggestion**: It would be nice if `gitignore` here wouldn't be empty and there would be a file that would be read if `gitignore` would be empty to test that we actually ignore files ignored by git.

daddykotex: I added a `target/` entry to the gitignore file then added this file `target / Somefile.scala`  and the test confirms it is not read.

fthomas: Thanks!

fthomas:
This LGTM, thanks @daddykotex!

As I was looking at this change, I remembered that we already have [`GitAlg#findFilesContaining`](https://github.com/scala-steward-org/scala-steward/blob/e60240fdd8c21e2631fd91dc3e48eb9f7807e027/modules/core/src/main/scala/org/scalasteward/core/git/FileGitAlg.scala#L92) that uses [`git grep`](https://git-scm.com/docs/git-grep) to find files with a given fixed string. If we would use that function in [`VersionScanner#findPathsContaining`](https://github.com/scala-steward-org/scala-steward/pull/2994/files#diff-be0679497c64e44046717f77f5117a31fbf066e5f22e86d65ed7f92f47f68206L59) instead of `FileAlg#findFiles`, we would get the exclusion of git ignored files for free. I suspect that using `GitAlg#findFilesContaining` in combination with `FileAlg#readFile` (which would be required because the former only returns a list of matching files) would be faster than our own simple `FileAlg#findFiles`. The only thing that gets more complicated would be the `EditAlgTest` and `RewriteTest` which would actually need to call `git grep` for them to work.

There is no need to change this PR, @daddykotex. I just wanted to throw this idea out and if we think this is a worthwhile change, we can do this in a follow-up PR.

fthomas (approved):

PR #2995, opened 2023-02-28
Check up to 5 nested project folders

mkurz:
Some projects set up plugins in `project/project/plugins.sbt` and even in `project/project/project/plugins.sbt` (I worked on such a private project, so I can't share details, but such things do exists :wink:)

For example, Play and the sbt-web-build-base project put [sbt-buildinfo](https://github.com/sbt/sbt-buildinfo) in their `project/project/plugins.sbt` files:
* https://github.com/playframework/playframework/blob/main/project/project/buildinfo.sbt
* https://github.com/sbt/sbt-web-build-base/blob/master/project/project/buildinfo.sbt

However, as you can see, even though scala-steward is enabled and visiting those repos, that plugin never gets updated ([latest release is 0.11.0](https://github.com/sbt/sbt-buildinfo/releases), also see [maven central](https://repo1.maven.org/maven2/com/eed3si9n/sbt-buildinfo_2.12_1.0/0.11.0/))

Making scala-steward check and update those dependencies is actually pretty easy and straight forward:
We just need to set up the special `scala-steward-StewardPlugin_<version>.scala` plugin in the corresponding `project/project/project/project/` and `project/project/project/` folders so they are available as plugins for the parent "build" and then just enter those sub "build" projects via `reloadPlugins` again and also run `stewardDependencies` which also logs the dependencies of those projects.
Nothing special, just a couple of more lines logged and parsed by scala steward.
I also think the test that I adjusted in this pull request explains everything.

To make sure everything works 100% I set up following test repo:
 * https://github.com/mkurz/three-level-sbt-plugins-project

As you can see it has three levels of `project` folders including `plugins.sbt`:
* https://github.com/mkurz/three-level-sbt-plugins-project/tree/main/project/project/project
* https://github.com/mkurz/three-level-sbt-plugins-project/tree/main/project/project
* https://github.com/mkurz/three-level-sbt-plugins-project/tree/main/project

So I added outdated plugins to each `plugins.sbt` and the `build.sbt` and ran scala steward with this pull request applied [locally](https://github.com/scala-steward-org/scala-steward/blob/main/docs/running.md#running-scala-steward) and voilà, scala steward opened all expected pull requests:
* https://github.com/mkurz/three-level-sbt-plugins-project/pull/3/files
* https://github.com/mkurz/three-level-sbt-plugins-project/pull/4/files
* https://github.com/mkurz/three-level-sbt-plugins-project/pull/1/files
* https://github.com/mkurz/three-level-sbt-plugins-project/pull/2/files

BTW:
If you want to test out what the output of the sbt command ran by sbt looks like, you can check out the [`steward-plugin-file` branch](https://github.com/mkurz/three-level-sbt-plugins-project/tree/steward-plugin-file) of my test repo and ran the sbt command I show in the README. It is [exactly what scala steward runs](https://github.com/scala-steward-org/scala-steward/blob/f8bc50b768b371be40ba07799333ad631e881c55/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala#L139-L143) to [figure out the versions](https://github.com/scala-steward-org/scala-steward/blob/f8bc50b768b371be40ba07799333ad631e881c55/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala#L59-L60). (For this to `stewardDependencies` command to work I [manually added](https://github.com/mkurz/three-level-sbt-plugins-project/commit/3f1d38431c81e60622b064cfc609377e0f543d03) the `StewardPlugin_1_3_11.scala` files myself, see the README again).

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
@@ -51,21 +51,24 @@ final class SbtAlg[F[_]](config: Config)(implicit
       version = maybeProperties.flatMap(parser.parseBuildProperties)
     } yield version
 
+  private def metaBuildsCount(buildRootDir: File): F[Int] =
+    fs2.Stream
+      .iterate(buildRootDir / project)(_ / project)
+      .take(5L) // Use an upper bound for the meta-builds count to prevent DOS attacks.
+      .evalMap(fileAlg.isDirectory)
+      .takeWhile(identity)
+      .compile
+      .count
+      .map(_.toInt)
+
   override def getDependencies(buildRoot: BuildRoot): F[List[Scope.Dependencies]] =
     for {
       buildRootDir <- workspaceAlg.buildRootDir(buildRoot)
       maybeSbtVersion <- getSbtVersion(buildRootDir)
-      metaBuilds = Math.max(
-        LazyList
-          .iterate(buildRootDir / project)(_ / project)
-          .takeWhile(p => p.exists && p.isDirectory)
-          .size,
-        1 // There is always at least one meta build, even if there is no project folder
-      )
+      metaBuilds <- metaBuildsCount(buildRootDir)
       lines <- addStewardPluginTemporarily(buildRootDir, maybeSbtVersion, metaBuilds).surround {
-        val commands = Nel.of(crossStewardDependencies) ++ List
-          .fill(metaBuilds)(List(reloadPlugins, stewardDependencies))
-          .flatten
+        val commands = Nel.of(crossStewardDependencies) ++
+          List.fill(metaBuilds)(List(reloadPlugins, stewardDependencies)).flatten
         sbt(commands, buildRootDir)
       }
       dependencies = parser.parseDependencies(lines)
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
@@ -54,7 +54,7 @@ final class SbtAlg[F[_]](config: Config)(implicit
   private def metaBuildsCount(buildRootDir: File): F[Int] =
     fs2.Stream
       .iterate(buildRootDir / project)(_ / project)
-      .take(5L) // Use an upper bound for the meta-builds count to prevent DOS attacks.
+      .take(5L) // Use an upper bound for the meta-builds count to prevent DoS attacks.
       .evalMap(fileAlg.isDirectory)
       .takeWhile(identity)
       .compile
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
@@ -55,8 +55,17 @@ final class SbtAlg[F[_]](config: Config)(implicit
     for {
       buildRootDir <- workspaceAlg.buildRootDir(buildRoot)
       maybeSbtVersion <- getSbtVersion(buildRootDir)
-      lines <- addStewardPluginTemporarily(buildRootDir, maybeSbtVersion).surround {
-        val commands = Nel.of(crossStewardDependencies, reloadPlugins, stewardDependencies)
+      metaBuilds = Math.max(
+        LazyList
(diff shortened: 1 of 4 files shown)


mkurz:
Would be nice if this PR makes it into v0.22.0 :wink:

mkurz:
Hey! I am pretty sure this pull request is good to go, reviews would be highly appreciated. Thanks!

fthomas:
Thanks for the PR, @mkurz! I'm fine with updating dependencies in meta-meta-builds and even meta-meta-meta-builds but I think it would be better to check how many nested `project` directories there are before calling `reload plugins; stewardDependencies` thrice. My concern is that calling `reload plugins` can take a few seconds and we're slowing down Scala Steward unnecessarily for the majority of sbt projects.

I would check how many (nested) `project` directories there are and then only adding the `StewardPlugin_*.scala` to them if they exist. We would then only need to call `reload plugins; stewardDependencies` as much as needed.

mkurz:
I get it. Will update when I find time, thanks!

mkurz:
@fthomas Updated. It works exactly like you asked for, depending on the levels of meta builds. I tested that again on a real project: https://github.com/mkurz/three-level-sbt-plugins-project/pulls (I deleted the repo and recreated it so it was fresh again). As you can see a pull request for each meta build was created. After one run I even created one more level with the scalafmt plugin and on the second run scala-steward also created the according pull request (you can see that the sbt-scalafmt pull request was opened 4 minutes later).

modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
```diff
@@ -55,8 +55,17 @@ final class SbtAlg[F[_]](config: Config)(implicit
...
       buildRootDir <- workspaceAlg.buildRootDir(buildRoot)
       maybeSbtVersion <- getSbtVersion(buildRootDir)
-      lines <- addStewardPluginTemporarily(buildRootDir, maybeSbtVersion).surround {
-        val commands = Nel.of(crossStewardDependencies, reloadPlugins, stewardDependencies)
+      metaBuilds = Math.max(
+        LazyList
+          .iterate(buildRootDir / project)(_ / project)
+          .takeWhile(p => p.exists && p.isDirectory)
+          .size,
+        1 // There is always at least one meta build, even if there is no project folder
```
fthomas, line 63: I'm wondering if we should drop this `Math.max` call. If there is no `project` folder, there aren't any `*.sbt` files either which could contain dependencies (plugins) that Scala Steward could update. And if there is nothing to update, why loading the meta-build and calling `stewardDependencies`? :thinking:

mkurz: In theory yes. However removing `Math.max` will break many many tests because of lines like
https://github.com/scala-steward-org/scala-steward/blob/a6ff326cb8c7d423ccbd5a656a771d1ef9c634a4/modules/core/src/test/scala/org/scalasteward/core/buildtool/sbt/SbtAlgTest.scala#L29
https://github.com/scala-steward-org/scala-steward/blob/a6ff326cb8c7d423ccbd5a656a771d1ef9c634a4/modules/core/src/test/scala/org/scalasteward/core/buildtool/sbt/SbtAlgTest.scala#L43

which assume scala-steward always checks/updates the `project` folder. And I would keep all those tests how they are currently.
Also in real life I never saw a sbt project without a `project` folder (sbt even creates the `project` folder when running in an empty folder).

I would just keep it.

fthomas: > And I would keep all those tests how they are currently.

We can keep them if we use `FileAlg#isDirectory` for checking the `project` folders as I've done in  https://github.com/scala-steward-org/scala-steward/pull/2995/commits/1a97fc56c00bdaa9b468d4b6ae52ed87cd351489. Another benefit of using `FileAlg` is that the tests are now sensitive to the actual number of `project` folders.

mkurz: Great! Thanks!

mkurz:
LGTM :+1:

mkurz:
Tested again with a real world repo, everything works great: https://github.com/mkurz/three-level-sbt-plugins-project/pulls

fthomas:
Awesome, thanks for testing again!

fthomas (approved):

PR #2996, opened 2023-02-28
Make sure remote is always called "origin" when cloning git repo

mkurz:
While working on #2995 I ran into a problem that prevented me from running scala steward locally on my machine.
That's because in my `~/.gitconfig` I configured that the default remote of a freshly cloned git repo should always be called `upstream`:
```
[clone]
        defaultRemoteName = upstream
```

However, if you grep through the scala-steward source code, you will see that scala-stewared heavily and basically everywhere and always relies on the remote's name to be `origin`.

So this pull requests make sure that when scala-steward clones a git repository the name of the remote is always called "origin" by overriding global git config (user) setting directly at the git-clone command.
With this fix I was able to run scala-steward locally, without it it always failed (or better: the git commands scala-steward calls failed in their sub-processes).

BTW, I had the same issue already with pacman (package manager for arch linux), where I applied the same fix:
* https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/19 (you can read more details about why I use `-c clone.defaultRemoteName=origin` in that merge request)
* that merge request of mine :point_up: is now be incorporated into https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/56

I am 100% sure this fix is correct (since I tested it locally :wink:). Thanks!

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/git/FileGitAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/git/FileGitAlg.scala
@@ -55,7 +55,9 @@ final class FileGitAlg[F[_]](config: GitCfg)(implicit
   override def clone(repo: File, url: Uri): F[Unit] =
     for {
       rootDir <- workspaceAlg.rootDir
-      _ <- git_("clone", url.toString, repo.pathAsString)(rootDir)
+      _ <- git_("clone", "-c", "clone.defaultRemoteName=origin", url.toString, repo.pathAsString)(
+        rootDir
+      )
     } yield ()
 
   override def cloneExists(repo: File): F[Boolean] =
--- a/modules/core/src/test/scala/org/scalasteward/core/forge/ForgeRepoAlgTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/forge/ForgeRepoAlgTest.scala
@@ -37,7 +37,14 @@ class ForgeRepoAlgTest extends CatsEffectSuite {
     val expected = MockState.empty.copy(
       trace = Vector(
         Log("Clone scala-steward/datapackage"),
-        Cmd(gitCmd(config.workspace), "clone", forkUrl, repoDir.toString),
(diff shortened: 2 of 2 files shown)


alejandrohdezma:
Hey @mkurz! From what I see in [git-glone docs](https://git-scm.com/docs/git-clone/2.30.0#Documentation/git-clone.txt---originltnamegt) the suggested way to override the global configuration is using the `--origin` flag. Would you mind changing it to that?

mkurz:
@alejandrohdezma I explained the reason why I use `-c clone.defaultRemoteName=...` and not `--origin` in https://gitlab.archlinux.org/pacman/pacman/-/merge_requests/19 already:

> You might think why use `-c clone.defaultRemoteName=...` and not `-o ...`? That's because `-o` is not allowed together with `--bare` (which is implied when using `--mirror`). This is strange and makes no sense, that's why upcoming (not yet released) git v2.38.2 will lift this restriction:
- https://github.com/git/git/blob/c03801e19cb8ab36e9c0d17ff3d5e0c3b0f24193/Documentation/RelNotes/2.38.2.txt#L20-L21
- https://github.com/git/git/commit/0d5d92906a8c18d55174531312587baf1dc1d2b5

So `--clone` doesn't play well with `--bare` for git versions <2.38.2. I know that in the scala-steward source code neither the `--bare` nor the `--mirror` flags are used, however in case after this PR a bare clone command gets added to scala steward, for whatever reason, scala-steward would break for people using an older git version (Ubuntu 18.04 or 20.04 pretty sure still have an older git version installed).

Further I wrote in the pacman issue:
> I think it's reasonable to just use the `-c ...` config for now, it's equivalent anyway and setting it on the command line will overrule any global settings (precedence are listed in the git message [here](https://github.com/git/git/commit/de9ed3ef3740f8227cc924e845032954d1f1b1b7))

So I think it's safer for now to stick with `-c clone.defaultRemoteName=...´

mzuehlke (approved):

PR #3000, opened 2023-03-03
refactoring: remove ForgeApiAlg.labelPullRequest

danielleontiev:
ForgeApiAlg.labelPullRequest only used during creation of pull request. Some providers such as GitLab, Azure, Gitea can add labels to PRs in the same API call used for PR creation and it's unreasonable to make it as a separate call. For providers that do not support it, labelPullRequest() is kept as private member


fthomas (approved):
Looks great, thanks @danielleontiev!

PR #3007, opened 2023-03-06
#3006 Support for removing source branch on GitLab merge requests

### LINKED ISSUE #3006: Support for removing source branch on GitLab merge requests
We are currently using existing Gitlab flags (`gitlab-merge-when-pipeline-succeeds` and `gitlab-required-reviewers`) to merge successful merge request automatically. While this flow works great, there is a small nuisance - source branch hangs around after the merge request is merged. Luckily, there is a `remove_source_branch` parameter which we can pass as part of `create merge request` payload to remove source branches once the merge request is merged! 

I already added `remove_source_branch` flag and created a pull request. Happy to address any feedback. Let me know what you think! 

As side note, we already have the following box checked in Gitlab which is a repository level setting but it doesn't apply to merge requests created by scala-steward for some reason;

image

endertunc:
Fixes #3006 

Once we agree that this is a useful feature to add, I can update the relevant tests etc.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
@@ -238,7 +238,7 @@ final class GitLabApiAlg[F[_]: Parallel](
           res <-
             client
               .put[MergeRequestOut](
-                url.mergeWhenPiplineSucceeds(repo, mr.iid),
+                url.mergeWhenPiplineSucceeds(repo, mr.iid, gitLabCfg.removeSourceBranch),
                 modify(repo)
               )
               // it's possible that our status changed from can be merged already,
--- a/modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
@@ -238,7 +238,7 @@ final class GitLabApiAlg[F[_]: Parallel](
           res <-
             client
               .put[MergeRequestOut](
-                url.mergeWhenPiplineSucceeds(repo, mr.iid, gitLabCfg.removeSourceBranch),
+                url.mergeWhenPiplineSucceeds(repo, mr.iid),
                 modify(repo)
               )
               // it's possible that our status changed from can be merged already,
--- a/modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
@@ -40,7 +40,7 @@ final private[gitlab] case class MergeRequestPayload(
     assignee_ids: Option[List[Int]],
     reviewer_ids: Option[List[Int]],
     target_project_id: Long,
-    remove_source_branch: Boolean,
+    remove_source_branch: Option[Boolean],
     source_branch: String,
     target_branch: Branch
 )
@@ -63,7 +63,7 @@ private[gitlab] object MergeRequestPayload {
       reviewer_ids = Option.when(reviewers.nonEmpty)(reviewers),
       labels = Option.when(data.labels.nonEmpty)(data.labels),
       target_project_id = projectId,
-      remove_source_branch = removeSourceBranch,
+      remove_source_branch = Option.when(removeSourceBranch)(removeSourceBranch),
       source_branch = data.head,
       target_branch = data.base
     )
@@ -138,7 +138,9 @@ private[gitlab] object GitLabJsonCodec {
     }
 
   implicit val projectIdDecoder: Decoder[ProjectId] = deriveDecoder
-  implicit val mergeRequestPayloadEncoder: Encoder[MergeRequestPayload] = deriveEncoder
+  implicit val mergeRequestPayloadEncoder: Encoder[MergeRequestPayload] =
+    deriveEncoder[MergeRequestPayload].mapJson(_.dropNullValues)
+
   implicit val updateStateEncoder: Encoder[UpdateState] = Encoder.instance { newState =>
     val encoded = newState.state match {
       case PullRequestState.Open   => "open"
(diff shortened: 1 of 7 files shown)


fthomas:
👍 Looks useful to me.

Should the new field in `MergeRequestPayload` be optional so that older GitLab versions do not fail if an unknown field is present?

endertunc:
@fthomas thanks for your reply. I thought `Option[Boolean]` would look strange but I think it's a good idea to have it optional. I will update the pull request accordingly. 👍

endertunc:
Hi @fthomas, 

There isn't a magical way that is being currently used to drop Nulls from produces Json, right?

I know that circe has capabilities which you can configure (`_.asJson.deepDropNullValues` or custom Printer with `dropNullValues = true`) but as far as I can tell scala-steward uses the default one which keeps the null values. To drop `remove_source_branch` when it's None, I will need write the encoder by hand. Something like this;

```scala
  implicit val mergeRequestPayloadEncoder: Encoder[MergeRequestPayload] = Encoder.instance {
    mergeRequestPayload =>
      val payloadMap: Map[String, Json] = Map(
        "id" -> mergeRequestPayload.id.asJson,
        "title" -> mergeRequestPayload.title.asJson,
        "description" -> mergeRequestPayload.description.asJson
        ...
      )

      mergeRequestPayload.remove_source_branch
        .fold(payloadMap)(removeSourceBranch =>
          // Only add when it's defined
          payloadMap + ("remove_source_branch" -> Json.fromBoolean(removeSourceBranch))
        )
        .asJson
  }
```

Just wanted to double check we are on the same page and I didn't miss anything.

fthomas:
> There isn't a magical way that is being currently used to drop Nulls from produces Json, right?

That is correct.

> To drop remove_source_branch when it's None, I could do something like this but this will also affect other None fields in MergeRequestPayload as well;

I would assume that this is fine. These other optional fields are for reviewers and asignees, right? They have only been recently added, so I think it is fine to drop them if they are `null`.

modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
```diff
@@ -61,6 +63,7 @@ private[gitlab] object MergeRequestPayload {
       reviewer_ids = Option.when(reviewers.nonEmpty)(reviewers),
       labels = Option.when(data.labels.nonEmpty)(data.labels),
       target_project_id = projectId,
+      remove_source_branch = Option.when(removeSourceBranch)(removeSourceBranch),
```
endertunc, line 66: I am using _flag_ to read `removeSourceBranch` config we can't actually pass `remove_source_branch = false` but I guess that's ok. Alternative is to make CLI option `Option[Boolean]` something like this;

```scala
implicit val readBoolean: Argument[Boolean] =
  Argument.from("boolean") { string =>
    try
      Validated.valid(string.toBoolean)
    catch {
      case _: IllegalArgumentException => Validated.invalidNel(s"Invalid Boolean: $string")
    }
  }

private val gitlabRemoveSourceBranch_v2: Opts[Option[Boolean]] =
  option[Boolean](
    "gitlab-remove-source-branch",
    "Flag indicating if a merge request should remove the source branch when merging."
  ).orNone

```

Let me know what you think.

modules/core/src/test/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlgTest.scala
```diff
@@ -232,7 +268,7 @@ class GitLabApiAlgTest extends CatsEffectSuite with Http4sDsl[MockEff] {
   }
 
   test("createPullRequest -- auto merge") {
-    val prOut = gitlabApiAlgNoFork
+    val prOut = gitlabApiAlgAutoMerge
```
endertunc, line 271: This test case was identical to "createPullRequest -- no fork". I think the intention was to test payload with `mergeWhenPipelineSucceeds = true` so created `gitlabApiAlgAutoMerge` and used in here.

modules/core/src/test/scala/org/scalasteward/core/forge/gitlab/MergeRequestPayloadTest.scala
```diff
@@ -22,15 +22,13 @@ class MergeRequestPayloadTest extends FunSuite {
...
+    val obtained =
+      MergeRequestPayload(id, projectId, data, Map.empty, removeSourceBranch = false).asJson
     val expected =
       json"""{
                "id" : "123",
                "title" : "Test MR title",
                "description" : "Test MR body",
-               "labels" : null,
-               "assignee_ids" : null,
-               "reviewer_ids" : null,
```
endertunc, line 33: `mergeRequestPayloadEncoder` drops null values now thus keys with null values are removed.

modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/Url.scala
```diff
@@ -39,9 +39,17 @@ class Url(apiHost: Uri) {
...
+  def mergeWhenPiplineSucceeds(
+      repo: Repo,
+      number: PullRequestNumber,
+      removeSourceBranch: Boolean
+  ): Uri =
     (existingMergeRequest(repo, number) / "merge")
       .withQueryParam("merge_when_pipeline_succeeds", "true")
+      .withOptionQueryParam(
+        "should_remove_source_branch",
+        Option.when(removeSourceBranch)(removeSourceBranch)
```
endertunc, line 51: I also found this `should_remove_source_branch` can also be provided as part of `/merge` request so I added as well. When it's `false` query parameter won't be added.

endertunc: > Is the field in MergeRequestPayload not enough for the source branch to be removed after merge?

In theory, it should be but I haven't tested the real behavior to be honest. I found `should_remove_source_branch` last night and I believe it's just another way to remove source branch once merge request is merged.

I think I was too eager to add it but I agree it can be confusing. At the end of the they, if they are doing the same thing we should only have one of them.

I will verify the API behavior and let you know! 

Thanks for taking a look and sorry for the confusion!

endertunc: @fthomas I just reverted the commit where I added the `should_remove_source_branch` query parameter to  "/merge" request.

I tested that adding `remove_source_branch` to `MergeRequestPayload` is sufficient and there is no need for anything additional 👍

fthomas (approved):
LGTM. Thanks @endertunc!

PR #3009, opened 2023-03-13
Reformat changed files in all build roots

fthomas:
`ScalafmtAlg#reformatChanged` is working on the `Repo`-level but we're reading the Scalafmt version in `ScalafmtAlg#getScalafmtVersion` on the `BuildRoot`-level. This means we detect a dependency on Scalafmt in a build root but when we reformat changed files, we'll do it in the root directory of the repo. If the root directory contains no `.scalafmt.conf`, reformating fails with an error as shown in
https://github.com/scala-steward-org/scala-steward-action/issues/475#issuecomment-1466314269

With this change we now run Scalafmt in all build roots instead of only the root directory. This fixes the error mentioned above. In the future we could restrict running Scalafmt in only those build roots where we did changes by matching the `updateReplacements` (that have been determined in `EditAlg`) with the build roots.


mzuehlke (approved):

alejandrohdezma (approved):
🎉

PR #3014, opened 2023-03-19
update configuration settings and integrate new API calls to gitlab flow

endertunc:


modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
```diff
@@ -44,6 +46,21 @@ object Cli {
     val processTimeout = "process-timeout"
   }
 
+  implicit val mergeRequestApprovalsConfigArgument: Argument[MergeRequestApprovalRulesCfg] =
+    Argument.from("approvals_rule_name=required_approvals") { s =>
+      s.split(":").toList match {
```
endertunc, line 51: I split with `:` but I can switch to `=` like we do in env variables. I don't have any preferences.

modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
```diff
@@ -156,9 +156,11 @@ object Config {
   final case class GitHubCfg(
   ) extends ForgeSpecificCfg
 
+  final case class MergeRequestApprovalRulesCfg(approvalRuleName: String, requiredApprovals: Int)
+
   final case class GitLabCfg(
       mergeWhenPipelineSucceeds: Boolean,
-      requiredReviewers: Option[Int],
+      requiredApprovals: Option[Either[Int, Nel[MergeRequestApprovalRulesCfg]]],
```
endertunc, line 163: Here is how to read this;
Left side the old flow, user just provides the old configuration via `--gitlab-required-reviewers <integer>`
Right side is the new flow, user provides the new configuration via `--merge-request-level-approval-rule <approvals_rule_name=required_approvals>`

I also renamed the name of this variable. It's not reviewers, it's actually approvals. 

I can wrap left side to a value class. It might be more explanatory then just Int. WDYT?

modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
```diff
@@ -222,7 +244,13 @@ final class GitLabApiAlg[F[_]: Parallel](
         for {
           mr <- mergeRequest
           mrWithStatus <- waitForMergeRequestStatus(mr.iid)
-          _ <- maybeSetReviewers(repo, mrWithStatus)
+          _ <- gitLabCfg.requiredApprovals match {
+            case Some(Right(approvalRules)) =>
+              setApprovalRules(repo, mrWithStatus, approvalRules)
+            case Some(Left(requiredReviewers)) =>
+              setReviewers(repo, mrWithStatus, requiredReviewers)
+            case None => F.unit
```
endertunc, line 252: Here we call one flow or the other depending on the configuration provided.

endertunc:
I will cover the missing lines reported by Codecov.

modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
endertunc: `=` makes more sense I guess so I changed it already.

endertunc:
I added missing test cases related to the changes in this pull request.

endertunc:
@fthomas just following up on this one. Do you think this is worth merging?

As described in the ticket, implementation in main depends on deprecated API. However, I am not sure if there is enough interest from your side or community to have this PR merged.

I liked working on the issue. I learnt few things from the code base and I also learned few things from Gitlab API so I overall it was a nice experience for me and I wouldn't want to push forward if there is no interest. 

We can keep it around for a while and see if people show any interest. If you wish to close the PR now without merging it I am also fine with it :)

modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
henricook: @endertunc I'd suggest a `-gitlab` prefix for this new config param perhaps, for consistency?

docs/help.md
```diff
@@ -80,6 +80,8 @@ Options and flags:
         Whether to merge a gitlab merge request when the pipeline succeeds
     --gitlab-required-reviewers <integer>
         When set, the number of required reviewers for a merge request will be set to this number (non-negative integer).  Is only used in the context of gitlab-merge-when-pipeline-succeeds being enabled, and requires that the configured access token have the appropriate privileges.  Also requires a Gitlab Premium subscription.
+    --merge-request-level-approval-rule <approvals_rule_name=required_approvals>
+        Additional repo config file (can be used multiple times)
```
henricook, line 84: Is this description of the flag accurate?

modules/core/src/main/scala/org/scalasteward/core/forge/gitlab/GitLabApiAlg.scala
```diff
@@ -270,28 +298,86 @@ final class GitLabApiAlg[F[_]: Parallel](
...
     }
+  import cats.implicits._
 
-  private def maybeSetReviewers(repo: Repo, mrOut: MergeRequestOut): F[MergeRequestOut] =
-    gitLabCfg.requiredReviewers match {
-      case Some(requiredReviewers) =>
-        for {
-          _ <- logger.info(
-            s"Setting number of required reviewers on ${mrOut.webUrl} to $requiredReviewers"
+  private def setReviewers(
```
henricook, line 303: It might be worth adding a comment here reminding readers that this is retained only for Gitlab versions older than... 12.3 I think?

modules/core/src/test/scala/org/scalasteward/core/application/CliTest.scala
```diff
@@ -151,27 +155,88 @@ class CliTest extends FunSuite {
     assert(clue(obtained).startsWith("Usage"))
   }
 
-  test("parseArgs: non-default GitLab arguments") {
+  test("parseArgs: non-default GitLab arguments and required reviewers") {
     val params = minimumRequiredParams ++ List(
       List("--gitlab-merge-when-pipeline-succeeds"),
+      List("--gitlab-remove-source-branch"),
```
henricook, line 161: Is this a drive-by? Or a bad merge with main?

henricook:
I came here to do this work so that we can have approvers set by SS on Gitlab and was happy to find your MR @endertunc - I've thrown in a few review comments, hopefully that's a help. It'd be really great to have this after that @fthomas 🙏🏻 . Happy to help in any way I can.

modules/core/src/test/scala/org/scalasteward/core/application/CliTest.scala
endertunc: I think there were some missing test cases (config flags in this case) so I think I just did a bit of improvement here if I remember correctly.

PR #3015, opened 2023-03-20
Fixed retry and duplicated replacements

Seetaramayya:
For good amount of time our `scala-steward` jobs are failing. Following changes fixed in our case, so creating the PR.

#### New Retry logic is

- If `Retry-After` header present, then use the header value to retry (this was there before)
- otherwise retry request after time specified by `X-Ratelimit-Reset` header. Header value is UTC epoch seconds. Here I took some assumptions
     - if the difference between header value and current UTC epoch seconds is more than 60 seconds then I retry after `60 seconds`. That means max wait time would be `60 seconds`. 
     - if the difference is negative then I retry after 1 second. That means min wait time is `1 second` 
- if neither of the headers present, wait `1.second`

#### Another issue 

If there are duplicated replacements then the job was failing. This is relatively very easy fix, use `Set` instead of `List`

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Context.scala
@@ -115,7 +115,7 @@ object Context {
       userAgent <- Resource.eval(F.fromEither(`User-Agent`.parse(1)(userAgentString)))
       middleware = ClientConfiguration
         .setUserAgent[F](userAgent)
-        .andThen(ClientConfiguration.retryAfter[F](maxAttempts = 5))
+        .andThen(ClientConfiguration.withRetry[F](maxAttempts = 5))
       defaultClient <- ClientConfiguration.build(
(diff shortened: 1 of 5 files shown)


Seetaramayya:
image

fthomas:
> Another issue

This deserves its own PR. I would also prefer to have a test for that in [`RewriteTest`](https://github.com/scala-steward-org/scala-steward/blob/f1f2e5919fe043fe940d5909880475a863a6458b/modules/core/src/test/scala/org/scalasteward/core/edit/RewriteTest.scala) instead of in `EditAlgTest`.

Seetaramayya:
Another PR crated to address the other issue https://github.com/scala-steward-org/scala-steward/pull/3016

Seetaramayya:
Are you going to merge this PR?

odisseus:
In order to make this PR mergeable, you must resolve the conflicts first.

PR #3023, opened 2023-04-01
Remove duplication around `TraceEntry.Cmd` instances

fthomas:


mkurz:
I think you can now delete the `topic/reduce-cmd-duplication` branch:
https://github.com/scala-steward-org/scala-steward/branches

PR #3024, opened 2023-04-01
Don't install Coursier via `git.io/coursier-cli`

fthomas:
Recent builds failed intermittently because Coursier could not be installed in the Docker image via `git.io/coursier-cli` (see https://github.com/coursier/coursier/issues/2724). The [current Coursier installation instructions](https://get-coursier.io/docs/cli-installation) does not mention `git.io/coursier-cli` anymore and recommends to download the appropriate binary directly.

Coursier is now installed by downloading and extracting https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-linux-static.gz. The installation processes of all binaries has been equalized by using `curl` and a common structure.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -358,7 +358,7 @@ lazy val dockerSettings = Def.settings(
       Cmd("RUN", s"wget $sbtUrl && tar -xf $sbtTgz && rm -f $sbtTgz"),
       Cmd("RUN", s"curl -fL $millUrl > $millBin && chmod +x $millBin"),
       Cmd("RUN", installScalaCliStep),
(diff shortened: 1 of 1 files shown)


alejandrohdezma (approved):
Good catch! I noticed this in the action a few months ago but didn't realise we were doing the same here for the Docker image 😓

fthomas:
I probably saw that PR in the action but also didn't realized the process needs to be updated here too.

I'll tweak the installation a bit further by using the version from `Dependencies.coursierCore` for reproducibility and by renaming the binary to `cs` to match the installation instruction.

PR #3033, opened 2023-04-15
Add support for multiple configuration file locations

### LINKED ISSUE #2735: Fallback to `.github/.scala-steward.conf` if config not found in root
I think it's a good idea to fall back to the `.github` directory when loading the config.
It's getting more and more a convention to put things GitHub related in that folder. E.g. see our repo: https://github.com/playframework/playframework/tree/main/.github
~We will also move the `.mergify.yml` from the root to `.github` soon, so `.scala-steward.conf` is the last one missing.~ Done!

The change needs to be done in
* https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
* maybe here too (?): https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigLoader.scala
* Update docs: https://github.com/scala-steward-org/scala-steward/blob/main/docs/repo-specific-configuration.md (maybe there is more, not sure)

blaz-kranjc:
Fixes #2735

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -1,6 +1,8 @@
 # Repository-specific configuration
 
-You can add `<YOUR_REPO>/.scala-steward.conf` to configure how Scala Steward updates your repository.
+You can add a configuration file `.scala-steward.conf` to configure how Scala Steward updates your repository.
+The `.scala-steward.conf` configuration file can be located in the root of your repository, in `.github` directory or in `.config` directory (searched in this order).
+If a configuration file exists in more than one location, only the first found file is taken into account.
 
 ```properties
 # pullRequests.frequency allows to control how often or when Scala Steward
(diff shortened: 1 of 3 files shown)


mzuehlke (approved):
Sorry for ignoring your PR for so long.
Thank you for your contribution.

mkurz:
The `.github` and `.config` folder are both meant to be hidden (prefixed with a dot) and the `.scala-steward.conf` in the root folder is meant to be hidden as well.

Now, wouldn't it make sense to support a (non-hidden) `scala-steward.conf` within the `.github` and `.config` folder since both folder are hidden already?

E.g. look at how mergify does it (https://docs.mergify.com/configuration/)

> Mergify uses the configuration file that is:
>
> * The file named `.mergify.yml`, or, as a fallback, `.mergify/config.yml` or `.github/mergify.yml` from the root directory.

mkurz:
Opened #3115

PR #3071, opened 2023-06-06
Add GitHub Action Run Summary summarising failing repos

rtyley:
This change provides a GitHub Action Job Summary that is visible in the GitHub UI without digging into the workflow logs:

* https://github.blog/2022-05-09-supercharging-github-actions-with-job-summaries/
* https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#adding-a-job-summary

It will create a job summary that looks something like this:

image

When Scala Steward's GitHub Action is running a job over many repos, and one or more of those repos fails, Scala Steward will return a failure ExitCode, but determining _which_ of those repos has failed, and why, requires ponderously searching the large workflow job logs for strings like 'ERROR Steward xxx/yyy failed'.

https://github.com/scala-steward-org/scala-steward/assets/52038/89092032-451b-4a14-8076-cbe8eb976e11

A Scala Steward run can fail consistently for individual specific repos, without preventing it from succeeding on the other repos in the same run. When we see the big red danger cross reported against the workflow, this is often a concern solely for the team that owns that specific repo - not the owners of the many other repos that may have been in the run. Failure on a repo can be caused by a variety of causes, for instance:

* `org.scalasteward.core.forge.github.GitHubException$RepositoryArchived` - if the repo has been _archived_, but not removed from the list of repos for Scala Steward to process.
* Invalid sbt configuration in the target repo - for instance, obsolete resolvers, version conflicts. These are problems for that repo in general, not specific to Scala Steward - but still, not of direct interest to the owners of other repos in the organisation that are actually being successfully processed by Scala Steward.

Although this change is for the benefit of running Scala Steward in GitHub Actions, the best place for this change to be implemented is in the main Scala Steward executable (https://github.com/scala-steward-org/scala-steward), which has the necessary diagnostic information on the Scala Steward run, rather than in the Scala Steward GitHub Action itself (https://github.com/scala-steward-org/scala-steward-action), which is focused on _launching_ Scala Steward.


modules/core/src/main/scala/org/scalasteward/core/application/RunResults.scala
```diff
@@ -0,0 +1,49 @@
...
+ * limitations under the License.
+ */
+
+package org.scalasteward.core.application
+
+import cats.effect.ExitCode
+import cats.syntax.all._
+import org.scalasteward.core.data.Repo
+
+case class RunResults(results: List[Either[(Repo, Throwable), Repo]]) {
```
alejandrohdezma, line 23: **suggestion:** We could also create this class as a utility method (something like `GitHubSummary.from(results)`), since it is only used on one place for generating the summary.

alejandrohdezma (commented):
Hey @rtyley, thank you so much for this I really like the idea. I'm thinking it could make more sense to write the summary (or the results) to a generic place (like the app's workspace) and read it from the action there. That way other forges could benefit from this too. WDYT?

rtyley:
> Hey @rtyley, thank you so much for this I really like the idea. I'm thinking it could make more sense to write the summary (or the results) to a generic place (like the app's workspace) and read it from the action there. That way other forges could benefit from this too. WDYT?

Thanks @alejandrohdezma, good idea!

I've now implemented that with commit 4a42929bfae61393c3bc1ae3a5a98ae6a150d6f1 -  the file will be written to `run-summary.md` at the root of the Scala Steward workspace - this does mean there will need to be a subsequent PR in `scala-steward-action`, but it also simplifies this PR quite a bit! 👍

modules/core/src/main/scala/org/scalasteward/core/application/RunResults.scala
rtyley: It does have two outputs (`exitCode`, as well as `markdownSummary`), so maybe it makes sense to keep it?

https://github.com/scala-steward-org/scala-steward/blob/4a42929bfae61393c3bc1ae3a5a98ae6a150d6f1/modules/core/src/main/scala/org/scalasteward/core/application/StewardAlg.scala#L109-L113

rtyley:
> I recently came across https://github.com/jrudolph/sbt-hackers-digest maybe this functionality could help here, too.

Ah that's an interesting plugin! Sounds generally quite useful - for my own specific use-case I'm trying to address here, it might not be so applicable, as I've got Scala Stewards GitHub Action setup to run over many repos at once - I suspect it's not easy for a GitHub Action running in one repo (eg [guardian/scala-steward-public-repos](https://github.com/guardian/scala-steward-public-repos)) to [annotate](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-creating-an-annotation-for-an-error) files in other repos as it runs.

alejandrohdezma:
I really like the idea @laughedelic. I'll add to my to-do list to add it to the action's docs.

fthomas:
This is a cool feature! :clap: @rtyley

It could also be useful for the public instance @tgodzik is operating. If the run summary would be committed to https://github.com/VirtusLab/scala-steward-repos after every run, people could find out by themselves if Scala Steward has problems with their repos without asking @tgodzik to look at the logs.

PR #3076, opened 2023-06-09
Improve grouping feature

terryhendrix1990:
Solves 2 issues in the `grouping` feature:
1. Failed grouping PRs (using `Update.Group`) get stuck: When a grouping PR is created by Scalasteward and the PR fails to build, then scalasteward does not create a new PR when new versions are released, because an existing PR is still open and the branch names are not unique. Solution: Introduce unique branch names in the grouping feature. This is similar to `Update.Single`
2. Commit messages are unclear. Solution: Added artifactIds and versions to the commit message. This is similar to `Update.Single`

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/git/CommitMsg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/git/CommitMsg.scala
@@ -53,20 +53,41 @@ object CommitMsg {
         CommitMsg(title = title)
       },
       g => {
-        val artifactsWithVersions = g.updates
-          .groupBy(_.nextVersion)
-          .map { case (version, updates) =>
-            s"${updates.map(u => s"${u.artifactId.name}").mkString(", ")} to ${version.value}"
-          }
-          .mkString(" - ")
-
-        val defaultTitle =
-          s"Update for group ${g.name}${baseBranch.fold("")(branch => s" in ${branch.name}")}: $${artifactVersions}"
-
-        val title = g.title
-          .getOrElse(defaultTitle)
-          .replace("${artifactVersions}", artifactsWithVersions)
-        CommitMsg(title = title)
+        CommitMsg(title = groupMessage(group = g, baseBranch))
       }
     )
+
+  def groupMessage(group: Update.Grouped, baseBranch: Option[Branch]): String = {
+    def innerGroupMessage(shortNotation: Boolean) = {
+      val artifactsWithVersions = group.updates
+        .groupBy(_.nextVersion)
+        .map {
+          case (version, List(update)) =>
+            s"${update.artifactId.name} to ${version.value}"
+
+          case (version, updates) if shortNotation =>
+            val firstArtifactName = updates.headOption.map(_.artifactId.name).getOrElse("")
+            s"$firstArtifactName and ${updates.size - 1} more to ${version.value}"
+
+          case (version, updates) =>
+            s"${updates.map(_.artifactId.name).mkString(", ")} to ${version.value}"
+        }
+        .mkString(" - ")
+
+      val defaultTitle = group.title.getOrElse(s"Update for group ${group.name}") +
+        baseBranch.fold("")(branch => s" in ${branch.name}")
+
+      group.title
+        .getOrElse(defaultTitle)
+        .replace("${artifactVersions}", artifactsWithVersions)
+    }
+
+    val title = innerGroupMessage(shortNotation = false)
+
+    if (title.sizeIs <= 200) {
+      title
+    } else {
+      innerGroupMessage(true)
+    }
+  }
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/git/CommitMsg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/git/CommitMsg.scala
@@ -52,10 +52,21 @@ object CommitMsg {
           .replace("${branchName}", baseBranch.map(_.name).orEmpty)
         CommitMsg(title = title)
       },
-      g =>
-        CommitMsg(title =
-          g.title.getOrElse(s"Update for group ${g.name}") +
-            baseBranch.fold("")(branch => s" in ${branch.name}")
-        )
+      g => {
+        val artifactsWithVersions = g.updates
+          .groupBy(_.nextVersion)
+          .map { case (version, updates) =>
+            s"${updates.map(u => s"${u.artifactId.name}").mkString(", ")} to ${version.value}"
+          }
+          .mkString(" - ")
+
+        val defaultTitle =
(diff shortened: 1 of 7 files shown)


alejandrohdezma (changes requested):
Loved the idea! Left a comment with an improvement for configuration

modules/core/src/main/scala/org/scalasteward/core/git/package.scala
```diff
@@ -30,7 +30,10 @@ package object git {
     val base = baseBranch.fold("")(branch => s"${branch.name}/")
     update.on(
       update = u => Branch(s"$updateBranchPrefix/$base${u.name}-${u.nextVersion}"),
-      grouped = g => Branch(s"$updateBranchPrefix/$base${g.name}")
+      grouped = g => {
+        val uniqueness = Math.abs(g.updates.map(_.nextVersion).sortBy(_.value).hashCode())
```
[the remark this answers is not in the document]
terryhendrix1990: Not a bad idea, however I believe this should be a default because the behaviour is now similar to Update.Single. 
WDYT?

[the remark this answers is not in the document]
alejandrohdezma: 🤔 I don't think that should be the default value since it will mean we are breaking compatibility (people already using it won't have the same branch anymore).

Regarding how to add the configuration, adding more options will make the config more complicated, while adding a magic string is easy and something people are already familiar with on other parts of the configuration.

Also take into account this is something people can already do on their own, for example, if using GitHub Actions, you can add a super-simple step to update the configuration for you on every run (essentially creating a new value each time):

```yaml
on:
  schedule:
    - cron: '0 0 * * 0'

name: Launch Scala Steward

jobs:
  scala-steward:
    runs-on: ubuntu-22.04
    name: Launch Scala Steward
    steps:
      - name: Checkout project
         uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0

      - name: Update `.scala-steward.conf`
         run: timestamp=$(date +%s); sed -i '' -e "s/{{timestamp}}/$timestamp/g" .github/.scala-steward.conf

      - name: Launch Scala Steward
         uses: scala-steward-org/scala-steward-action@v2
         with:
           github-app-id: ${{ secrets.APP_ID }}
           github-app-installation-id: ${{ secrets.APP_INSTALLATION_ID }}
           github-app-key: ${{ secrets.APP_PRIVATE_KEY }}
```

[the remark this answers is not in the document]
terryhendrix1990: I see, alright, I'll add the `hash` variable to the grouping names.

modules/core/src/main/scala/org/scalasteward/core/git/CommitMsg.scala
```diff
@@ -52,10 +52,21 @@ object CommitMsg {
           .replace("${branchName}", baseBranch.map(_.name).orEmpty)
         CommitMsg(title = title)
       },
-      g =>
-        CommitMsg(title =
-          g.title.getOrElse(s"Update for group ${g.name}") +
-            baseBranch.fold("")(branch => s" in ${branch.name}")
-        )
+      g => {
+        val artifactsWithVersions = g.updates
```
alejandrohdezma, line 56: **issue:** This won't affect commit messages for individual updates, those are created with the previous function in this method (first method of `update.on`).

[These are the places where this method is called](https://github.com/search?q=repo%3Ascala-steward-org%2Fscala-steward+%22replaceVariables%22&type=code). As you can see, every call is done for an `Update.Single`, except for [this one](https://github.com/scala-steward-org/scala-steward/blob/5a643140dbbbf11548335997bfb139f13362a238/modules/core/src/main/scala/org/scalasteward/core/forge/data/NewPullRequestData.scala#L252) which creates the pull request title. This change will create PR titles appending updates information until GitHub truncates the title, which is something we don't want. So I would suggest reverting this part.

If commit messages aren't descriptive enough for you, maybe you can change it in your codebase using [`commits.message`](https://github.com/scala-steward-org/scala-steward/blob/609104a04cd536418ebcda553e7f48177a5264ae/docs/repo-specific-configuration.md?plain=1#L126) configuration.

alejandrohdezma (changes requested):

modules/core/src/main/scala/org/scalasteward/core/git/CommitMsg.scala
terryhendrix1990: Reverting this would no longer disclose the artifacts and versions for an update. Which is something I like to see.  After a squash merge, the individual commit messages are gone. 

I propose: Keep the default the same (`s"Update for group ${g.name}") +             baseBranch.fold("")(branch => s" in ${branch.name}"`), but still allow for `${artifactVersions}` to be specified in the grouping configuration through a variable.

Is that okay to you?

alejandrohdezma: Yes, I think that would be preferable, although this won't help much unless you have a very small group... since as I mention, GitHub will truncate the title

terryhendrix1990: I've not seen the title being truncated yet (in my ecosystem) However, I could do something in the lines of:
`Update artifact-1, ...., artifact-2 to version 5` to make the titles shorter, in case they are too long.

modules/core/src/main/scala/org/scalasteward/core/git/CommitMsg.scala
```diff
@@ -53,20 +53,41 @@ object CommitMsg {
...
-          .getOrElse(defaultTitle)
-          .replace("${artifactVersions}", artifactsWithVersions)
-        CommitMsg(title = title)
+        CommitMsg(title = groupMessage(group = g, baseBranch))
       }
     )
+
+  def groupMessage(group: Update.Grouped, baseBranch: Option[Branch]): String = {
+    def innerGroupMessage(shortNotation: Boolean) = {
+      val artifactsWithVersions = group.updates
```
alejandrohdezma, line 62: **question:** Could we add a test for this part?

terryhendrix1990: If you agree with an implementation like this: Of course :)

alejandrohdezma: oh yeah! I think is a good approach 😊

terryhendrix1990: I've added the tests.

alejandrohdezma (approved):
👏🏼 @terryhendrix1990 I'm sure lots of people will benefit from this change

PR #3083, opened 2023-06-14
Remove obsolete dependency on Mill

lefou:
Everthing Mill specific is handled by the mill-plugin.

We only need a version to download the Mill bootstrap script. Version `0.11.0-M11` is the last one still using Github Release pages, which is required as Scala-Steward needs to support Mill versions down to `0.6`.

Fix https://github.com/scala-steward-org/scala-steward/issues/3082

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -265,7 +265,6 @@ lazy val dummy = myCrossProject("dummy")
   .settings(noPublishSettings)
   .settings(
     libraryDependencies ++= Seq(
-      Dependencies.millMain,
       Dependencies.scalaStewardMillPlugin
     )
   )
--- a/build.sbt
+++ b/build.sbt
@@ -344,7 +344,7 @@ lazy val dockerSettings = Def.settings(
       s"tar -xf $sbtTgz",
       s"rm -f $sbtTgz"
     ).mkString(" && ")
-    val millVer = Dependencies.millVersion
+    val millVer = Dependencies.millScriptVersion
     val millBin = s"$binDir/mill"
     val installMill = Seq(
       s"$curl $millBin https://github.com/lihaoyi/mill/releases/download/${millVer.split("-").head}/$millVer",
--- a/build.sbt
+++ b/build.sbt
@@ -347,7 +347,7 @@ lazy val dockerSettings = Def.settings(
     val millVer = Dependencies.millScriptVersion
     val millBin = s"$binDir/mill"
     val installMill = Seq(
-      s"$curl $millBin https://github.com/lihaoyi/mill/releases/download/${millVer.split("-").head}/$millVer",
+      s"$curl $millBin https://github.com/lihaoyi/mill/releases/download/${millVer.split("-").takeWhile(!_.startsWith("M")).mkString("-")}/$millVer",
       s"chmod +x $millBin"
     ).mkString(" && ")
     val csBin = s"$binDir/cs"
--- a/build.sbt
+++ b/build.sbt
@@ -346,8 +346,12 @@ lazy val dockerSettings = Def.settings(
     ).mkString(" && ")
     val millVer = Dependencies.millScriptVersion
     val millBin = s"$binDir/mill"
+    val releasePageVersion = millVer.split("-") match {
+      case Array(v, m, _*) if m.startsWith("M") => s"${v}-${m}"
+      case Array(v, _*)                         => v
+    }
     val installMill = Seq(
-      s"$curl $millBin https://github.com/lihaoyi/mill/releases/download/${millVer.split("-").takeWhile(!_.startsWith("M")).mkString("-")}/$millVer",
+      s"$curl $millBin https://github.com/lihaoyi/mill/releases/download/${releasePageVersion}/$millVer",
       s"chmod +x $millBin"
     ).mkString(" && ")
     val csBin = s"$binDir/cs"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -34,6 +34,7 @@ object Dependencies {
   val jjwtApi = "io.jsonwebtoken" % "jjwt-api" % "0.11.5"
   val jjwtImpl = "io.jsonwebtoken" % "jjwt-impl" % jjwtApi.revision
   val jjwtJackson = "io.jsonwebtoken" % "jjwt-jackson" % jjwtApi.revision
+  val millScriptVersion = "0.11.0"
   val monocleCore = "dev.optics" %% "monocle-core" % "3.2.0"
   val munit = "org.scalameta" %% "munit" % "0.7.29"
   val munitCatsEffect = "org.typelevel" %% "munit-cats-effect-3" % "1.0.7"
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -34,8 +34,6 @@ object Dependencies {
   val jjwtApi = "io.jsonwebtoken" % "jjwt-api" % "0.11.5"
   val jjwtImpl = "io.jsonwebtoken" % "jjwt-impl" % jjwtApi.revision
   val jjwtJackson = "io.jsonwebtoken" % "jjwt-jackson" % jjwtApi.revision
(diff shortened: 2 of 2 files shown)


build.sbt
```diff
@@ -345,7 +344,7 @@ lazy val dockerSettings = Def.settings(
       s"tar -xf $sbtTgz",
       s"rm -f $sbtTgz"
     ).mkString(" && ")
-    val millVer = Dependencies.millVersion
+    val millVer = Dependencies.millScriptVersion
     val millBin = s"$binDir/mill"
     val installMill = Seq(
       s"$curl $millBin https://github.com/lihaoyi/mill/releases/download/${millVer.split("-").head}/$millVer",
```
daddykotex, line 350: you'll want to update that part, because `https://github.com/lihaoyi/mill/releases/download/0.11.0/0.11.0-M11` is wrong

daddykotex:
I tried that and `https://github.com/lihaoyi/mill/releases/download/0.11.0-M11/0.11.0-M11` uses maven, 0.11.0-M10 seems fine though

lefou:
> I tried that and `https://github.com/lihaoyi/mill/releases/download/0.11.0-M11/0.11.0-M11` uses maven, 0.11.0-M10 seems fine though

Looks like we got fooled by the mutable nature of Github Release, which is the reason why we changed to Maven in the first place. Those `mill` scripts were change later and do not represent the git tags. 🤔

daddykotex:
ahh that makes sense

in any case, it still does not work because the old script will append `-assembly` which is not there for 0.11.0:

https://github.com/lihaoyi/mill/releases/download/0.11.0/0.11.0-assembly

daddykotex:
maybe we should just use your `millw` wrapper?

daddykotex:
oh there is an assembly now :o

daddykotex:
Ok so it's my bad, I saw a failure and concluded it was just as broken but it was wrong
using the 0.10 launcher, we download mill and run it, but then it fails to resolve the plugin: 
```
show Cannot resolve external module org.scalasteward.mill.plugin.StewardPlugin
	at org.scalasteward.core.io.process$.$anonfun$slurp$7(process.scala:75)
	at org.scalasteward.core.io.process$.$anonfun$slurp$7$adapted(process.scala:69)
```

lefou:
> Ok so it's my bad, I saw a failure and concluded it was just as broken but it was wrong using the 0.10 launcher, we download mill and run it, but then it fails to resolve the plugin:
> 
> ```
> show Cannot resolve external module org.scalasteward.mill.plugin.StewardPlugin
> 	at org.scalasteward.core.io.process$.$anonfun$slurp$7(process.scala:75)
> 	at org.scalasteward.core.io.process$.$anonfun$slurp$7$adapted(process.scala:69)
> ```

Where is that output from? The current CI runs seem successful to me. I double checked that we use the latest `mill-plugin` version `0.18.0`, which also supports Mill `0.11`.

lefou:
> oh there is an assembly now :o

It was there the whole time. But sometimes, fetching from Github can time out. That's one of the other reasons we switched over to Maven Central.

fthomas:
> The current CI runs seem successful to me.

Note that the CI in this project only ensures that a `mill` executable is available in the Docker image. It does not test if that `mill` can be used by a Scala Steward instance.

daddykotex:
The errors I get are from running the docker image build w/ your changes, I've got the following issue:

1. `https://github.com/lihaoyi/mill/releases/download/0.11.0-M11/0.11.0-M11` is wrong because it still refers to maven (that's the URL that's generated when building the docker images with 70097702c0957bcd960bfcaddbdcb03a393d723a)
2. when I change the version to target 0.11.0.M10, then mill start correctly in both `0.10.12` projects and `0.11.0` projects

but then it fails to run the `scala-steward.sc` script because it can't pull the plugin

I can reproduce locally, remove any cached downloaded version, then call the mill launcher:

```bash
> mill --version
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 63.8M  100 63.8M    0     0  45.2M      0  0:00:01  0:00:01 --:--:-- 61.0M
Mill Build Tool version 0.11.0-M9
Java version: 17.0.2, vendor: Eclipse Adoptium, runtime: /Users/David.Francoeur/jvm/temurin/jdk-17.0.2-8/Contents/Home
Default locale: en_CA, platform encoding: UTF-8
OS name: "Mac OS X", version: 13.4, arch: aarch64
```

Then hop into a project where 0.11.0 is defined in the `.mill-version`:

```bash
> mill --version
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 69.7M  100 69.7M    0     0  51.5M      0  0:00:01  0:00:01 --:--:-- 79.4M
Mill Build Tool version 0.11.0
Java version: 17.0.2, vendor: Eclipse Adoptium, runtime: /Users/David.Francoeur/jvm/temurin/jdk-17.0.2-8/Contents/Home
Default locale: en_CA, platform encoding: UTF-8
OS name: "Mac OS X", version: 13.4, arch: aarch64
```

Now mill works in project: `mill __.compile` is good.

But if I introduce the `scala-steward.sc`, 

```bash
> cat scala-steward.sc
import $ivy.`org.scala-steward::scala-steward-mill-plugin_mill$MILL_BIN_PLATFORM:0.18.0`

```

If I run:
```bash
> mill -i -p scala-steward.sc show org.scalasteward.mill.plugin.StewardPlugin/extractDeps
[build.sc] [41/49] compile 
[info] compiling 1 Scala source to /Users/David.Francoeur/workspace/disney/smithy4s-docs/out/mill-build/compile.dest/classes ...
[info] done compiling
[1/1] show 
1 targets failed
show Cannot resolve external module org.scalasteward.mill.plugin.StewardPlugin
```

lefou:
`mill --prefef` is non-functional in Mill 0.11. The option was overlooked in 0.11.0 but is removed now.

The preferred way to include a plugin that only provides external modules is to use `--import`.

lefou:
I changed to version to 0.11.0-M10, which should work for all previous versions. 

Please note that this PR will not fix running Mill 0.11.0, which is already broken in `main` branch, as its scope is to remove an orphan dependency as well as restoring support for older Mill versions.

A proper fix to support Mill 0.11.0 needs to use the `--import` CLI option instead of the predef script.

daddykotex:
That's good for me, I'm actually working on a fix for 0.11, I will open another PR

daddykotex (approved):

daddykotex:
> `mill --predef` is non-functional in Mill 0.11. The option was overlooked in 0.11.0 but is removed now.
> 
> The preferred way to include a plugin that only provides external modules is to use `--import`.

So I'm trying to run `mill  --import 'org.scala-steward::scala-steward-mill-plugin_mill:0.18.0' show org.scalasteward.mill.plugin.StewardPlugin/extractDeps`

I tried different variants:
1. with 2_13 w/o the double `::` `org.scala-steward:scala-steward-mill-plugin_mill0.11_2.13:0.18.0`
2. with `0.11` suffix `org.scala-steward::scala-steward-mill-plugin_mill0.11:0.18.0`

etc and none seem to work

daddykotex:
ahhh this way: `ivy:org.scala-steward::scala-steward-mill-plugin::0.18.0`

lefou:
> ahhh this way: `ivy:org.scala-steward::scala-steward-mill-plugin::0.18.0`

Exactly. See https://mill-build.com/mill/Using_Plugins.html#_use_the_specific_mill_binary_platform_notation

PR #3086, opened 2023-06-16
Use `Option.when(cond)(value)` where appropriate

rtyley:
The [`Option.when(cond)(value)`](https://www.scala-lang.org/api/2.13.x/scala/Option$.html#when[A](cond:Boolean)(a:=%3EA):Option[A]) Option builder was added to Scala 2.13 with https://github.com/scala/scala/pull/5995, and it's a nice, slightly more concise alternative to using if/else, eg: `if (cond) Some(value) else None`.

In this change, I've just updated the Scala Steward codebase to use `Option.when` where it's appropriate.


alejandrohdezma (approved):
Great change! Thanks @rtyley! ❤️

PR #3088, opened 2023-06-16
Remove any run-specific files before starting run

rtyley:
To ensure that its workspace is clean, and that unintended files are not held over from previous runs, Scala Steward already erases the `repos` folder before starting its run. With new `run-summary.md` file introduced by https://github.com/scala-steward-org/scala-steward/pull/3071, we again want to be careful that the summary file from a previous run isn't picked up in a subsequent run.

https://github.com/scala-steward-org/scala-steward-action/pull/503 addressed this for the particular concern of GitHub Actions, by ensuring that the file wasn't persisted to cache, but @fthomas [pointed out](https://github.com/scala-steward-org/scala-steward-action/pull/503#issuecomment-1594477950) that this additional point at the start of Scala Steward's code was also a good place for removing any run-specific files.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
@@ -17,7 +17,7 @@
 package org.scalasteward.core.io
 
 import better.files.File
-import cats.FlatMap
+import cats.Monad
 import cats.syntax.all._
 import org.scalasteward.core.application.Config
 import org.scalasteward.core.buildtool.BuildRoot
@@ -25,9 +25,11 @@ import org.scalasteward.core.data.Repo
 import org.typelevel.log4cats.Logger
 
 trait WorkspaceAlg[F[_]] {
-  def cleanReposDir: F[Unit]
+  def removeAnyRunSpecificFiles: F[Unit]
 
   def rootDir: F[File]
+
+  def runSummary: F[File]
 
   def repoDir(repo: Repo): F[File]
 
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
```diff
@@ -17,45 +17,63 @@
...
 import org.scalasteward.core.data.Repo
 import org.typelevel.log4cats.Logger
 
 trait WorkspaceAlg[F[_]] {
-  def cleanReposDir: F[Unit]
+  def removeAnyRunSpecificFiles: F[Unit]
 
   def rootDir: F[File]
 
+  def runSummary: F[File]
```
mzuehlke, line 32: the directories have a `Dir` suffix. So `runsummaryFile` would align better.

modules/core/src/main/scala/org/scalasteward/core/io/WorkspaceAlg.scala
```diff
@@ -17,45 +17,63 @@
...
       fileAlg: FileAlg[F],
       logger: Logger[F],
-      F: FlatMap[F]
+      F: Monad[F]
   ): WorkspaceAlg[F] =
     new WorkspaceAlg[F] {
       private val reposDir: File =
         config.workspace / "repos"
 
+      private val _runSummary: File =
```
mzuehlke, line 52: As far as I know we use nowhere the `_` prefix. If you rename the method to `runSummaryFile` the name without the underscore would work.

mzuehlke (approved):
Good. Only some name nit picking 😄

rtyley:
> Good. Only some name nit picking 😄

Thanks! Done with 72dd9143.

PR #3101, opened 2023-06-23
Ensure `mockRoot` test folder is always cleared before tests run

rtyley:
Several of Scala Steward's tests (eg `PullRequestRepositoryTest`) write files to the filesystem, using `MockConfig.mockRoot` as the base folder.

https://github.com/scala-steward-org/scala-steward/blob/b1de3ab58194f543f3526e013b363a52c675c769/modules/core/src/test/scala/org/scalasteward/core/mock/MockConfig.scala#L8

However, the folder is not erased between runs, and (possibly depending on operating system or Java version) the folder _can be_ the same path for multiple runs. In that case, tests may pass first time, but fail the second (this is the case for `PullRequestRepositoryTest`), as unexpected old files are encountered.

Here's a video showing that on my machine (Java 17.0.4, macOS 12.6):

https://github.com/scala-steward-org/scala-steward/assets/52038/bd5a1317-3828-460f-8b79-936645e97e64

## Fix

This change alters the way `mockRoot` is constructed:

* OLD: `better.files.File.temp` - this was simply the value of [`java.io.tmpdir`](https://stackoverflow.com/a/1924159/438886), which could be simply `/tmp/`, or some other value, but not guaranteed to change between runs.
* NEW: `better.files.File.newTemporaryDirectory()` - uses [`java.nio.file.Files.createTempDirectory()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Files.html#createTempDirectory(java.lang.String,java.nio.file.attribute.FileAttribute...)), guaranteed to be a new folder each time.

...with the new method, the tests consistently pass when I run `PullRequestRepositoryTest`, whereas they would previously consistently fail, after passing once, before the change. Note that typically, where ever the temporary folders are created, they are deleted on reboot, so there's a low chance of them taking up excessive amounts of space.

### DIFF
--- a/modules/core/src/test/scala/org/scalasteward/core/mock/MockConfig.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/mock/MockConfig.scala
@@ -6,6 +6,7 @@ import org.scalasteward.core.application.{Cli, Config}
 
 object MockConfig {
   val mockRoot: File = File.temp / "scala-steward"
+  mockRoot.delete(true) // Ensure folder is cleared of previous test files
   private val args: List[String] = List(
     s"--workspace=$mockRoot/workspace",
     s"--repos-file=$mockRoot/repos.md",
--- a/build.sbt
+++ b/build.sbt
@@ -212,8 +212,6 @@ lazy val core = myCrossProject("core")
     // Uncomment for remote debugging:
     // run / javaOptions += "-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=5005",
     Test / fork := true,
-    Test / testOptions +=
-      Tests.Cleanup(() => Path(file(Properties.tmpDir) / "scala-steward").deleteRecursively()),
     Compile / resourceGenerators += Def.task {
       val outDir = (Compile / resourceManaged).value
       def downloadPlugin(v: String): File = {


fthomas:
There is some kind of cleanup in the sbt build: https://github.com/scala-steward-org/scala-steward/blob/b1de3ab58194f543f3526e013b363a52c675c769/build.sbt#L215-L216

Too bad this is ignored by IntelliJ.

alejandrohdezma:
We could also change the config to be created within an MUnit Fixture, so it can be properly teardown. WDYT @fthomas?

modules/core/src/test/scala/org/scalasteward/core/mock/MockConfig.scala
```diff
@@ -5,7 +5,7 @@ import org.scalasteward.core.application.Cli.ParseResult.Success
 import org.scalasteward.core.application.{Cli, Config}
 
 object MockConfig {
-  val mockRoot: File = File.temp / "scala-steward"
+  val mockRoot: File = File.newTemporaryDirectory() / "scala-steward"
```
rtyley, line 8: > There is some kind of cleanup in the sbt build:
> 
> Too bad this is ignored by IntelliJ.

Ah, interesting, that's two things I wasn't aware of- that there was test cleanup code in Scala Steward's build.sbt, and also that IntelliJ wouldn't _run_ that test cleanup code (I've reported that as [SCL-21353](https://youtrack.jetbrains.com/issue/SCL-21353/Scala-IntelliJ-test-runner-doesnt-perform-sbt-Tests.Setup-or-Tests.Cleanup-defined-in-testOptions?s=Scala-IntelliJ-test-runner-doesnt-perform-sbt-Tests.Setup-or-Tests.Cleanup-defined-in-testOptions) against IntelliJ). I think the _discoverability_ of those two bits of information is an issue for newcomers trying to contribute to this project - how can we help them?

There are a couple of options for helping out newcomers here:

### A. Inform IntelliJ users that they _must_ run the tests via `sbt`

Although the information could be added to [CONTRIBUTING.md](https://github.com/scala-steward-org/scala-steward/blob/main/docs/CONTRIBUTING.md), it probably stands a better chance of attracting the developers attention if it's displayed at the most pertinent point, ie when the developer runs the tests, like this: 

```suggestion
  val mockRoot: File = File.temp / "scala-steward"
  require(mockRoot.isEmpty,
    s"""Please ensure that the following folder is *empty*
       |(note that build.sbt `testOptions` is set to auto-cleanup contents, but
       |running tests in IDEs like IntelliJ may not execute the `CleanUp`):
       |$mockRoot""".stripMargin
  )
```

### B. Change test clean-up so IntelliJ works without contributors needing to adapt

> We could also change the config to be created within an MUnit Fixture, so it can be properly teardown.

Personally, that sounds great to me - an even smaller change (than an MUnit fixture) would be to add 1 line to perform the deletion in `MockConfig`: 

```suggestion
  val mockRoot: File = File.temp / "scala-steward"
  mockRoot.delete() // Ensure folder is cleared of previous test files
```

How does that sound to you?

fthomas: 👍 for plan B. I guess we could then remove the cleanup in the build too?

rtyley: Yeah, sounds good - I've amended the commit to go for option B, and also removed the corresponding sbt configuration 👍  - new commit is f44e79df92008a39c59b6b3627b686ee1683b8bd.

fthomas (approved):
LGTM. Thanks, @rtyley!

PR #3103, opened 2023-06-27
Use actual default values in `.scala-steward.conf` doc

fthomas:
This generates `docs/repo-specific-configuration.md` from a mdoc source file such that the default values referenced in that documentation are always in sync with the actual default values in the source code.


alejandrohdezma (approved):
👏🏼 👏🏼

mzuehlke (approved):
🎉

PR #3111, opened 2023-07-02
Ensure PR bodies and titles are recreated when updating PRs

### LINKED ISSUE #3110: Scala Steward won't update body of PR if new dependency added, or tracked versions changed
Hi, I'm not sure if this is a me problem or not, i'm running latest scala steward periodically against a number of repos, my config is as follows:

```
updates.fileExtensions = [".scala", ".sbt", ".sbt.shared", ".sc", ".yml", ".md", ".markdown", ".txt", ".properties"]
updatePullRequests = "always"
pullRequests.grouping = [
  { name = "all", "title" = "[SCALA STEWARD] Dependency updates", "filter" = [{"group" = "*"}] }
]
```

When it initially creates the PR, it's well formed and the raised PR commits line up with the body of the PR: 
image
and commits:
image

And that works fine, however if the PR is not merged for a while and maybe a new dependency is added, or an existing dependency changes version number, such as here: 
image
I can confirm that it adds the new version as a commit, but it doesn't seem to update the main PR body, it stays the same as pictured above. I also can't see any `edited` flags against the main PR, i can't see anything in the logs to indicate whether or not this is intended behaviour.

Is this intentional, am i missing something or is this a bug?

Also as an aside, another question: when steward is re-ran against a repo with an already open PR using the grouping functionality, even if the version numbers it wants to update to do not change, it will still revert all PR numbers and generate new commits, this works fine and is a non issue when using squash and merge but results in potentially very many commits being added to the PR and is more bloat than required, is this a bug?

For example:
image

Is then immediately followed by 
image
Where the version numbers of the update are identical, ideally this would only happen when a version number has changed.

Thanks again

alejandrohdezma:
## 💻 How to review this PR?

This PR was created with the idea of being reviewed commit by commit. Each commit contains an incremental change that makes it easier to review. Also some of the commits contain additional information in their description to help understand why the change was made.

<details><summary><b>I also recommend checking "Hide whitespace" when reviewing this PR!</b></summary></br>

</details>

## 🚀 What's included in this PR?

This PR includes a new method in `ForgeApiAlg` for updating a PR details. This method will be used when updating a PR to ensure new updates (for example, when using grouping feature) are correctly detailed in the PR body/title.

Currently this is only supported for GitHub forge. Other forges will be added in subsequent PRs.

## References

This PR fixes #3110.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/forge/ForgeApiAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/forge/ForgeApiAlg.scala
@@ -27,6 +27,12 @@ trait ForgeApiAlg[F[_]] {
 
   def createPullRequest(repo: Repo, data: NewPullRequestData): F[PullRequestOut]
 
+  def updatePullRequest(
+      number: PullRequestNumber,
+      repo: Repo,
+      data: NewPullRequestData
+  ): F[PullRequestOut]
+
   def closePullRequest(repo: Repo, number: PullRequestNumber): F[PullRequestOut]
 
   def getBranch(repo: Repo, branch: Branch): F[BranchOut]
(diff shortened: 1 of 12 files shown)


exoego (approved):
Nice

PR #3115, opened 2023-07-10
Support `scala-steward.conf` (makes sense in `.github` and `.config` folder)

mkurz:
Without the leading dot (not hidden),

Now that the config can be located inside the already hidden `.github` and `.config` folder (#3033) there is not need for the config file to be hidden as well inside that folders.
To stay backward compatible a hidden config `.scala-steward.conf` will always be preferred, only when such a hidden file can't be found anywhere, only then we try to read the "unhidden" `scala-steward.conf` in the same search order (root, `.github`, `.config` folders)

See https://github.com/scala-steward-org/scala-steward/pull/3033#issuecomment-1629376018

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
@@ -43,7 +43,7 @@ final class RepoConfigAlg[F[_]](maybeGlobalRepoConfig: Option[RepoConfig])(impli
       )(readRepoConfigFromFile(_))
       _ <- configParsingResult.fold(
         F.unit,
-        error => logger.info(s"Failed to parse $repoConfigBasename: ${error.getMessage}"),
+        error => logger.info(s"Failed to parse $activeConfigFile: ${error.getMessage}"),
         repoConfig => logger.info(s"Parsed repo config ${repoConfig.show}")
       )
     } yield configParsingResult
@@ -86,8 +86,10 @@ object RepoConfigAlg {
   private def activeConfigFile[F[_]](
       repoDir: File
   )(implicit fileAlg: FileAlg[F], logger: Logger[F], F: Monad[F]): F[Option[File]] = {
-    val configFileCandidates: F[List[File]] = repoConfigFileSearchPath
-      .map(_ :+ repoConfigBasename)
+    val configFileCandidates: F[List[File]] = (repoConfigFileSearchPath
+      .map(_ :+ repoConfigBasename) ++
+      repoConfigFileSearchPath
+        .map(_ :+ repoConfigBasename.substring(1)))
       .map(path => path.foldLeft(repoDir)(_ / _))
       .filterA(fileAlg.isRegularFile)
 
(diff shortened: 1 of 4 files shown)


modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
```diff
@@ -86,8 +86,10 @@ object RepoConfigAlg {
   private def activeConfigFile[F[_]](
       repoDir: File
   )(implicit fileAlg: FileAlg[F], logger: Logger[F], F: Monad[F]): F[Option[File]] = {
-    val configFileCandidates: F[List[File]] = repoConfigFileSearchPath
-      .map(_ :+ repoConfigBasename)
+    val configFileCandidates: F[List[File]] = (repoConfigFileSearchPath
+      .map(_ :+ repoConfigBasename) ++
+      repoConfigFileSearchPath
+        .map(_ :+ repoConfigBasename.substring(1)))
```
mkurz, line 92: The fix is here: The `.substring(1)` removes the leading dot from the filename.
As you can see I just append the possible non-hidden config file locations to the end of the list, which later will be searched for the first existing config file.

modules/core/src/test/scala/org/scalasteward/core/repoconfig/RepoConfigAlgTest.scala
```diff
@@ -278,7 +278,21 @@ class RepoConfigAlgTest extends FunSuite {
...
+    val rootConfigFile = repoDir / "scala-steward.conf"
+    val dotGithubConfigFile = repoDir / ".github" / "scala-steward.conf"
+    val initialState = MockState.empty.addFiles(dotGithubConfigFile -> "").unsafeRunSync()
+    val config = repoConfigAlg.readRepoConfig(repo).runA(initialState).unsafeRunSync()
+
+    assert(!fileAlg.isRegularFile(rootConfigFile).unsafeRunSync())
+    assert(fileAlg.isRegularFile(dotGithubConfigFile).unsafeRunSync())
+
+    assert(config.maybeRepoConfig.isDefined)
+  }
```
mkurz, line 293: The tests I added a basically just duplicates of the preceding test, just with the dot removed from the filename.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
blaz-kranjc: Would it make sense to change the repoConfigBasename to be without the `.` instead, and then add the dot here? substring use here adds a bit of a mental overhead.

PS: not associated with the project, but got a notification because of my PR and had to look at the code! Feel free to ignore :)

mkurz: The `repoConfigBasename` variable is publicly accessible therefore I don't want to change it's value for backwards compatibility.

blaz-kranjc: makes sense!

mzuehlke: In line 46 of this file it maybe good to use the actual filename (`activeConfigFile`) instead of `repoConfigBasename`

mzuehlke (approved):
Small comment for a better exception message.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfigAlg.scala
mkurz: @mzuehlke Done and pushed, thanks!

PR #3133, opened 2023-07-31
Run Scalafix build migrations using sbt

DavidGregory084:
Disclaimer: I have not tested this on a real build yet!

This PR changes the `SbtAlg` to make use of sbt to run Scalafix migrations instead of the Scalafix CLI.

It does this by writing the required sbt configuration files recursively up to the deepest meta-build found in the project.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
@@ -25,15 +25,15 @@ import org.scalasteward.core.buildtool.sbt.command._
 import org.scalasteward.core.buildtool.{BuildRoot, BuildToolAlg}
 import org.scalasteward.core.coursier.VersionsCache
 import org.scalasteward.core.data.{Dependency, Scope, Version}
-import org.scalasteward.core.edit.scalafix.{ScalafixCli, ScalafixMigration}
+import org.scalasteward.core.edit.scalafix.ScalafixMigration
 import org.scalasteward.core.io.process.SlurpOptions
 import org.scalasteward.core.io.{FileAlg, FileData, ProcessAlg, WorkspaceAlg}
 import org.scalasteward.core.util.Nel
+import org.scalasteward.core.buildtool.sbt.scalaStewardSbtScalafix
 
 final class SbtAlg[F[_]](config: Config)(implicit
     fileAlg: FileAlg[F],
     processAlg: ProcessAlg[F],
-    scalafixCli: ScalafixCli[F],
     workspaceAlg: WorkspaceAlg[F],
     versionsCache: VersionsCache[F],
     F: Concurrent[F]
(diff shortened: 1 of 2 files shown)


DavidGregory084:
@fthomas is it possible to dry run Scala Steward (i.e. not open any PRs)? I'd like to test this out but I don't really want to open PRs against other folks' projects.

DavidGregory084:
[This test](https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/test/scala/org/scalasteward/core/edit/EditAlgTest.scala#L138) no longer makes a lot of sense in light of this change, as it's testing for the usage of the Scalafix CLI - can I simply remove that test or would you prefer to maintain some kind of testing of build migrations at the level of `EditAlg`?

fthomas:
> is it possible to dry run Scala Steward (i.e. not open any PRs)?

There is some kind of dry run (https://github.com/scala-steward-org/scala-steward/pull/1828) but that would also not run any Scalafix migrations. You could temporarily comment [this line](https://github.com/scala-steward-org/scala-steward/blob/ea94b76ed633b10bc41936b3e9f0f2308ddaa3db/modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala#L175) to not push any commits and not creating PRs.

fthomas:
What I also use sometimes for testing is a test repo (like https://github.com/scala-steward-org/test-repo-2) with multiple branches. If Scala Steward creates a PR for a specific branch and I want to test my changes again, I just push another branch (without Scala Steward's change) to the repo and point Scala Steward to that new branch.

DavidGregory084:
OK, I've tested this now on a few projects with some mixed results:

* In spotify/scio [the sbt-tpolecat migration ran successfully](https://github.com/spotify/scio/compare/main...DavidGregory084:scio:update/sbt-tpolecat-0.5.0), although it looks like I have some more work to do on the migration rules as wildcard imports from some of the symbols are not being migrated. Gosh how I hate wildcard imports! They make understanding code difficult and it turns out they make migrating it difficult too.
* In some builds with older (<1.5.x) sbt, I was expecting `in` to be translated in the main `build.sbt`, but it didn't seem to work in at least [uclid-org/uclid](https://github.com/uclid-org/uclid/compare/master...DavidGregory084:uclid:update/sbt-1.9.3) and [sdkman/sdkman-candidates](https://github.com/sdkman/sdkman-candidates/compare/master...DavidGregory084:sdkman-candidates:update/sbt-1.9.3). I wonder if this implies that build migrations need to run at the root level to touch `build.sbt`, in which case there would actually be no difference between `sources` and `build` migration logic except that `sources` migrations would only operate at buildDepth = 0.

DavidGregory084:
>  it looks like I have some more work to do on the migration rules 

I have done some work on this in [sbt-tpolecat](https://github.com/typelevel/sbt-tpolecat/compare/v0.5.0..HEAD) now and will send a PR to update the migration SHA in scala-steward

> In some builds with older (<1.5.x) sbt, I was expecting `in` to be migrated to `/` in the main `build.sbt`, but it didn't seem to work

I have asked the Scalafix experts in the Scalameta Discord about this 🤞 that they have some idea how to fix it

DavidGregory084:
OK, so the word is that `sbt-scalafix` can't currently update `*.sbt` files because sbt doesn't return those as part of the build sources, and semantic migrations that update `*.sbt` files aren't supported and have never worked as it's so hard to generate SemanticDB for `*.sbt` files.

That means we need to continue running syntactic migrations on `*.sbt` files with the CLI, perhaps also disabling semantic migrations in the CLI invocations as they won't work in any case.

fthomas:
> OK, so the word is that `sbt-scalafix` can't currently update `*.sbt` files because sbt doesn't return those as part of the build's `unmanagedSources`, and semantic migrations that update `*.sbt` files aren't supported and have never worked as it's so hard to generate SemanticDB for `*.sbt` files.

:-(

> That means we need to continue running syntactic migrations on `*.sbt` files with the CLI, perhaps also disabling semantic migrations in the CLI invocations as they won't work in any case.

You mean via the [`--syntactic`](https://scalacenter.github.io/scalafix/docs/users/installation.html#help) option? Would that have made a difference for the sbt-tpolecat migration or would it just have not run the migration?

DavidGregory084:
> You mean via the [`--syntactic`](https://scalacenter.github.io/scalafix/docs/users/installation.html#help) option? Would that have made a difference for the sbt-tpolecat migration or would it just have not run the migration?

I think it would just have not run the migration, avoiding the errors due to missing semanticdb

modules/core/src/main/scala/org/scalasteward/core/io/FileAlg.scala
```diff
@@ -136,7 +136,13 @@ object FileAlg {
...
 
       override def walk(dir: File, maxDepth: Int): Stream[F, File] =
-        Stream.eval(F.delay(dir.walk(maxDepth))).flatMap(Stream.fromBlockingIterator(_, 1))
+        Stream.force(
+          F.blocking(dir.exists)
+            .ifF(
+              Stream.eval(F.delay(dir.walk(maxDepth))).flatMap(Stream.fromBlockingIterator(_, 1)),
+              Stream.empty.covary[F]
+            )
+        )
```
DavidGregory084, line 145: It seems that `Files.walk` throws an exception if the directory does not exist rather than simply returning an empty iterator

Issue #3143, opened 2023-08-18
JSON parse error "parent.full_name" on Bitbucket Cloud


zzvara:
No changes have been made to how Scala Steward is started. (It runs every 15 minutes on top of Kubernetes as a Cron Job.)

```
org.scalasteward.core.util.DecodeFailureWithContext: uri: https://api.bitbucket.org/2.0/repositories/redacted/redacted
method: GET
message: Invalid message body: Could not decode JSON: {
  # redacted
  "parent" : null
}
at org.http4s.circe.CirceInstances$.org$http4s$circe$CirceInstances$$jsonDecodeErrorHelper(CirceInstances.scala:282)
	at org.http4s.circe.CirceInstances$.$anonfun$defaultJsonDecodeError$1(CirceInstances.scala:269)
	at org.http4s.circe.CirceInstances.$anonfun$jsonOfWithMediaHelper$2(CirceInstances.scala:115)
	at scala.util.Either.fold(Either.scala:190)
	at org.http4s.circe.CirceInstances.$anonfun$jsonOfWithMediaHelper$1(CirceInstances.scala:116)
	at cats.data.EitherT.$anonfun$flatMap$1(EitherT.scala:446)
	at cats.effect.IOFiber.succeeded(IOFiber.scala:1198)
	at cats.effect.IOFiber.runLoop(IOFiber.scala:240)
	at cats.effect.IOFiber.asyncContinueSuccessfulR(IOFiber.scala:1366)
	at cats.effect.IOFiber.run(IOFiber.scala:113)
	at cats.effect.unsafe.WorkerThread.run(WorkerThread.scala:632)
Caused by: io.circe.DecodingFailure$DecodingFailureImpl: DecodingFailure at .parent.full_name: Missing required field
```

Arguments used:

```
scalaStewardArgs:
      - --forge-type
      - bitbucket
      - --forge-api-host
      - https://api.bitbucket.org/2.0
      - --forge-login
      - redacted
      - --git-author-email
      - redacted
      - --git-ask-pass
      - /opt/scala-steward/password.sh # Bitbucket PAT is provided here.
      - --workspace
      - /var/scala-steward/workspace
      - --repos-file
      - /opt/scala-steward/repos.md
      - --scalafix-migrations
      - /opt/scala-steward/scalafix.conf
      - --cache-ttl
      - 0s
      - --do-not-fork
```

[...]

PR #3150, opened 2023-08-28
Handle null value in bitbucket API response for repository

jduval87:
As pointed out in issue [#3143](https://github.com/scala-steward-org/scala-steward/issues/3143) it looks like bitbucket cloud API has started to emit null jsvalue which break the circe custom decoder for RepositoryResponse.

The proposed solution is a bit naive. And I hacked a test around with what is already available.

Feel free to edit.


zzvara:
Scala Steward has failed ever since on Bitbucket Cloud. Could someone merge this PR? <3

alejandrohdezma (approved):
Thanks for the contribution @jduval87! And sorry it took so long, I approved the CI and forgot to come back later on 😓

PR #3167, opened 2023-09-14
Keep sbt at 1.9.4 to investigate a potential issue in 1.9.5

lrytz:


mdedetrich:
@fthomas @mzuehlke Can you look into this? Its somewhat urgent

alejandrohdezma:
I believe the version should be set to one being ignored (`1.9.5`)

lrytz:
Ah right, the change is in the `updates.ignore` block. Updated the patch.

alejandrohdezma (approved):

PR #3169, opened 2023-09-15
Hold back artifacts published with sbt 1.9.5

mkurz:
These libraries where published with sbt 1.9.5 and should not be advertisted by scala-steward since they might be broken, see [sbt 1.9.5 release notes](https://github.com/sbt/sbt/releases/tag/v1.9.5):

> **Update**: ⚠️ sbt 1.9.5 is broken, because it causes Scala compiler to generate wrong class names for anonymous class on lambda. While we investigate please refrain from publishing libraries with it.
https://github.com/scala/bug/issues/12868#issuecomment-1720848704

Also see
* #3167
* https://github.com/scala/bug/issues/12868


mkurz:
@fthomas @mzuehlke @alejandrohdezma can you please merge? Thanks!

mzuehlke (approved):

PR #3188, opened 2023-10-18
support all scala-cli dependency directives

### LINKED ISSUE #3174: scala-cli updates not getting picked up
I have a scala-cli repo that I noticed isn't getting PRs sent in. I'm running the steward action on it and you can see [here](https://github.com/ckipp01/chris-kipp.io/actions/runs/6244185904/job/16950746242#step:5:345) that it finds no updates for skan. (ignore the failure in the other repos, it's a mill issue). However there are updates for this repo. For example upickle is outdated and when you run it with scala-cli you can see this.

bishabosha:
I have made a list of dependency aliases from https://github.com/VirtusLab/scala-cli/blob/9e22d4a91ba8699ac2727d2ac3042d64abe951e1/modules/directives/src/main/scala/scala/build/preprocessing/directives/Dependency.scala#L33-L48

as far as I know these are not exposed programatically, but I guess they could be scraped using a source generator

fixes #3174


modules/core/src/main/scala/org/scalasteward/core/buildtool/scalacli/ScalaCliAlg.scala
```diff
@@ -42,8 +63,8 @@ final class ScalaCliAlg[F[_]](implicit
   override def containsBuild(buildRoot: BuildRoot): F[Boolean] = {
     val buildRootPath = buildRoot.relativePath.dropWhile(Set('.', '/'))
     val extensions = Set(".sc", ".scala")
-    gitAlg
-      .findFilesContaining(buildRoot.repo, "//> using lib ")
+    ScalaCliAlg.directives
+      .flatTraverse(gitAlg.findFilesContaining(buildRoot.repo, _))
```
bishabosha, line 67: I guess there is probably a more efficient way to do this, e.g. using one command, but I didn't spend too long looking into this

bishabosha:
I updated the BuildToolDispatcherTest to account for the new searches

bishabosha:
@fthomas is there something I should change about the approach? We'd like to enable scala-steward on scala/toolkit and this would help

modules/core/src/main/scala/org/scalasteward/core/buildtool/scalacli/ScalaCliAlg.scala
```diff
@@ -28,6 +28,27 @@ import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
...
+      "test.dependency",
+      "test.dep",
+      "test.deps",
+      "test.dependencies",
+      "compileOnly.lib",
+      "compileOnly.libs",
+      "compileOnly.dep",
+      "compileOnly.deps",
+      "compileOnly.dependencies"
+    ).map(alias => s"//> $alias ")
```
fthomas, line 49: Isn't `using` missing here?
```suggestion
    ).map(alias => s"//> using $alias ")
```

bishabosha: oh good point!

modules/core/src/main/scala/org/scalasteward/core/buildtool/scalacli/ScalaCliAlg.scala
fthomas: Yeah, `git grep` (which is used by `findFilesContaining`) can search for multiple patterns in one go by combining them via [`--or`](https://git-scm.com/docs/git-grep#Documentation/git-grep.txt---or).

I just measured the wall clock time of `containsBuild` before and with this change for this repo and it went from 7 to 82 milliseconds. IMO this is acceptable since it is still dwarfed by actually running any build tool to extract dependencies.

fthomas (approved):

modules/core/src/main/scala/org/scalasteward/core/buildtool/scalacli/ScalaCliAlg.scala
```diff
@@ -28,6 +28,27 @@ import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
...
+      "test.dependency",
+      "test.dep",
+      "test.deps",
+      "test.dependencies",
+      "compileOnly.lib",
+      "compileOnly.libs",
+      "compileOnly.dep",
+      "compileOnly.deps",
+      "compileOnly.dependencies"
+    ).map(alias => s"//> using $alias ")
```
armanbilge, line 49: @bishabosha do we need `toolkit` in this list as well? So we can get updates for directives like
```
//> using toolkit 0.2.1
```

bishabosha: From what I could see this would need a bigger change as the algorithm searches for explicit strings so unless you want to search for all possible toolkits this isn't very flexible

armanbilge: @bishabosha thanks for looking at it but I'm not sure I understand what you mean. Are you saying that adding `toolkit` here _wouldn't_ match
```
//> using toolkit typelevel:0.1.19
```
why not?

fthomas: Scala Steward can already update toolkits: https://github.com/scala-steward-org/test-repo-1/pull/82/files
So I think it makes sense to add `//> using toolkit` here so that projects, that have no other dependencies than a toolkit, are recognized as Scala CLI projects.

PR #3191, opened 2023-10-19
Play 3 will migrate to new groupId

ihostage:


mkurz (changes requested):
@ihostage Two artifacts are missing:
* play-run-support (https://repo1.maven.org/maven2/org/playframework/play-run-support_2.12/)
* sbt-play-ebean (https://repo1.maven.org/maven2/org/playframework/sbt-play-ebean_2.12_1.0/)


Otherwise LGTM, thanks!

ihostage:
Don't understand how I could miss them 🤦‍♂️ 
I remember clearly these modules in my list to migrate 🙃

mkurz (approved):
:+1:

mzuehlke (approved):

modules/core/src/main/resources/artifact-migrations.v2.conf
```diff
@@ -175,6 +175,316 @@ changes = [
...
   },
+  {
+    groupIdBefore = com.typesafe.play
+    groupIdAfter = org.playframework
+    artifactIdAfter = play-bom
+  },
+  {
+    groupIdBefore = com.typesafe.play
+    groupIdAfter = org.playframework
+    artifactIdAfter = play
```
rtyley, line 186: Can I ask - should `artifactIdAfter` be `play_3.0.0` here, rather than than just `play`, to indicate that the new group id of `org.playframework` [only comes into effect with Play 3](https://www.playframework.com/documentation/3.0.x/Migration30#Changed-groupId)?

Because the current configuration for the `org.playframework` migration doesn't have a version threshold configured, Scala Steward is now raising PRs that fail (eg https://github.com/guardian/play-secret-rotation/pull/413) - because there is no `"org.playframework" %% "play" % "2.8.21"`, only [`"com.typesafe.play" %% "play" % "2.8.21"`](https://search.maven.org/artifact/com.typesafe.play/play_2.13/2.8.21/jar), but Scala Steward is trying to [update them to `org.playframework` anyway](https://github.com/guardian/play-secret-rotation/pull/413/files#diff-5634c415cd8c8504fdb973a3ed092300b43c4b8fc1e184f7249eb29a55511f91R56):

image


I'm asking whether `artifactIdAfter` should be `play_3.0.0`, but I can't see in the [documentation](https://github.com/scala-steward-org/scala-steward/blob/main/docs/artifact-migrations.md#how-does-this-work) that this artifact-version-threshold syntax would actually work... all I do see is, looking elsewhere in this `artifact-migrations.v2.conf` file, a handful of other migrations did have that syntax:

https://github.com/scala-steward-org/scala-steward/blob/a47139440e501e536c1b1ce65f4e654fa179f8cb/modules/core/src/main/resources/artifact-migrations.v2.conf#L338-L342

I guess it could be argued that Scala Steward should be able to work out if a particular artifact actually exists or not, so should refrain from making changes like `"org.playframework" %% "play" % "2.8.21"` - ie this is a bug/problem with Scala Steward itself, and it shouldn't be up to the changes in the `artifact-migrations.v2.conf` file to try and describe what artifact version a group id should switch on.

Just for reference, here are some previous PRs on custom group/artifact migration:

* https://github.com/scala-steward-org/scala-steward/pull/1298
* https://github.com/scala-steward-org/scala-steward/pull/1686
* https://github.com/scala-steward-org/scala-steward/pull/2474

ihostage: Hi, @rtyley! `argonaut-refined_6.2` is a part of artifactId and isn't related to version of this artifact. In fact it was a migration from `argonaut-refined_6.2_2.11:1.2.0` to `argonaut-refined_6.3_2.12:1.3.0`. 
I don't really know how to configure Steward to set a version start from that need to change a groupId 🤷‍♂️ 

@mkurz do you know?

fthomas: The change in this PR looks fine and there should be no need to set a version threshold. Scala Steward should figure out that the update from 2.8.20 to 2.8.21 does not require an artifact change. It very much looks like a bug in Scala Steward.

PR #3198, opened 2023-10-27
OpenTelemetry Semantic Conventions groupId migrated

azthec:
I noticed that we didn't automatically get prompted to version bump opentelemetry-semconv.

This should be the needed additions for the groupId migration

### DIFF
--- a/modules/core/src/main/resources/artifact-migrations.v2.conf
+++ b/modules/core/src/main/resources/artifact-migrations.v2.conf
@@ -994,4 +994,9 @@ changes = [
     groupIdAfter = com.github.sbt
     artifactIdAfter = sbt-osgi
   },
+  {
+    groupIdBefore = io.opentelemetry.semconv
+    groupIdAfter = io.opentelemetry
+    artifactIdAfter = opentelemetry-semconv
+  },
 ]


fthomas:
👍 

Could you provide a link to an issue, pull request or some other documentation in the opentelemetry-semconv project that is about this groupId change?

azthec:
@fthomas I updated the description

fthomas:
I'm confused. If I look at https://repo1.maven.org/maven2/io/opentelemetry/opentelemetry-semconv/ and https://repo1.maven.org/maven2/io/opentelemetry/semconv/opentelemetry-semconv/, I see that the newest version from 2023-10-26 uses `io.opentelemetry.semconv` as groupId and it looks to me that the build also uses this groupId: https://github.com/open-telemetry/semantic-conventions-java/blob/8bbebbcf6d66230178dba60332375fb41d4f4e6c/build.gradle.kts#L37. But the migration in this PR would change the groupId from `io.opentelemetry.semconv` to `io.opentelemetry`. What am I missing?

azthec:
It seems like they handle versioning differently for this, thanks anyway

PR #3211, opened 2023-11-13
Catch exceptions in Substring.Replacement.applyAll

fthomas:


alejandrohdezma (approved):

fthomas:
I just want to note that exceptions thrown by `applyAll` are currently caught by Cats Effect and adapted by https://github.com/scala-steward-org/scala-steward/blob/v0.26.0/modules/core/src/main/scala/org/scalasteward/core/io/FileAlg.scala#L69 as can be observed from the log excerpt in https://github.com/scala-steward-org/scala-steward/issues/3206#issue-1983959630. Not sure where CE catches the exceptions although they are thrown outside the `F`-context, but it is better to be explicit about `applyAll` being able to throw as is done here and hopefully everywhere else in the code base.

PR #3215, opened 2023-11-17
Use `GitAlg.findFilesContaining` in `ScannerAlg`

fthomas:
This implements the idea mentioned in https://github.com/scala-steward-org/scala-steward/pull/2994#issuecomment-1455174931

`ScannerAlg` now uses `GitAlg.findFilesContaining` instead of our own `FileAlg.findFiles` to find files that contain versions and groupIds. This simplifies `findPathsContaining` because we do not need the gitignore dependency anymore to skip files that are ignored by Git since `git grep`, which is used by `GitAlg.findFilesContaining`, does this automatically for us.

The biggest impact of this change in the code base is in the tests. `EditAlgTest` and `RewriteTest` now need to actually execute `git grep` to find files that should be edited. For this, the `execCommands` flag is used (#3005) and Git repositories are created for these tests. One advantage of this change is that we mock less and the `MockState` traces are closer to what Steward is actually doing in production.


.github/workflows/ci.yml
```diff
@@ -67,6 +67,10 @@ jobs:
       - name: Check that workflows are up to date
         run: sbt githubWorkflowCheck
 
+      - uses: coursier/setup-action@v1
+        with:
+          apps: scalafmt
```
fthomas, line 72: This needed since `RewriteTest."applyUpdate with scalafmt update"` now actually executes `scalafmt`.

PR #3219, opened 2023-11-22
Reset instead of merge to update branches

### LINKED ISSUE #3218: Update branches via reset instead of reverting changes
This continues the feature request and discussion about changing the method how branches are updated that started in https://github.com/scala-steward-org/scala-steward/issues/1761#issuecomment-1584014610.

fthomas:
Closes #3218


alejandrohdezma (approved):
This change would make daily life of reviewers a lot easier! 🎉

PR #3237, opened 2023-12-09
Allow multiple and non-local repos files

fthomas:
This allows the `--repos-file` to be repeated so that repositories can be read from multiple files and it changes the type from `File` to `Uri` to allow those files to be non-local. It makes `--repos-file` similar to the `--artifact-migrations` and `--scalafix-migrations` options which both can be repeated and take URIs.


exoego (approved):
Awesome

PR #3241, opened 2023-12-10
Fail on `--vcs-*` options

fthomas:
This makes `--vcs-*` options unusable. If they are used, Scala Steward will print a message and exit with an error.

These options are deprecated since https://github.com/scala-steward-org/scala-steward/pull/2916 and should be replaced with the corresponding `--forge-*` options.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Cli.scala
@@ -107,7 +107,7 @@ object Cli {
       "vcs-type",
       s"deprecated in favor of --${name.forgeType}",
       visibility = Visibility.Partial
-    )
+    ).validate(s"--vcs-type is deprecated; use --${name.forgeType} instead")(_ => false)
 
   private val forgeType = {
     val help = ForgeType.all.map(_.asString).mkString("One of ", ", ", "") +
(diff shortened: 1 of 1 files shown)


fthomas:
Heads-up @tgodzik: Is the public Scala Steward instance already using the `--forge-*` options instead of `--vcs-*`? If not, it will stop working once this PR is merged.

exoego (commented):
To not surprise users in winter holiday, let's merge this in 2024 January or later.

fthomas:
Sounds good to me.

tgodzik:
> Heads-up @tgodzik: Is the public Scala Steward instance already using the `--forge-*` options instead of `--vcs-*`? If not, it will stop working once this PR is merged.

Thanks for the heads up! I updated it and deployed the new config.

fthomas:
My plan is to merge this on 2024-01-15 and maybe also release 0.29.0 then.

rtyley:
Just to note that the minimum necessary version of [`scala-steward-action`](https://github.com/scala-steward-org/scala-steward-action) (if you're using it) after this PR is [**v2.62.0**](https://github.com/scala-steward-org/scala-steward-action/releases/tag/v2.62.0) - otherwise the run will fail, and you'll see error messages like [these](https://github.com/guardian/scala-steward-public-repos/actions/runs/7623944293/job/20765048127):

```
Error: --vcs-api-host is deprecated; use --forge-api-host instead
Error: --vcs-login is deprecated; use --forge-login instead
```

image

fthomas:
> I haven't been seeing the deprecation warnings appear in the actual Scala Steward runs themselves, eg this [recent run](https://github.com/guardian/scala-steward-public-repos/actions/runs/7615995487/job/20741769675) doesn't contain the string 'deprecated in favor of'.

That string was only visible in the help message of the command-line options. Scala Steward did not log a warning if that option was used after #2916. It would have been nice if it had done that.

PR #3244, opened 2023-12-13
Add default implementation for `BuildToolAlg#runMigration`

fthomas:
This adds a default implementation for `BuildToolAlg#runMigration` that logs a warning that Scalafix migrations are not supported by the build tool. It saves some repeated text that is used for every build tool except sbt.


exoego (approved):
It makes a sense to me

PR #3247, opened 2023-12-15
Remove `StewardContext` hierarchy

fthomas:
This removes the `Context.StewardContext` hierarchy. IMO `StewardContext` makes `Context.scala` unnecessarily more complicated only for saving one `case` in `Main.run` and a little duplication.


alejandrohdezma (approved):
LGTM

PR #3248, opened 2023-12-20
Update short description of Scala Steward

fthomas:
The previous description was from the days when Scala Steward only worked with sbt builds. The new description is more generic and lists the supported build tools.


exoego (approved):
LGTM

PR #3250, opened 2023-12-22
Do not cache the output of `gitAskPass`

fthomas:
Prior to this change, the `gitAskPass` program was called once on start-up and its output was cached and used for API calls to the forge while Git itself does not cache its output but calls it anytime a password is needed. This means if the output of `gitAskPass` changes during a Scala Steward run, the new password is only used for Git operations but not for forge API calls.

With this change we now do the same as Git and call `gitAskPass` everytime the password is needed. This should make it easier to support GitHub Apps proper which require different access tokens during a run. See also https://github.com/scala-steward-org/scala-steward/issues/2973#issuecomment-1866224921.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/Config.scala
@@ -17,19 +17,13 @@
 package org.scalasteward.core.application
 
 import better.files.File
-import cats.Monad
-import cats.syntax.all._
 import org.http4s.Uri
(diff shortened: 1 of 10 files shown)


exoego (approved):
Makes sense to me

fthomas:
> This should make it easier to support GitHub Apps proper which require different access tokens during a run.

I tried this successfully in https://github.com/scala-steward-org/scala-steward/compare/topic/gh-app-2 which modifies `gitAskPass` between runs for GitHub App installations.

PR #3264, opened 2024-01-08
Automatically merge PRs by the Scala Steward dev account

fthomas:


fthomas:
Btw, I'm now using @scala-steward-dev instead of @scala-steward for developing/testing Scala Steward and I can hand out access tokens for the dev account to anyone from @scala-steward-org/core for local testing. Just ping me if you want an access token.

PR #3282, opened 2024-02-06
Increase default `--max-buffer-size`

fthomas:
See https://github.com/VirtusLab/scala-steward-repos/issues/98#issuecomment-1929368687

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -57,7 +57,7 @@ Options and flags:
     --disable-sandbox
         Whether to not use the sandbox
     --max-buffer-size <integer>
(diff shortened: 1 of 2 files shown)


clairemcginty:
Hi @fthomas, would you be open to bumping this value again? (I can create a PR.) it looks like low buffer size has been a blocker for a repo in my org, [Scio](https://github.com/spotify/scio), to receive updates, but I'm able to run it locally with a `--max-buffer-size` of `49152`.

mzuehlke:
Please do so 👍

Issue #3283, opened 2024-02-09
Scala Steward runs as root when using the published Docker image (but it probably shouldn't)


bpholt:
The `Dockerfile` generated by the current build configuration leaves the user set to `root`, meaning the application runs as `root` inside the container. This is not recommended, and I think was probably an oversight.

https://github.com/scala-steward-org/scala-steward/blob/1f8ea306972463c40c36ab68bd74078fdc260e7b/build.sbt#L379-L393

Line 380 sets the user to `root` so the build can configure the dependencies inside the image, but as you can see there's no subsequent `USER {user}` directive to reset the user context.

We're in the process of testing this locally, but I think adding this as the final `Dockerfile` directive should reset the user context back to the user created by sbt-native-packager.

```scala
Cmd("USER", (Docker / daemonUser).value)
```

I'll make a PR to fix this if the tests we're doing internally look promising.

PR #3284, opened 2024-02-09
Improve documentation for workspace ownership mismatches

bpholt:
[Git requires](https://github.com/git/git/commit/3b0bf2704980b1ed6018622bdf5377ec22289688) that a repository directory be owned by the user executing `git`. In the normal case, it will print a message

> fatal: detected dubious ownership in repository …

which is easily searchable online, but unfortunately `git config` does not print such a nice message, and simply prints

> fatal: not in a git directory

which is misleading and hard to diagnose.

This came up for us when running Scala Steward as a AWS Fargate task using AWS EFS as its persistent storage. We configured the task to mount the EFS volume using UID 1001, which is the default UID for sbt-native-packager Docker containers, but in fact, Scala Steward current runs as root (#3283). While looking into the issue, we noticed that the article titled [_Running scala-steward periodically on AWS Fargate_](https://medium.com/@tanishiking/running-scala-steward-periodically-on-aws-fargate-3d3d202f0f7) no longer exists, so this PR also removes that link from the documentation. (Hopefully we'll be able to publish some details about how we're running Scala Steward using Fargate as a replacement!)

### DIFF
--- a/docs/running.md
+++ b/docs/running.md
@@ -203,7 +203,6 @@ pipelines:
 There is multiple articles on how to run Scala Steward on-premise:
 
 * [Running Scala Steward On-premise](https://engineering.avast.io/running-scala-steward-on-premise)
-* [Running scala-steward periodically on AWS Fargate](https://medium.com/@tanishiking/running-scala-steward-periodically-on-aws-fargate-3d3d202f0f7)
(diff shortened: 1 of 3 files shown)


bpholt:
Looks like Codecov is complaining about the fallback code not being tested. This isn't the kind of thing I would typically write a test for, so I'll probably hold off on doing so until I get some feedback about whether this change would be welcome or not. If it's welcome and just needs to be tested, I'll follow up!

exoego (approved):
LGTM

PR #3285, opened 2024-02-14
Ignore Scala 3.4.0 until it's announced

rtyley:
We're already seeing [Scala Steward raise PRs for Scala 3.4.0](https://github.com/guardian/play-secret-rotation/pull/432), this might be unintentional, as there are no [GitHub release notes for 3.4.0](https://github.com/lampepfl/dotty/releases/tag/3.4.0)?

See also https://github.com/scala-steward-org/scala-steward/pull/2337 & https://github.com/scala-steward-org/scala-steward/issues/1104.

cc @Kordyjan

### DIFF
--- a/modules/core/src/main/resources/default.scala-steward.conf
+++ b/modules/core/src/main/resources/default.scala-steward.conf
@@ -31,6 +31,9 @@ updates.ignore = [
   // Artifacts below are ignored because they are not yet announced.
 
   // Ignore the next Scala 3 version until it is announced.
+  { groupId = "org.scala-lang", artifactId = "scala3-compiler",     version = { exact = "3.4.0" } },
+  { groupId = "org.scala-lang", artifactId = "scala3-library",      version = { exact = "3.4.0" } },
+  { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1", version = { exact = "3.4.0" } },
   { groupId = "org.scala-lang", artifactId = "scala3-compiler",     version = { exact = "3.3.2" } },
   { groupId = "org.scala-lang", artifactId = "scala3-library",      version = { exact = "3.3.2" } },
   { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1", version = { exact = "3.3.2" } },


modules/core/src/main/resources/default.scala-steward.conf
```diff
@@ -31,6 +31,9 @@ updates.ignore = [
   // Artifacts below are ignored because they are not yet announced.
 
   // Ignore the next Scala 3 version until it is announced.
+  { groupId = "org.scala-lang", artifactId = "scala3-compiler",     version = { exact = "3.4.0" } },
+  { groupId = "org.scala-lang", artifactId = "scala3-library",      version = { exact = "3.4.0" } },
+  { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1", version = { exact = "3.4.0" } },
   { groupId = "org.scala-lang", artifactId = "scala3-compiler",     version = { exact = "3.3.2" } },
   { groupId = "org.scala-lang", artifactId = "scala3-library",      version = { exact = "3.3.2" } },
   { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1", version = { exact = "3.3.2" } },
```
rtyley, line 39: I'm not sure if there's a more concise way to express these lines? eg 
```suggestion
  { groupId = "org.scala-lang", artifactId = "scala3-compiler",     version = [ { exact = "3.4.0" }, { exact = "3.3.2" } ]},
  { groupId = "org.scala-lang", artifactId = "scala3-library",      version = [ { exact = "3.4.0" }, { exact = "3.3.2" } ]},
  { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1", version = [ { exact = "3.4.0" }, { exact = "3.3.2" } ]},
```

Kordyjan (approved):

rtyley:
> Scala Steward should never update from LTS to not-LTS

I think that's a good rule, but I'm not sure if it will be respected by Scala Steward going forward - I don't think Scala Steward currently has any conception of Scala LTS releases...

I wonder: what _is_ the correct configuration here when 3.4.0 is released? We might want to update this configuration to ignore both 3.4.0 & 3.4.1... but then when 3.4.1 is released, you can't configure 3.4.1 to be ignored any longer (users of 3.4.0 _should_ upgrade) - and at that point, I'm uncertain if Scala Steward _would_ hold back on recommending an update from Scala 3.3 to 3.4.1.


Another relevant issue is https://github.com/scala-steward-org/scala-steward/issues/3153, with commentary from @mzuehlke.

mzuehlke (approved):

mzuehlke:
> I wonder: what _is_ the correct configuration here when 3.4.0 is released? We might want to update this configuration to ignore both 3.4.0 & 3.4.1... but then when 3.4.1 is released, you can't configure 3.4.1 to be ignored any longer (users of 3.4.0 _should_ upgrade) - and at that point, I'm uncertain if Scala Steward _would_ hold back on recommending an update from Scala 3.3 to 3.4.1.

I guess without additional logic Scala Steward doesn't know about the concept of an LTS. maybe @fthomas has a good ies

Issue #3302, opened 2024-02-28
Avoid pushing scala updates for non LTS versions


etspaceman:
It would be very helpful if updates for the scala language were limited to the LTS versions unless otherwise configured. The Steward pushed out versions for 3.4 recently and this version is not intended for libraries.

bpholt:
Related https://github.com/scala-steward-org/scala-steward/issues/2467

SethTisue:
(now that I'm back on my computer I see that) It's possible this was already fixed, as at modules/core/src/main/resources/default.scala-steward.conf I see

```
  // Ignore the next Scala 3 version until it is announced.
  { groupId = "org.scala-lang", artifactId = "scala3-compiler",     version = { exact = "3.4.0" } },
  { groupId = "org.scala-lang", artifactId = "scala3-library",      version = { exact = "3.4.0" } },
  { groupId = "org.scala-lang", artifactId = "scala3-library_sjs1", version = { exact = "3.4.0" } },
```

@Kordyjan but of course this is specific to 3.4.0; is there some update to the Scala 3 release steps that is still needed to prevent this from reoccurring...?

Kordyjan:
For now, 3.4 releases are ignored. It will remain that way for some time. Later, we will propose a change to the Scala steward to only bump the version to the Scala Next version if the project is already on Scala Next.

valencik:
@Kordyjan 
> For now, 3.4 releases are ignored.

It looks like updates are still going out: https://github.com/typelevel/typelevel.g8/pull/135

mpilquist:
Another example: https://github.com/Comcast/ip4s/pull/575

mzuehlke:
#3323 should fix that

Issue #2467, opened 2022-01-18
Should scala-steward stop recommending to upgrade Scala 3 minor versions?


jodersky:
Scala 3.x is backward compatible, but not forward compatible. For example, this means that a library built with 3.0.2 can be consumed by a library built with 3.1.0, but NOT the other way around (See https://twitter.com/SethTisue/status/1450260785762947076 for some more info). Considering these compatibility semantics, should scala-steward by default stop recommending to upgrade scala versions if they are >= 3?

I think that there will be little disagreement that for library developers, the best practice is to be conservative and indeed use the lowest Scala 3 version as reasonably possible (3.0.2 as of this writing). End-user application developers may have more liberty, but even then I am not sure that the benefits of always using the latest Scala version are that important. Hence, unless there is a way for projects to indicate to scala-steward if they are a library or application, I would recommend that the default behavior should be changed to not upgrade Scala 3 versions beyond the newest patch-level.

fthomas:
My understanding from https://github.com/scala/scala-lang/blob/5303453ebe4d1597adce544871f6712e217cff19/blog/_posts/2022-01-17-scala-3.1.1-released.md#312-rc1-and-forward-compatibility-improvements is that library authors can use the latest Scala 3 version if they set `-Yscala-release=3.0`. I think it is fine then if Scala Steward keeps proposing the new Scala 3 versions.

jodersky:
That is an interesting blog post, I was not aware of this new flag. IIRC the `-Y` flags are private, so I'm not sure how stable this flag will be. If it does end up being a stable feature, then I agree that it would nullify this issue.

fthomas:
Here is a new blog post that is relevant for this issue: [Long-term compatibility plans for Scala 3](https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html)

mzuehlke:
In light of the last blog post mentioned by @fthomas I see no reason to build some special handling for Scala version upgrades into Scala Steward. If a project wants to restrict upgrades of Scala 3 it can `pin` a specific minor version:

```HOCON
updates.pin = [
  { groupId = "org.scala-lang", artifactId="scala3-compiler", version = "3.1." },
  { groupId = "org.scala-lang", artifactId="scala3-library", version = "3.1." },
  { groupId = "org.scala-lang", artifactId="scala3-library_sjs1", version = "3.1." }
]
```
[...]

PR #3305, opened 2024-02-29
Fix the merging order updates.pin

mzuehlke:
The merging of the default Scala Steward configuration for pinned artifacts was reversed.

In contrast to the comment and intend:
```Scala
//  Strategy: union with repo preference in terms of revision
```

This pull-request fixes the merging order and adapts the test. In addition it adds tests that show how `updates.pin` can  be used to control Scala LTS release updates.
Discussed in 
- #3302 
- #2467


exoego (approved):
nice catch

armanbilge:
Not entirely sure, but I think this change had an unexpected result here.
- https://github.com/typelevel/sbt-typelevel/pull/703

PR #3323, opened 2024-03-28
Ignore any non LTS version

mzuehlke:
Users should not be updated to non-LTS versions automatically

see #3302


SethTisue:
just to check: if somebody is already on 3.4.0, will it still suggest 3.4.1, 3.4.2, etc?

mzuehlke:
> just to check: if somebody is already on 3.4.0, will it still suggest 3.4.1, 3.4.2, etc?

At the moment no  3.4.x version will be suggested. 
#3328 should handle this better

PR #3328, opened 2024-03-31
Handle Scala LTS and Next version update

mzuehlke:
Scala Steward now distinguishes between Scala Next and Scala LTS.

When the current Scala 3 versions is below 3.4.0 no updates to a newer Scala Next version will be created. But updates to newer Scala 3.3.x version will be created.

If the current version is already a Scala Next version (3.4.x), updates to newer Scala Next versions will be created.

In the future the `versionLine` property can be used to identify LTS and Next version. (https://github.com/scala/scala3/pull/19986)

fixes: #3302
followup to: #3323

I am bit unsure if I should release the changes  to `default.scala-steward.conf` later, after cutting a new release and a bit of wait time, as otherwise older versions would again create pull-requests for Scala 3.4.1. But that probably has already happened in most cases.

### DIFF
--- a/modules/core/src/main/resources/default.scala-steward.conf
+++ b/modules/core/src/main/resources/default.scala-steward.conf
@@ -30,10 +30,10 @@ postUpdateHooks = [
 updates.ignore = [
   // Artifacts below are ignored because they are not yet announced.
 
-  // No upgrades to non LTS versions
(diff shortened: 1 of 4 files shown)


mkurz:
Looks like this is not working as intended (anymore): https://github.com/playframework/play-json/pull/1035#issuecomment-2099773207, maybe #3347 broke things?

mzuehlke:
@mkurz
 I found a bug and created a PR that will prevent such PRs in the future: #3354

Issue #3348, opened 2024-05-07
Wrong targetRefName when create azure-repo PullRequestPayload


GreyPlane:
configured repo was like $owner/$repo:develop, other functionality was running fine, but when Steward trying to create a PR, it gets
```json
{
        "$id": "1",
        "errorCode": 0,
        "eventId": 3000,
        "innerException": null,
        "message": "TF401398: The pull request cannot be activated because the source and/or the target branch no longer exists, or the requested refs are not branches",
        "typeKey": "GitPullRequestCannotBeActivated",
        "typeName": "Microsoft.TeamFoundation.Git.Server.GitPullRequestCannotBeActivated, Microsoft.TeamFoundation.Git.Server"
}
```

after investigation, it seems the `targetRefName` was `develop` instead of `refs/head/develop`, also it appears in both tests and [Azure repo API docs](https://learn.microsoft.com/en-us/rest/api/azure/devops/git/pull-requests/create?view=azure-devops-rest-4.1&tabs=HTTP#examples), it should be `refs/head/develop`.

Is this due to something changed in `NewPullRequestData.base.name`, or AzureRepos changed their API?

PR #3349, opened 2024-05-07
fix azure repos targetRefName

GreyPlane:
#3348


exoego (approved):
Thanks.
As you pointed out, `"targetRefName": "refs/heads/new_feature"` appears to be a valid format.

https://learn.microsoft.com/en-us/rest/api/azure/devops/git/pull-requests/create?view=azure-devops-rest-4.1&tabs=HTTP#examples

PR #3354, opened 2024-05-08
Fix Scala 3 check

mzuehlke:
* the artifact contains a cross version, so only comparing the artifact name leads to the correct result 
* adapt lowest Scala 3 NEXT version

Tested with  https://github.com/playframework/play-json to prevent in the future pull requests like https://github.com/playframework/play-json/pull/1035


mkurz:
Thanks!

PR #3356, opened 2024-05-10
use foldMap instead of collectFold

GreyPlane:
since `fileAlg.createTemporarily` is a total function, it's a little better to use `foldMap` rather than `collectFold`


exoego (approved):
LGTM

PR #3360, opened 2024-05-22
Add migration for munit-cats-effect

armanbilge:


mzuehlke (approved):

armanbilge:
For posterity, the artifact name change originally happened in https://github.com/typelevel/munit-cats-effect/pull/223.

PR #3361, opened 2024-05-23
Fix ambiguous argument 'main'

olivierdeckers:
When scala-steward attempts to update a PR and a file/folder in the repository has the same name as the branch that it is being run against, it fails with a fatal error: 
`fatal: ambiguous argument 'main': both revision and filename`
This change should fix the ambiguity


exoego (approved):
Thanks

PR #3366, opened 2024-05-24
Add migrations for `sttp-oauth2`

mzuehlke:
See: https://github.com/polyvariant/sttp-oauth2/issues/477

@majk-p

### DIFF
--- a/modules/core/src/main/resources/artifact-migrations.v2.conf
+++ b/modules/core/src/main/resources/artifact-migrations.v2.conf
@@ -1344,4 +1344,9 @@ changes = [
     artifactIdBefore = munit-cats-effect-3
     artifactIdAfter = munit-cats-effect
   },
+  {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/resources/artifact-migrations.v2.conf
```diff
@@ -1344,4 +1344,9 @@ changes = [
     artifactIdBefore = munit-cats-effect-3
     artifactIdAfter = munit-cats-effect
   },
+  {
+    groupIdBefore = org.polyvariant
+    groupIdAfter = com.ocadotechnology
```
exoego, line 1349: reversed?


image

mzuehlke: Good catch.

majk-p (approved):

PR #3369, opened 2024-05-31
Only return a failure exit code when all repos have failed

ioannakok:
When even one repository fails in a scala steward run the workspace is not persisted by GitHub Actions because GH Actions won't persist the workspace of a failing job. Persisting workspace is crucial for respecting `pullRequests.frequency` configuration because the workspace is how scala steward records that it has opened a pull request. This means that one failing repository can cause a lot of user annoyance because their configuration is no longer respected and they get too many PRs opened. This change aims to fix that by only returning a failure code if *all* repos have failed. So long as at least one repository succeeded the exit code will be success and the workspace will persist. If administrators need to know what repos are failing they can use the jobs summary introduced by: https://github.com/scala-steward-org/scala-steward/pull/3071

See more here: https://github.com/guardian/scala-steward-public-repos/issues/60

cc @alejandrohdezma

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -4,7 +4,7 @@ All command line arguments for the `scala-steward` application.
 
 ```
 Usage:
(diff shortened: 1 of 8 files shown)


rtyley:
@mzuehlke how does this look to you? To us at the Guardian this does seems like the best fix to the problem?

One possible alternative would be to modify the behaviour of `scala-steward-action` so that `workspace.saveWorkspaceCache()` [gets invoked](https://github.com/scala-steward-org/scala-steward-action/blob/a24b9bbdcee47f6f4e03e6b995878915ef34e040/src/action/main.ts#L86-L89) even _if_ `coursier.launch()` [throws an exception](https://github.com/scala-steward-org/scala-steward-action/blob/a24b9bbdcee47f6f4e03e6b995878915ef34e040/src/modules/coursier.ts#L96-L98), though that probably runs the risk of persisting some inconsistent state.

alejandrohdezma:
Hey @ioannakok, thanks for this! I think this could make sense for a lot of use cases, but not for others. Maybe we can expand on it and put it behind a flag that's been sent in the params. That way we can propagate that flag from the action and people can decide whether to fail if everything is failing or not. What do you think?

alejandrohdezma:
Great work! Thanks! ❤️

mzuehlke (commented):
Sorry for being absent, looks good 👍

PR #3372, opened 2024-06-03
`scala-cli`: `using dep` and `using test.dep` directives support

horothesun:
Fixes https://github.com/scala-steward-org/scala-steward/issues/3371.

I tested these changes on the same `scala-cli`-based repository I was having issues with and the correct PRs got created.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/update/Selector.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/update/Selector.scala
@@ -92,7 +92,7 @@ object Selector {
     Pattern.compile("""//>\s+using\s+dep\s+""")
 
   private def scalaCliUsingTestDep: Pattern =
-    Pattern.compile("""//>\s+using\s+test.dep\s+""")
+    Pattern.compile("""//>\s+using\s+test\.dep\s+""")
 
   private def scalaValInDependencyDefPositions(
       versionPositions: List[VersionPosition],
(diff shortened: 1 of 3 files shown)


modules/core/src/main/scala/org/scalasteward/core/edit/update/Selector.scala
```diff
@@ -84,6 +88,12 @@ object Selector {
   private def scalaCliUsingLib: Pattern =
     Pattern.compile("""//>\s+using\s+lib\s+""")
 
+  private def scalaCliUsingDep: Pattern =
+    Pattern.compile("""//>\s+using\s+dep\s+""")
+
+  private def scalaCliUsingTestDep: Pattern =
+    Pattern.compile("""//>\s+using\s+test.dep\s+""")
```
horothesun, line 95: Not sure about the `.` in `test.dep`. From the [docs](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html), it looks like the it represents "Any character".

We might have to use escape it.

exoego (commented):
Could you add test?

horothesun:
@exoego thanks for taking a look!
I've added some tests 👍

modules/core/src/main/scala/org/scalasteward/core/edit/update/Selector.scala
mzuehlke: does it make sense to have these 3 pattern as a `val` instead of re creating them over and over again ?

horothesun: Absolutely 👍 Thanks for spotting that!

exoego (approved):
LGTM
Thanks!

PR #3377, opened 2024-06-25
Fix exit code policy

esamson:
Aligning expectation that "all repos succeed" means there are no repos with failures.

This brings back the old behavior from RunResults:

    if (reposWithFailures.isEmpty) ExitCode.Success else ExitCode.Error

See:
https://github.com/scala-steward-org/scala-steward/compare/6259adb0...9d9460af

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/application/ExitCodePolicy.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/application/ExitCodePolicy.scala
@@ -28,5 +28,5 @@ object ExitCodePolicy {
 
   val SuccessIfAnyRepoSucceeds: ExitCodePolicy = successIf(_.successRepos.nonEmpty)
 
-  val SuccessOnlyIfAllReposSucceed: ExitCodePolicy = successIf(_.reposWithFailures.nonEmpty)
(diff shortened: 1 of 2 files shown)


alejandrohdezma (approved):

alejandrohdezma:
Good catch @esamson! Thanks!

modules/core/src/main/scala/org/scalasteward/core/application/ExitCodePolicy.scala
```diff
@@ -28,5 +28,5 @@ object ExitCodePolicy {
 
   val SuccessIfAnyRepoSucceeds: ExitCodePolicy = successIf(_.successRepos.nonEmpty)
 
-  val SuccessOnlyIfAllReposSucceed: ExitCodePolicy = successIf(_.reposWithFailures.nonEmpty)
+  val SuccessOnlyIfAllReposSucceed: ExitCodePolicy = successIf(_.reposWithFailures.isEmpty)
```
rtyley, line 31: Good catch! Sorry I didn't test that with https://github.com/scala-steward-org/scala-steward/pull/3369 - thanks for fixing it!

PR #3383, opened 2024-07-27
Simplify KeyValueStore.modifyF

Masynchin:
I came across this line, proved that `fa.flatMap(a => f(a).flatTap(g)) <-> fa.flatMap(f).flatTap(g)`, and applied it to this line. Hope this small change reduces complexity a bit.


alejandrohdezma (approved):
❤️ Thanks!

PR #3410, opened 2024-09-17
Support more Mill buildfiles (build.mill, build.mill.scala)

lefou:
Mill `0.12` will support more build file names.

* `build.sc`
* `build.mill`
* `build.mill.scala`

This change makes sure Scala Steward will detect these.

Fix https://github.com/scala-steward-org/scala-steward/issues/3409

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala
@@ -38,7 +38,12 @@ final class MillAlg[F[_]](defaultResolver: Resolver)(implicit
   override def containsBuild(buildRoot: BuildRoot): F[Boolean] =
     workspaceAlg
       .buildRootDir(buildRoot)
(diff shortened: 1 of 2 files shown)


mzuehlke (approved):
Thanks for the contribution

lefou:
Can we please have a new release including this change?

fthomas:
One thing is missing here for Scala Steward to actually update versions in `build.mill` files. It now can extract versions from that file but [`UpdatesConfig.defaultFileExtensions`](https://github.com/scala-steward-org/scala-steward/blob/02af3e9e77c1d9e67acffc8a1d6f65c5654b2142/modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala#L93) needs to be updated so that Scala Steward also edit `build.mill` files.

fthomas:
Another thing I just noticed is that this line
https://github.com/lefou/scala-steward/blob/58a6a29e94e1626051c1ed4916965df5727eccef/modules/core/src/main/scala/org/scalasteward/core/buildtool/mill/MillAlg.scala#L108 wasn't updated which means that we aren't extracting plugins from `build.mill` files. I guess this should have been updated, too?

PR #3416, opened 2024-09-23
Add `.sdkmanrc` to the default file extensions of files to update

### LINKED ISSUE #3414: Update Scala version in `.sdkmanrc`
It would be nice if Scala Steward could update the Scala version in the `.sdkmanrc` as the SDKMAN became one of the official ways to install Scala.

achinaou:
closes #3414


exoego (approved):
File extensions are used here:
https://github.com/scala-steward-org/scala-steward/blob/b7dbc0b7d8db310947df9c53c3f17d30f8d686d4/modules/core/src/main/scala/org/scalasteward/core/edit/update/ScannerAlg.scala#L69

Looks good to me

PR #3417, opened 2024-09-24
Add multi-arch image support by moving to temurin 21

barryoneill:
Addresses https://github.com/scala-steward-org/scala-steward/issues/2590

See: https://adoptium.net/supported-platforms/ for architecture matrix

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         scala: [2.13]
-        java: [temurin@17, temurin@11]
+        java: [temurin@21, temurin@17, temurin@11]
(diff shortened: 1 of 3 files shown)


mzuehlke (approved):
This looks good, but I am also not using the docker image. 
So I am waiting for feedback from the other maintainers before merging.

barryoneill:
Hi @mzuehlke  - thanks for the help.  I think though there's going to be more to this.   While the contents of the image are now multi-arch friendly, I think the current CI setup is still building a single-architecture image.  I'm not familiar with the sbt-ecr plugin (I abandoned it years ago in favour of native Docker configuration), but I'll try and figure out what's needed.

barryoneill:
oops, meant to comment on the issue.  Moving there.

PR #3419, opened 2024-09-25
Add dockerBuildxPlatforms

barryoneill:
Continuing on https://github.com/scala-steward-org/scala-steward/issues/2590, this allegedly will trigger buildx during a CI build

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -394,6 +394,7 @@ lazy val dockerSettings = Def.settings(
     )
   },
   Docker / packageName := s"fthomas/${name.value}",
+  Docker / dockerBuildxPlatforms := List("linux/amd64", "linux/arm64"),
   dockerUpdateLatest := true,
   dockerAliases ++= {
     if (!isSnapshot.value) Seq(dockerAlias.value.withTag(Option("latest-release"))) else Nil


mzuehlke:
Did you tried if this locally would build a multi arch image ?
CI isn't a help here ...

barryoneill:
Sorry, I phrased that badly.    Adding `dockerBuildxPlatforms` _should_ trigger `Docker / publish` to use buildx to generate the multi-arch manifests.  

Unfortunately, I can't use `Docker / publish` locally, since I use podman as my container engine and it doesn't support the [necessary push flags](https://github.com/containers/buildah/pull/4677).   `Docker / publishLocal` will use the regular build mechanisms, which is what I use locally, hence my comment about CI.
 
If I have time this week I'll try and find/setup a non-podman environment to test this.

barryoneill:
Alright, I think I have made progress.  For anyone else fighting with this on mac m1 - don't use podman.  I switched to orbstack, and added the following to the 'Advanced Engine Config' in the orbstack settings:

```
{
  "features" : {
    "containerd-snapshotter" : true
  }
}
```

And I was able to use `sbt Docker / publish` up till it tried to publish to Docker hub (which of course I don't have permission to :))

I changed the supported architectures to what I could see in `eclipse-temurin:21-alpine`, i.e.

```
❯ docker manifest inspect eclipse-temurin:21-alpine | jq '.manifests[].platform'
{
  "architecture": "amd64",
  "os": "linux"
}
{
  "architecture": "arm64",
  "os": "linux",
  "variant": "v8"
}
```

@mzuehlke Should be good to go now.

mzuehlke (approved):
Thanks for investigating, makes sense 👍

mzuehlke:
I reverted this PR,  because publishing failed: https://github.com/scala-steward-org/scala-steward/actions/runs/11083098359/job/30797020608

It seems the whole publish docker procedure would need to be adapted su support using buildx like described here:
https://docs.docker.com/build/building/multi-platform/

barryoneill:
To me it seems like this project would be better served by building the docker image manually, giving more direct control over the build process, rather than being limited by whatever `sbt-docker` is doing with buildx under the hood.   

Might also be easier to just wait till later in the year and build on arm64 when github [makes it available open source projects later in the year](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/), but that would mean duplicate builds, rather than one buildx invocation.

@mzuehlke I really appreciate the help with these PRs, sorry it didn't work out.  Can you please reopen https://github.com/scala-steward-org/scala-steward/issues/2590 in case someone else wants to get involved!

mzuehlke:
I reopened the ticket

PR #3427, opened 2024-10-08
Add option for --signoff

### LINKED ISSUE #2408: Feature request: Sign-off the Git commit
[Background](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)

This will help repositories which have a DCO check, for example in [this repo](https://github.com/librecaptcha/lc-core/pull/106)

Of course, Scala Steward isn't a legal entity and the commits are very small anyway, so it doesn't really matter from a legal POV. The sign-off will only make the process of merging the PR smoother.

michaelmior:
This should fix #2408

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -4,7 +4,7 @@ All command line arguments for the `scala-steward` application.
 
 ```
 Usage:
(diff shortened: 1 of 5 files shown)


mzuehlke (approved):
Looks good, thanks

michaelmior:
@mzuehlke Thanks! Would there be any chance of having this enabled for the public instance? If preferred, I can add a repo config option as an alternative.

mzuehlke:
Enabling this by default for all repos is something I wouldn't prefer.
But having this as an option in the per repo config file (https://github.com/scala-steward-org/scala-steward/blob/main/docs/repo-specific-configuration.md), to turn this on, sounds like a good idea to me 👍

PR #3432, opened 2024-10-09
Add repo-specific signoff option

michaelmior:


modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
```diff
@@ -88,7 +89,11 @@ object RepoConfig {
               buildRoots = x.buildRoots |+| y.buildRoots,
               assignees = x.assignees |+| y.assignees,
               reviewers = x.reviewers |+| y.reviewers,
-              dependencyOverrides = x.dependencyOverrides |+| y.dependencyOverrides
+              dependencyOverrides = x.dependencyOverrides |+| y.dependencyOverrides,
+              signoffCommits =
+                if (x.signoffCommits.isDefined || y.signoffCommits.isDefined)
+                  Some(x.signoffCommits.getOrElse(false) || y.signoffCommits.getOrElse(false))
+                else None
```
mzuehlke, line 96:
```suggestion
              signoffCommits = x.signoffCommits.orElse(y.signoffCommits)
```
I thinks this should be identical and simpler

mzuehlke (commented):
Thank. That looks really good.
I as you didn't saw a simpler way than passing this like you did.
I have just a tiny suggestion for improving the config code

michaelmior:
@mzuehlke Good call on that change! I amended my previous commit to include that.

mzuehlke (approved):
Thanks

michaelmior:
@mzuehlke Thanks! Just wondering how often the public instance is updated. That is, when should I expect to be able to use this with one of my repos currently updated by @scala-steward?

mzuehlke:
In the past the public steward was always running the latest version. So it should already be live.
Please give it a try. sadly I don't have access to the logs as they are now operated by https://github.com/VirtusLab/scala-steward-repos/

michaelmior:
@mzuehlke Great! I'll give it a shot :)

mzuehlke:
@tgodzik am I right?

tgodzik:
> @tgodzik am I right?

IT fetches the latest image, so that should work depending on the steward schedule which I think runs twice a week.

PR #3433, opened 2024-10-10
add migration for bucket4j

martijnhoekstra:
bucket4j went for a versioning and deployment scheme split in 8+, 11+ and 17+. I moved to 17+ as that should be the most common, and the scalasteward user should now at least be notified of the option to update, allowing to override to a different one if wanted.

### DIFF
--- a/modules/core/src/main/resources/artifact-migrations.v2.conf
+++ b/modules/core/src/main/resources/artifact-migrations.v2.conf
@@ -1405,61 +1405,51 @@ changes = [
     artifactIdAfter = odin-zio
   },
   {
-    groupIdBefore = com.bucket4j
     groupIdAfter = com.bucket4j
     artifactIdBefore = bucket4j-core
     artifactIdAfter = bucket4j_jdk17-core
   },
   {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/resources/artifact-migrations.v2.conf
```diff
@@ -1404,4 +1404,64 @@ changes = [
     groupIdAfter = dev.scalafreaks
     artifactIdAfter = odin-zio
   },
+  {
+    groupIdBefore = com.bucket4j
+    groupIdAfter = com.bucket4j
```
mzuehlke, line 1409: If the `groupId` isn't changing is is enough to just specify `groupIdAfter`.
You will find such examples in the file.

mzuehlke (changes requested):

modules/core/src/main/resources/artifact-migrations.v2.conf
martijnhoekstra: nice, I removed the `groupIdBefore`'s

mzuehlke (approved):

PR #3450, opened 2024-10-22
add more scala-3 artifacts

xuwei-k:


mzuehlke (approved):
That's a lot more we have been missing in the past. Thanks

Issue #3463, opened 2024-11-04
Errors while running on a Mill project


lefou:
this looks like an issues with some older Mill bootstrap scripts (which fail to correctly parse the `--import` arg). I'm not sure if Scala Steward is still bundling https://github.com/lefou/millw, but updating to the latest version could already fix the issue.

The project is: https://github.com/com-lihaoyi/mill
The Mill version used is: `0.12.1`

The full output (https://github.com/com-lihaoyi/mill/actions/runs/11672731224/job/32501971856):

```
  Error: SLF4J(I): Connected with provider of type [ch.qos.logback.classic.spi.LogbackServiceProvider]
  2024-11-04 21:10:54,778 INFO   
    ____            _         ____  _                             _
   / ___|  ___ __ _| | __ _  / ___|| |_ _____      ____ _ _ __ __| |
   \___ \ / __/ _` | |/ _` | \___ \| __/ _ \ \ /\ / / _` | '__/ _` |
    ___) | (_| (_| | | (_| |  ___) | ||  __/\ V  V / (_| | | | (_| |
   |____/ \___\__,_|_|\__,_| |____/ \__\___| \_/\_/ \__,_|_|  \__,_|
   v0.31.0
   
  2024-11-04 21:10:55,539 INFO  Loaded 295 artifact migration(s)
  2024-11-04 21:10:55,702 INFO  Loaded 35 Scalafix migration(s)
  2024-11-04 21:10:55,842 INFO  Loaded 1 repo config(s)
  2024-11-04 21:10:56,098 INFO  Run self checks
  2024-11-04 21:10:56,253 INFO  Using git version 2.47.0
  2024-11-04 21:10:56,782 INFO  Using scalafix 0.13.0
  2024-11-04 21:10:57,337 INFO  Using scalafmt 3.8.4-RC1
  2024-11-04 21:10:58,017 INFO  Removing any run-specific files
  2024-11-04 21:10:58,029 INFO  ──────────── Steward com-lihaoyi/mill ────────────
  2024-11-04 21:10:58,031 INFO  Check cache of com-lihaoyi/mill
  2024-11-04 21:10:58,536 INFO  Clone com-lihaoyi/mill
  2024-11-04 21:12:00,531 INFO  Refresh cache of com-lihaoyi/mill
  2024-11-04 21:12:00,995 INFO  Get dependencies in . from Mill
  2024-11-04 21:12:03,176 ERROR Steward com-lihaoyi/mill failed
[...]

PR #3464, opened 2024-11-04
Replace com-lihaoyi/mill/mill script with lefou/millw/millw

lefou:
This replaces the `mill` script from the Mill project with the script from the dedicated [`millw`](https://github.com/lefou/millw) project.

There is no version of the former script, that properly support all Mill version that Scala Steward can handle.

The `lefou/millw` project supports all Mill version.

Hopefully fix https://github.com/scala-steward-org/scala-steward/issues/3463

I didn't test it locally due to not having set up a test environment for Scala steward. Let me know if that is a deal breaker for you.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -356,14 +356,10 @@ lazy val dockerSettings = Def.settings(
       s"tar -xf $sbtTgz",
       s"rm -f $sbtTgz"
     ).mkString(" && ")
-    val millVer = Dependencies.millScriptVersion
+    val millVer = Dependencies.millwScriptVersion
     val millBin = s"$binDir/mill"
-    val releasePageVersion = millVer.split("-") match {
-      case Array(v, m, _*) if m.startsWith("M") => s"${v}-${m}"
-      case Array(v, _*)                         => v
-    }
     val installMill = Seq(
-      s"$curl $millBin https://github.com/lihaoyi/mill/releases/download/${releasePageVersion}/$millVer",
+      s"$curl $millBin https://raw.githubusercontent.com/lefou/millw/refs/tags/${millVer}/millw",
       s"chmod +x $millBin"
     ).mkString(" && ")
     val csBin = s"$binDir/cs"
--- a/build.sbt
+++ b/build.sbt
@@ -359,7 +359,7 @@ lazy val dockerSettings = Def.settings(
     val millVer = Dependencies.millwScriptVersion
     val millBin = s"$binDir/mill"
     val installMill = Seq(
-      s"$curl $millBin https://raw.githubusercontent.com/lefou/millw/refs/tags/${millVer}/millw",
+      s"$curl $millBin https://github.com/lefou/millw/releases/download/${millVer}/millw-${millVer}",
       s"chmod +x $millBin"
     ).mkString(" && ")
     val csBin = s"$binDir/cs"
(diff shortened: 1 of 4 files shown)


mzuehlke:
Thanks for taking care of the `mill` support in Scala Steward. I have never used `mill` myself, so I am of no big help here.

The `mill` executable that you are changing is the one bundled into the Docker image. The public instance https://github.com/VirtusLab/scala-steward-repos/ uses this. 
But the https://github.com/scala-steward-org/scala-steward-action/ (maintained by @scala-steward-org/gh-action) use it's own way to get a `mill` executable. 
And the invocation is in both cases the same.

Would the command after the change still work with `mill` or only with `millw` ?

lefou:
Since I can't see the logs of the official instance, I don't know what the issues are there, but I can tell you, that for some of my repos there are no updates for quite some months while other repos work very well. Most repos seeing no updates are in the `com-lihaoyi` org, which could just means, the offical Scala Steward instance doesn't support Mill 0.12 yet.

To see some logs for the very Mill project, I enabled the `scala-steward-action` and identified the issue being a buggy script. If this PR isn't the right location to fix it, I'm grateful for any concrete pointers. I can try to look into https://github.com/scala-steward-org/scala-steward-action/ myself, but it's not really my field of expertise.

> Would the command after the change still work with `mill` or only with `millw` ?

TBH, I'm not sure how to read this question. 

First, `millw` can be renamed to `mill`, which is in fact the typical usage. (I only keep the `millw` name, to be able to distinguish in the communication.)

`millw` is a drop-in replacement for `mill`, accepting all parameters as `mill` does and any kind of version selection (`.mill-version` file, `.config/mill-version` file, `MILL_VERSION` env var, `DEFAULT_MILL_VERSION` env var). It especially supports older Mill version down to `0.3.x`. In fact, most Mill users just install `millw` into `$HOME/bin/mill` and never use the local `./mill` script you can find in many projects.

In contrast, the recent `mill` script of the Mill repo only supports the latest Mill versions, and older versions of that script don't fully support newer Mill releases.

Although I think the peculiar issue #3463 is only present in a small set of releases of the  `mill` script and definitely fixed in more recent versions, it's is still advisable to use `millw` instead, due to the better backward compatibility story. And since Scala Steward is all about updating older projects, it's important to be able to read those older projects. The Mill plugin Scala Steward uses, is still able to read Mill projects from version `0.6.0` upwards. So we should ensure the actual runner also works.

If you meant the other change I did (use of the `--no-server` instead of `-i`), it is just some house keeping, as the `--no-server` option is what we really want to use. I can split out that change, if you like.

mzuehlke (approved):

fthomas:
The `lefou/millw` repository has been archived on Dec 8, 2024. It seems to me that we should download the Mill script now from https://github.com/com-lihaoyi/mill again.

lefou:
Yes, all scripts now reside in the mill repository. 

* Shell Script: [`mill`](https://github.com/com-lihaoyi/mill/blob/main/mill)
* Windows Batch Script: [`mill.bat`](https://github.com/com-lihaoyi/mill/blob/main/mill.bat)
* Windows Power Shell Script: [`mill.ps1`](https://github.com/com-lihaoyi/mill/blob/main/ci/mill.ps1)

PR #3477, opened 2024-11-18
Remove trailing comments in parsed SBT module version

aartigao:
At $work we recently noticed that Scala Steward was not updating dependencies that have trailing comments like

```scala
"software.amazon.awssdk"                 % "dynamodb"                     % V.aws, // Overrides `scanamo` version
"software.amazon.awssdk"                 % "sso"                          % V.aws, // Needed for local dev!
"software.amazon.awssdk"                 % "sts"                          % V.aws  // Needed for EKS IAM!
```

these we're workarounding them by placing the comments elsewhere:

```scala
// The following comments are not inlined to avoid breaking Scala Steward!
// - Module `dynamodb` overrides `scanamo` version
// - Module `sso` is needed for local development
// - Module `sts` is needed for EKS IAM
"software.amazon.awssdk" % "dynamodb" % V.aws,
"software.amazon.awssdk" % "sso"      % V.aws,
"software.amazon.awssdk" % "sts"      % V.aws
```

but I think this can be easily avoided by changing the regex like in this PR.

I don't have any idea if this impacts anything else in the code (tests are green at least locally), so let's discuss it because I think the current behavior is not the expected one 🙏🏽

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/update/ModulePositionScanner.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/update/ModulePositionScanner.scala
@@ -45,7 +45,7 @@ object ModulePositionScanner {
   private def sbtModuleIdRegex(dependency: Dependency): Regex = {
     val g = Regex.quote(dependency.groupId.value)
     val a = Regex.quote(dependency.artifactId.name)
-    raw""""($g)"\s*%+\s*"($a)"\s*%+\s*(.*)""".r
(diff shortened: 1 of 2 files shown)


alejandrohdezma:
Hey @aartigao, thanks for this! Do you think it could be worth adding also a test when using a comment such as `/* my comment */`?

aartigao:
> Hey @aartigao, thanks for this! Do you think it could be worth adding also a test when using a comment such as `/* my comment */`?

Sure!

alejandrohdezma (approved):

alejandrohdezma:
Thank you!!

PR #3482, opened 2024-11-21
Switch to codecov action v4 using a token to upload

mzuehlke:
Similar to https://github.com/fthomas/fs2-cron/pull/532

@fthomas I can not add a secret to the repo


mzuehlke:
Thanks @fthomas

Issue #3445, opened 2024-10-19
Closing pull-requests for retracted versions


mzuehlke:
It happens to the best developers that they release a a new version premature. Scala 3.6.0 was just the last example. 
The https://github.com/scala-steward-org/scala-steward/blob/main/modules/core/src/main/resources/default.scala-steward.conf lists multiple blocked artifacts.

The idea is to close pull requests of artifacts that get listed in `updates.retracted` and/or leave at lest a comment that this version shouldn't be used. The affected should be stored in the `PullRequestRepository`.

@scala-steward-org/core @WojciechMazur @SethTisue @Kordyjan

fthomas:
Great idea! The entries in `updates.retracted` could contain an optional reason why a version was retracted and/or an URL that contains further information.

PR #3488, opened 2024-11-24
Retract previously opened Pull Request

mzuehlke:
The `updates.retraced` section of the `scala-steward.conf` allows to declare updates that should be retracted. Each entry must have a `reason`, a `doc` URL and a list of dependency patterns.

Example:
```
updates.retracted = [
  {
    reason = "Ignore version 3.6.0 as it is abandoned due to broken compatibility",
    doc = "https://contributors.scala-lang.org/t/broken-scala-3-6-0-release/6792",
    artifacts = [
      { groupId = "org.scala-lang", artifactId = "scala3-compiler", version = { exact = "3.6.0" } }
    ]
  }
]
```
Retraction of Pull Request is only possible, if the workspace is persisted correctly. (https://github.com/scala-steward-org/scala-steward/blob/main/docs/faq.md#why-doesnt-self-hosted-scala-steward-close-obsolete-prs)

Fixes: #3445

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -104,9 +104,9 @@ updates.pin  = [ { groupId = "com.example", artifactId="foo", version = "1.1." }
 # Defaults to empty `[]` which mean Scala Steward will not ignore dependencies.
 updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
-# The dependencies which match the given pattern are retracted: Their existing Pull-Request will be closed.
+# The dependencies which match the given pattern are retracted. Their existing pull-request will be closed.
 #
-# Each entry must have a `reason, a `doc` URL and a list of dependency patterns.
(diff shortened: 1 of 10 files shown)


SethTisue:
oh, a thought just occurred to me... and I hate to suggest it so late... suppose a Steward PR gets merged, but the version in question is retracted later. Will the steward leave a comment on the merged PR to notify the maintainers?

mzuehlke:
I haven't rolled this into a released, yet.
But wanted to do this later today.

Sadly I only went for the simpler option. All PRs that are still open will get closed with a comment. Otherwise I would have to keep book which closed PR did already got a retraction comment.

mzuehlke:
1. https://github.com/gemini-hlsw/refined-algebra/pull/95
2. https://github.com/gemini-hlsw/lucuma-core/pull/957

The 1st was closed, the 2nd got a comment even though is was already closed. Maybe the local state of Steward wasn't updated. For this use case even better.

fthomas (commented):
Great work! I like that it so neatly fits into the application.

Issue #3476, opened 2024-11-13
PRs with descriptions longer than 65536 characters exceed GitHub API limits, fail the Scala Steward run


rtyley:
One of our projects is failing our Scala Steward run, because it currently has a large number of dependencies (>75) requiring update, and we're using Scala Steward [grouped PRs](https://github.com/scala-steward-org/scala-steward/pull/2714) - with all those dependency updates in one PR description, the description exceeds the GitHub limit of [**65,536** 4-byte unicode characters](https://github.com/orgs/community/discussions/27190#discussioncomment-3254953), and the GitHub API call to create the PR [fails](https://gist.github.com/rtyley/6ea206f077d24dd98e290fb2661b918c):

```json
{
  "message": "Validation Failed",
  "errors": [
    {
      "resource": "Issue",
      "code": "custom",
      "field": "body",
      "message": "body is too long (maximum is 65536 characters)"
    }
  ],
  "documentation_url": "https://docs.github.com/rest/pulls/pulls#create-a-pull-request",
  "status": "422"
}
```

## PR description size - where are the characters going?

We can break the PR message down into two main parts:

1. **About this PR** - lists all the updated dependencies, with links to release notes, version diffs, etc.
2. **Usage** - contains '🔍 Files still referring to the old version numbers' & '⚙ Adjust future updates', which lists all the dependencies again, twice, once for `updates.ignore` & again for `dependencyOverrides`.

Looking at a large Scala Steward PR with 73 dependency updates, **these two main parts are actually almost equal in size**.

## Possible solutions

* Curtail only the description - so allow the PR to grow huge, but when listing the updated dependencies, only list the first 50 or so and say something like "plus `N` additional dependencies"
* Restrict the actual number of dependency updates - eg, only 50 dependency updates max. Ideally we would have a stable ordering that meant we were consistent in the list of dependency updates included. The PR description should also explain that `N` additional dependency updates have been excluded.

I would favour the second option, restricting the actual number of dependency updates - few projects need to be able to absorb more than 50 updates in a single PR, right?! I think we should proceed with this, and then if requirements come in suggesting more are needed, work on ways to make the description markdown more efficient so that we can raise the quota.

mzuehlke:
This has likely been caused by #3466 
The hidden JSON leads to much more (invisible) content.

PR #3466, opened 2024-11-05
Add JSON metadata in PRs

michaelmior:
This solves #3465. This ended up being quite straightforward as I didn't realize there was already a well-defined JSON serialization for updates. I think the only downside of this is a slightly larger payload that gets sent to GitHub each time a PR is created, but it does enable some interesting use cases of taking automated actions on the resulting PRs.

PR #3492, opened 2024-12-03
Ensure PR description is not too long

mzuehlke:
With the addition of structured metadata to PR description the content was growing too large when multiple updates got grouped together. Now the metadata gets only appended, if the length is still below 65536 characters.

Caused-by: #3466
Fixes: #3476


alejandrohdezma (approved):
LGTM

PR #3515, opened 2025-01-02
Make all values in config classes optional

fthomas:
This changes `RepoConfig` and the other `*Config` classes such that all values are optional and there are no default parameter values except `None`s. This allows us to get rid of circe-generic-extras' `Configuration.default.withDefaults` and `deriveConfiguredCodec` since missing `Option` values do not cause an error during decoding. So `{}` can still be decoded into a `RepoConfig`, and there are a lot of tests in `RepoConfigAlgTest` that ensure that sparse configurations with a lot of missing values can still be decoded. 

Another benefit of this change is that `RepoConfig.show` can now use `deepDropNullValues` so that logged `RepoConfig`s only include stuff which was actually set by a user. For example, before this change, logging a `RepoConfig` was similar to this:
```
INFO  Parsed repo config {
   "commits" : {
     "message" : null
   },
   "pullRequests" : {
     "frequency" : null,
     "grouping" : [
     ],
     "includeMatchedLabels" : null,
     "customLabels" : [
     ]
   },
   "scalafmt" : {
     "runAfterUpgrading" : null
   },
   "updates" : {
     "pin" : [
     ],
     "allow" : [
     ],
     "allowPreReleases" : [
     ],
     "ignore" : [
     ],
     "limit" : null,
     "fileExtensions" : null
   },
   "postUpdateHooks" : null,
   "updatePullRequests" : "always",
   "buildRoots" : null,
   "assignees" : [
   ],
   "reviewers" : [
   ],
   "dependencyOverrides" : [
   ]
 }
```
The same `RepoConfig` is now logged as:
```
INFO  Parsed repo config {
   "updatePullRequests" : "always"
 }
```

Before merging this PR, I'll squash the commits and provide a better commit message, but for potential reviewers, it should be easier to look at each commit individually.

This continues the work started in #3514 of getting rid of circe-generic-extras so that Scala Steward can be built with Scala 3.

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
@@ -19,8 +19,7 @@ package org.scalasteward.core.repoconfig
 import cats.syntax.all._
 import cats.{Eq, Monoid}
 import io.circe.Codec
-import io.circe.generic.extras.Configuration
-import io.circe.generic.extras.semiauto._
+import io.circe.generic.semiauto._
 import io.circe.syntax._
 import org.scalasteward.core.buildtool.BuildRoot
 import org.scalasteward.core.data.Repo
@@ -37,7 +36,7 @@ final case class RepoConfig(
     buildRoots: Option[List[BuildRootConfig]] = None,
     assignees: Option[List[String]] = None,
     reviewers: Option[List[String]] = None,
-    dependencyOverrides: List[GroupRepoConfig] = List.empty,
+    dependencyOverrides: Option[List[GroupRepoConfig]] = None,
     signoffCommits: Option[Boolean] = None
 ) {
   def commitsOrDefault: CommitsConfig =
@@ -64,6 +63,9 @@ final case class RepoConfig(
   def reviewersOrDefault: List[String] =
     reviewers.getOrElse(Nil)
 
+  def dependencyOverridesOrDefault: List[GroupRepoConfig] =
+    dependencyOverrides.getOrElse(Nil)
+
   def postUpdateHooksOrDefault: List[PostUpdateHook] =
     postUpdateHooks.getOrElse(Nil).map(_.toHook)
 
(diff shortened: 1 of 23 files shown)


alejandrohdezma (approved):

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
```diff
@@ -19,41 +19,61 @@ package org.scalasteward.core.repoconfig
...
 import org.scalasteward.core.data.Repo
 import org.scalasteward.core.edit.hooks.PostUpdateHook
 import org.scalasteward.core.repoconfig.RepoConfig.defaultBuildRoots
 
 final case class RepoConfig(
-    commits: CommitsConfig = CommitsConfig(),
-    pullRequests: PullRequestsConfig = PullRequestsConfig(),
-    scalafmt: ScalafmtConfig = ScalafmtConfig(),
-    updates: UpdatesConfig = UpdatesConfig(),
+    commits: Option[CommitsConfig] = None,
```
alejandrohdezma, line 30: **question (non-blocking):** Do we still need the `None` here? Is it just for tests?

fthomas: Yes, that is just for the tests. `RepoConfigAlgTest` would be cluttered with `None`s for example.

fthomas:
This change maybe also enables us to remove [nonExistingUpdatePattern](https://github.com/scala-steward-org/scala-steward/blob/8277c6ad5b59dad1703168c511ba08729dcae85f/modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala#L157-L158) since we now can differentiate between the default value and an empty list that was set by a user.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RepoConfig.scala
alejandrohdezma: Makes total sense. Thanks!

mzuehlke (approved):
This looks like a very systematic change 👍 
Do you think converting the fields of the the case classes into `private val` would be a good idea to prevent accidental accessing them and forgetting the default value ?

PR #2065, opened 2021-04-06
Compile with Scala 3

fthomas:

PR #3519, opened 2025-01-05
Import `MockEffOps` to fix compile errors with Scala 3

fthomas:
This is in preparation for compiling with Scala 3 and fixes errors like this:
```
 -- [E008] Not Found Error: modules/core/src/test/scala/org/scalasteward/core/BuiltinConfigFilesTest.scala:38:9
 36 |      val migrations = artifactMigrationsLoader
 37 |        .loadAll(ArtifactCfg(List(), disableDefaults = false))
 38 |        .runA(initialState)
    |                       ^
    |value runA is not a member of org.scalasteward.core.mock.MockEff[
    |  List[org.scalasteward.core.update.artifact.ArtifactChange]], but could be made available as an extension method.
    |
    |The following import might fix the problem:
    |
    |  import org.scalasteward.core.mock.MockEffOps
```

I think the change in the Scala 3 compiler that makes this import necessary, is ["3. Package prefixes no longer contribute to the implicit search scope of a type"](https://docs.scala-lang.org/scala3/reference/changed-features/implicit-resolution.html).

### DIFF
--- a/modules/core/src/test/scala/org/scalasteward/core/application/ReposFilesLoaderTest.scala
+++ b/modules/core/src/test/scala/org/scalasteward/core/application/ReposFilesLoaderTest.scala
@@ -4,7 +4,7 @@ import munit.CatsEffectSuite
 import org.scalasteward.core.data.Repo
 import org.scalasteward.core.git.Branch
 import org.scalasteward.core.mock.MockContext.context.reposFilesLoader
-import org.scalasteward.core.mock.{MockConfig, MockState}
+import org.scalasteward.core.mock.{MockConfig, MockEffOps, MockState}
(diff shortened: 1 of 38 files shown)


mzuehlke (approved):
Because #2065 isn't merged, yet.
CI only tests Scala 2.13 and not Scala 3.
I guess you verified this locally.

alejandrohdezma (approved):

fthomas:
I cherry-picked this commit from the [`topic/scala3`](https://github.com/scala-steward-org/scala-steward/tree/topic/scala3) branch, so it should be fine. My idea is to break down #2065 into smaller PRs that are uncontroversial, easy to digest, and are about just one topic.

PR #3529, opened 2025-01-08
Mark Scala.js 1.18.0 as dead-on-arrival.

sjrd:
See https://github.com/scala-js/scala-js/issues/5107

---

I copied an old line for 0.6.30, but the exclusion is too broad. We have other artifacts under `org.scala-js` that have separate repos and separate versioning lines. A future version 1.18.0 of any one of those will be ignored by Scala Steward, although it shouldn't.

What is the recommended solution to this issue? Should we come back and *remove* those exclusion entries once a later 1.18.x version gets released? Presumably if Scala Steward sees that 1.18.1 exists, it won't suggest to upgrade to 1.18.0. Or will it?


fthomas:
You could leave all the 1.18.0 lines as they are right now and create new ones for 1.19.0. It looks like that is done for the DOA Scala 3 releases as well.

> Presumably if Scala Steward sees that 1.18.1 exists, it won't suggest to upgrade to 1.18.0. Or will it?

Correct, it won't.

sjrd:
OK, if you don't think that's too much bloat for the config file. I amended the commit to keep the full list.

fthomas (approved):

PR #3533, opened 2025-01-10
Fix "Scala 2 syntax with -Xsource:3" warnings

fthomas:
Since #3532 IntelliJ produced "Scala 2 syntax with -Xsource:3" warnings. This fixes them.


mzuehlke (approved):
Did you fix the manually ?
Can we define some scalafmt or scalafix rules so the are not coming back ?
OR some compiler flags, as we didn't had any warnings before (as far as I know) 
Neither under Scala 2.13 nor Scala 3

exoego (approved):
LGTM

fthomas:
> Can we define some scalafmt or scalafix rules so the are not coming back ?

238a79d30032e179b9c38814e7ca72e20fa38971 should do that.

PR #3511, opened 2024-12-30
Run Mill 0.12 with `--ticker=false`

fthomas:
Closes: scala-steward-org/mill-plugin#64

PR #3538, opened 2025-01-12
Run Mill 0.11 with `--disable-ticker` instead of `--ticker false`

fthomas:
Running Mill 0.11 with `--ticker=false` (which has been added in #3511) produces the following error:
```
Parsing exception Position 1:9, found "=false"
```
The `--ticker <bool>` option has been added to Mill in 0.12.0 if I'm reading [this commit](https://github.com/com-lihaoyi/mill/commit/0c44ffa3e1c197bef13df66baf452b1bf064433d) correctly. We therefore need to use `--disable-ticker` with Mill 0.11 and `--ticker false` with Mill >= 0.12.


mzuehlke (approved):
Thanks for improving of the `mill` support

PR #3546, opened 2025-01-16
Use `allowOrDefault` in `UpdatesConfig.isAllowed`

fthomas:
I forgot to update this reference in #3515. This is a subtle change in semantics. If `allow == Some(Nil)`  `allow.isEmpty` is `false` and `allowOrDefault.isEmpty` is `true`.


mzuehlke (approved):
Sharp eyes 👍

fthomas:
> Sharp eyes

Guess who looked at Scala Steward logs this morning and wondered why Scala updates where logged with `Ignore org.scala-lang:scala-library : 2.13.15 -> 2.13.16 (reason: not allowed by config)` without any such config in sight. :-)

PR #3560, opened 2025-01-21
Rewrite `RepoCacheAlg.checkCache` as for-comprehension

fthomas:
This is a small refactoring that rewrites `checkCache` as for-comprehension which I think is a little bit easier to read. It also removes the second parameter from `skipIfFailedRecently` which was a needless complication.


mzuehlke (approved):
For sure easier to grasp

PR #3570, opened 2025-01-29
Support multiple default resolvers

fthomas:
The default resolver - which is Maven Central if none is provided by the `--default-maven-repo` option - is now used in multiple places where a list of resolvers is expected. It feels like an arbitrary restriction that we only allow one resolver and always wrap that in a list everywhere we use it. Why not allow `--default-maven-repo` to be a repeating option instead so that `Config` has a `List[Resolver]`. This PR does exactly that. It could be useful if a user wants to resolve Scalafmt via resolver A and Gradle dependencies via resolver B for example.

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -4,7 +4,7 @@ All command line arguments for the `scala-steward` application.
 
 ```
 Usage:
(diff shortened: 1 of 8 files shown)


mzuehlke (approved):

exoego (approved):
LGTM

PR #3576, opened 2025-02-03
Use `RepoConfig#signoffCommits` for migrations and hooks

fthomas:
This uses `RepoConfig#signoffCommits` for Scalafix migrations and post-update hooks commits and removes the `signoffCommits` fields from `PostUpdateHook` and `ScalafixMigration` (which were added in #3432). The latter fields concealed that `RepoConfig#signoffCommits` was partially ignored for post-update hooks and completely ignored for Scalafix migrations.


mzuehlke (approved):
Good catch 👍

PR #3589, opened 2025-02-18
Add whitelist organizations option

lenguyenthanh:
This is my attempt to solve https://github.com/scala-steward-org/scala-steward/issues/1468#issuecomment-2660100620. This add a cli option to whitelist organization(s) from `UrlChecker` more details in the first [commit](https://github.com/scala-steward-org/scala-steward/commit/6fd54dcdae3830fb5f1513d4988fa0424a8c44a9)'s message.

Another option is a to add this option in `RepoConfig` which in my opinion, is a better solution than this pr. But I didn't think about it until I was about to make this pr 😅 . I could do a pr for this option if you think it's indeed better.

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -4,7 +4,7 @@ All command line arguments for the `scala-steward` application.
 
 ```
 Usage:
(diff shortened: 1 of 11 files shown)


lenguyenthanh:
I just realized even with my last commit, it is still not cover all possible case, for example it'll doesn't work with private gitlab instance. I'll try to solve this, in the mean time any suggestions are welcomed.

lenguyenthanh:
hello again, I would love to have some feedback from maintainers if this approach is welcomed. I'm opened to any feedback even start it from scratch if necessary.

PR #3598, opened 2025-03-07
Update Scala to 3.6.4

WojciechMazur:
PR submitted automatically by Scala 3 release tooling.


mzuehlke:
There are already ignores for Scala "3.7.0" in place (above the change)
So I would suggest to either remove these ignore or ignore version "3.6.5"

mzuehlke (approved):

PR #3605, opened 2025-03-17
Add artifact migration for scala-uri

stkeky:
Per original repo [lemonlabsuk/scala-uri](https://github.com/lemonlabsuk/scala-uri?tab=readme-ov-file), it's no longer maintained and the new maintained fork is [indoorvivants/scala-uri](https://github.com/indoorvivants/scala-uri).


mzuehlke (approved):
Thanks

Issue #3612, opened 2025-03-21
feature for library authors: testing library updates using Scala Steward to raise draft PRs with specified library updates


rtyley:
At the Guardian, we publish over 40 Scala software libraries, some which are general-purpose (eg. [`etag-caching`](https://github.com/guardian/etag-caching), [`redirect-resolver`](https://github.com/guardian/redirect-resolver)), some quite Guardian-specific (eg. [`content-api-scala-client`](https://github.com/guardian/content-api-scala-client)).  We have dozens of development teams, and don’t use a single monorepo (we have 200 active Scala repos, over half of which are open-source), but instead many interlinked library dependencies - teams expose libraries to other teams.

When I'm changing a library that many Guardian repositories depend upon, I'd often like to know how deeply it affects them - of the 20 or so services using the library, which ones will break, and how? Will their CI fail? To find that out I need to release a preview version of the library (which I do using [`gha-scala-library-release-workflow`](https://github.com/guardian/gha-scala-library-release-workflow/blob/main/docs/making-a-release.md#choose-release-type)), then raise 20 draft PRs, updating the artifact version in each repo, to trigger it's CI.

Scala Steward has a very good engine (`NurtureAlg`) for working out what artifacts a repo uses, updating artifact versions in arbitrary build files (`EditAlg`) and raising PRs (`PullRequestRepository`) - I would just like to be able to tell it what specific artifact update to roll out, rather than let it choose!

My ideal workflow would be something like invoking our [existing scheduled Scala Steward workflow](https://github.com/guardian/scala-steward-public-repos/actions) manually, but passing in [`workflow_dispatch` inputs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-of-onworkflow_dispatchinputs) that let me specify which artifact and version I want updated. Then Scala Steward would raise PRs (only within my organisation, where our Scala Steward is authorised to run!) in draft, and ideally output a list of convenient links to each PR.

I realise this is a use case that might apply to a relatively small number of organisations - but I wonder if it would be possible to work on changes to Scala Steward that are relatively un-obtrusive that facilitate this workflow?

fthomas:
Except the draft part, this may be already possible by using an instance-wide repo config via [`--repo-config`](https://github.com/scala-steward-org/scala-steward/blob/main/docs/help.md) and [`updates.allowPreReleases`](https://github.com/scala-steward-org/scala-steward/blob/main/docs/repo-specific-configuration.md).

PR #3628, opened 2025-04-22
Allow configuring Pull Requests to be 'draft'

rtyley:


fthomas:
Thanks for the PR, @rtyley and @JamieB-gu!

Using a boolean for `pullRequests.draft` is fine with me but wrt to #3612 I wonder if it would better suit your needs if `pullRequests.draft` would be a list of `UpdatePattern` so that draft PRs are only used for updates matching a pattern in that list. Or maybe instead of using a boolean for all PRs we could have a boolean like `pullRequests.preReleasesAsDrafts` so that draft PRs are only used for the updates matching `updates.allowPreReleases`. Both of these would prevent Scala Steward from creating draft PRs for normal regular updates. With `preReleasesAsDrafts` you would also only need to change `updates.allowPreReleases` for #3612 and not flip `pullRequests.draft` from `false` to `true` and back.

fthomas:
I guess with `preReleasesAsDrafts` you probably wouldn't need to change your repo config often. You could set it
```hocon
pullRequests.preReleasesAsDrafts = true
updates.allowPreReleases = [
  { groupId = "com.gu" },
  # all other groupId of your internal libraries
]
```
and then get draft PRs for Guardian pre-releases and regular PRs for everything else.

rtyley:
> I wonder if it would better suit your needs if `pullRequests.draft` would be a list of `UpdatePattern` so that draft PRs are only used for updates matching a pattern in that list. Or maybe instead of using a boolean for all PRs we could have a boolean like `pullRequests.preReleasesAsDrafts` so that draft PRs are only used for the updates matching `updates.allowPreReleases`. Both of these would prevent Scala Steward from creating draft PRs for normal regular updates. With `preReleasesAsDrafts` you would also only need to change `updates.allowPreReleases` for #3612 and not flip `pullRequests.draft` from `false` to `true` and back.

Actually, for #3612, we've already decided we'll have to specifically [generate](https://github.com/guardian/scala-steward-public-repos/blob/d9dcdcecc844ccb41d3ff17d5a4c6ff99f382002/.github/workflows/reusable-scala-steward.yml#L63-L71) a `scala-steward.conf` for whichever update we want to push out - because we already have [a quite extensive `scala-steward.conf`](https://github.com/guardian/scala-steward-public-repos/blob/main/scala-steward.conf) in global use, and the rules in there clash with the actual minimal configuration we want this specific Scala Steward job to run with, which would be like this:

```
updates.allow = [{ groupId = "com.gu", artifactId = "redirect-resolver", version = "0.0.38-PREVIEW.try-out-tool-versions-support.2024-04-26T1649.b33d9ccd" }]
commits.message = "[TEST - please ignore]: Update ${artifactName} from ${currentVersion} to ${nextVersion}"
```

The clashes with elements in our normal config include:

* `pullRequests.frequency` & `dependencyOverrides.pullRequests.frequency` - we normally throttle the rate at which new PRs are raised to avoid overloading the teams that maintain repos. However, this is really unhelpful if we are trying to push out a preview of a new library release - we want all those draft PRs raised _now_ - and not be 'ignored' for days by Scala Steward due to our normal config.
* `pullRequests.grouping` - we have a [dedicated grouping](https://github.com/guardian/scala-steward-public-repos/blob/be6c87aa632837ca0d1c1f221662001b74aa5247/scala-steward.conf#L2) for _our_ dependencies (under the `com.gu` groupId), as they're generally more pertinent and important to our teams. However, when we are pushing out a preview of a new library release, we don't want it to interact with any existing PRs under that grouping- we want our own isolated PRs.

As we already have to generate a specific `scala-steward.conf`, it's no problem for us to have to add or remove a `pullRequests.draft` flag from the configuration!

Just to be clear, we anticipate having a normal scheduled Scala Steward job running with our normal `scala-steward.conf`, and then additionally ad-hoc runs of our Scala Steward workflow running with minimal custom-generated `scala-steward.conf` to perform specific updates as required by developers.

It's also possible, that although I'm focussing on the _preview_-of-a-new-library-release use-case, we might also want to raise draft PRs for _non_-preview releases - it's often the case that updating a dependency version number is not enough to get an update PR to compile - manual code updates may be required - and we might want to put that work on the developer of the update, rather than the teams maintaining the consuming repos, so draft PRs would still be appropriate until the PR is in a state fit for maintaining teams to review. Consequently we would probably prefer `pullRequests.draft` to `pullRequests.preReleasesAsDrafts`?

fthomas (approved):
Thanks for the detailed explanation @rtyley! The simple `pullRequests.draft` boolean (as in this PR) seems like a good fit for what you're trying to achieve.

rtyley:
Thanks for approving the PR @fthomas ! Is it ok to also merge at this point?

PR #3659, opened 2025-06-17
Prevent `version.isSnapshot` from incorrectly identifying versions with hashes as snapshots

jonathonherbert:
## What's changed?

At the moment, [we're attempting to create a workflow](https://github.com/guardian/scala-steward-public-repos/pull/90) to raise PRs against consuming projects when we publish prereleases.  

In attempting to [use `updates.allowPrereleases` to do that](https://github.com/scala-steward-org/scala-steward/issues/3612#issuecomment-2743182650), we've encountered an issue where the commit hash in our prerelease version (e.g. `0.0.37-PREVIEW.try-out-verify-artifact-hashes.2024-04-25T1652.c7fa7c2b`) results in scala-steward misidentifying our release as a snapshot. As a result, it won't consider it as a valid next version.

This PR corrects the behaviour of `version.isSnapshot`, which was incorrectly identifying versions with hashes as snapshots.

As far as we're aware, snapshotted versions are usually identified by a `-SNAPSHOT` suffix, which the case below already covers:
https://github.com/scala-steward-org/scala-steward/blob/ef70026241d282af7ee279b2f64af2d2d7abb451/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala#L173

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Version.scala
@@ -99,10 +99,9 @@ final case class Version(value: String) {
       case _                          => false
     }
 
-  private def isSnapshot: Boolean =
+  def isSnapshot: Boolean =
     components.exists {
       case a: Version.Component.Alpha => a.isSnapshotIdent
-      case _: Version.Component.Hash  => true
       case _                          => false
     }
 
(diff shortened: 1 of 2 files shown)


rtyley:
I couldn't find any evidence that version numbers with hashes in them should always be considered as snapshot releases - here's a counter-example, where a non-snapshot release of an `sbt` dependency contained a hash in its version number:

https://mvnrepository.com/artifact/org.scala-sbt.ivy/ivy/2.3.0-sbt-396a783bba347016e7fe30dacc60d355be607fe2

...so I think this PR is correct: `version.isSnapshot` should _not_ consider a version number to denote a snapshot just because it contains a hash.

PR #3665, opened 2025-06-25
Tighten up http4s-0.19.0 ignores

rossabaker:
http4s has multiple repos, and some of them have or will have 0.19.0 tags that were not affected by the ill-fated v0.19.0 on http4s/http4s.  This makes the list of artifacts explicit, and switches to exact version numbering.

### DIFF
--- a/modules/core/src/main/resources/default.scala-steward.conf
+++ b/modules/core/src/main/resources/default.scala-steward.conf
@@ -242,7 +242,23 @@ updates.ignore = [
   { groupId = "net.sourceforge.plantuml", artifactId = "plantuml", version = "2017." },
 
   // https://github.com/http4s/http4s/pull/2153
-  { groupId = "org.http4s", version = "0.19.0" },
+  { groupId = "org.http4s", artifactId = "http4s-argonaut", version = { exact = "0.19.0" } },
+  { groupId = "org.http4s", artifactId = "http4s-async-http-client", version = { exact = "0.19.0" } },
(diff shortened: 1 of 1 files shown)


rossabaker:
I found this because several sbt-http4s-org updates are [ignored by config](https://github.com/http4s/steward/actions/runs/15886414995/job/44799377461#step:6:99)

```
2025-06-25 20:25:14,518 INFO  Ignore org.http4s:sbt-http4s-org : 0.18.0 -> 1.0.1 -> 2.0.0 (reason: ignored by config)
```

I don't think this rule is the reason, because I think it's a prefix match?  But I don't see what else it would be, and I think this change is correct regardless: we don't want some future, _unrelated_ http4s 0.19.0 artifacts to be ignored.

fthomas:
> But I don't see what else it would be

I suspect it is because of entries like this: https://github.com/http4s/blaze/blob/d03480f5a00992ad87eaa3ea1849387040cad952/.scala-steward.conf#L9

rossabaker:
You're right.  The project I was looking at _did_ get its update, and the line I zeroed in on was from a different branch.  Sorry for the noise.

I still think this PR is a more precise solution to the issue that config references, but it's also a lot more verbose.

fthomas:
> I still think this PR is a more precise solution to the issue that config references, but it's also a lot more verbose.

I don't mind the verbosity.

Thanks for the PR! Merging.

PR #3674, opened 2025-07-17
Retract smithy4s 0.18.38

kubukoz:


mzuehlke:
You probably want to add these artifacts to the ignore list, too.
This isn't happening automatically.

Example: https://github.com/scala-steward-org/scala-steward/pull/3595/files

kubukoz:
thanks, good idea

PR #3694, opened 2025-09-04
Bump default maxBufferSize

clairemcginty:
See discussion [here](https://github.com/scala-steward-org/scala-steward/pull/3282#issuecomment-3250222315); some repos are still overflowing the buffer size and not receiving scala-steward updates.


clairemcginty:
cc @mzuehlke would you mind taking a look? :)

mzuehlke:
Thanks @clairemcginty 👍

PR #3714, opened 2025-09-24
ignore http4s 0.23.31

samspills:
- not sure if I have all the correct packages here, is it necessary to ignore everything in the same release?

### DIFF
--- a/modules/core/src/main/resources/default.scala-steward.conf
+++ b/modules/core/src/main/resources/default.scala-steward.conf
@@ -332,6 +332,20 @@ updates.ignore = [
 
   // Ignore munit 1.1.2, as it removes some traits required for binary compatibility with munit-scalacheck
   { groupId = "org.scalameta", artifactId = "munit", version = "1.1.2" }
+
+  // Ignore http4s 0.23.31, as it broke binary compatibility
+  { groupId = "org.http4s", artifactId = "http4s-circe", version = { exact = "0.23.31" } },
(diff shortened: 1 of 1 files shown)


rossabaker:
Only http4s-dsl-0.23.31 is bad, but I don't know what ignoring just that one does when these all come from the same repository.

rossabaker (approved):
I think this is bigger than it needs to be, but I am confident it does the right thing and I don't understand the full ramifications of calling out just the http4s-dsl, so, :+1:.

mzuehlke:
> I think this is bigger than it needs to be, but I am confident it does the right thing and I don't understand the full ramifications of calling out just the http4s-dsl, so, 👍.

I think if you define the versions individually like:
```Scala
val http4s = Seq(
  "org.http4s" %% "http4s-blaze-server" % "0.21.0-M6",
  "org.http4s" %% "http4s-circe"        % "0.21.0-M6",
  "org.http4s" %% "http4s-dsl"          % "0.21.0-M6",
  "io.circe"   %% "circe-derivation"    % "0.12.0-M7" % Compile,
  "io.circe"   %% "circe-fs2"           % "0.12.0"
)
```
and we would only ignore `http4s-dsl`, Steward would update the other artifacts.

rossabaker:
We probably would have been fine as long as nothing pulled in something that depends on http4s-dsl (i.e., http4s-core would have been fine.). But this requires is less thinking, and it was superseded within a day, so I think we did the right thing here.

PR #3731, opened 2025-10-07
feat: Add mise.toml to supported defaults

henricook:
Also: Cover some mise-specific behaviour - like not updating an entry that says 'latest' to tests

https://github.com/scala-steward-org/scala-steward/issues/3730


exoego (approved):
LGTM

I also use mise in my work projects since it is faster and more capable than other tools in this area (e.g. asdf, sdkman).

PR #3734, opened 2025-10-12
Update Mill dependencies

lefou:
There are newer versions, but the artifactIds have changed, so we didn't detect it in automation.


lefou:
Can I get a review, please? @fthomas, maybe.

I don't understand the test failure. It looks unrelated to these changes.

lefou:
@mzuehlke That helped a lot. Thank you!

lefou:
Yes. This is just the update to the dependencies, so that Scala Steward can automatically pick up newer versions when available.

PR #3752, opened 2025-11-11
Add scalafix for `weaver-test` `0.11.0`

zainab-ali:
Note that version `0.11.0` hasn't been released yet, but the scalafix rules for it are [here](https://github.com/typelevel/weaver-test/blob/main/scalafix/rules/src/main/resources/META-INF/services/scalafix.v1.Rule).

I assume we should update the scala steward config _before_ releasing `0.11.0`


mzuehlke:
Thanks

PR #3762, opened 2025-11-28
Add dependency-cooldowns / minimum-artifact-age

rtyley:
_@emdash-ie & myself (@rtyley) are co-authors of this work._

This PR adds a new `updates.cooldown` config option to allow configuring Scala Steward to defer suggesting an update to a new version of a dependency, until that version has survived a 'cooldown' period (eg 7 days).

This intended to be useful to users who want to reduce the risk of updating to a malicious new version of a dependency before it's been examined by their supply-chain security vendor, addressing:

* https://github.com/scala-steward-org/scala-steward/issues/3757

Note that applying a cooldown does not necessarily mean that you will get _fewer_ updates - just that any update offered will be at least as old as specified.

Artifacts that are _too recent_ will get rejected with a `TooRecentForCooldown` rejection reason.

## Internal changes

* `VersionsCache` now persists `VersionWithFirstSeen` rather than `Version`, allowing us to record when a new version was first seen by this Scala Steward instance (the only simple & reliable way to find out how old a Maven artifact is!). We've added `VersionsCacheTest` to ensure we can still decode both old and new versions of the cache data.
* `ArtifactUpdateCandidates`
  * now stores `newerVersionsWithFirstSeen: Nel[VersionWithFirstSeen]`, so that the candidate versions can be evaluated by age.
  * gains two convenience methods, `filterVersions()` & `filterVersionsWithFirstSeen()`, which filter the versions of the update and return an `Option[ArtifactUpdateCandidates]`, with a value of `None` if no versions are left - this extracts that logic from several places in the Scala Steward codebase (`isAllowed`, `isPinned`, `isIgnored` & `scalaLTSFilter`)

## Config

Right now there's a simple single value that can be configured - this is the cooldown period for all artifacts:

```
updates.cooldown = {
  minimumAge: "7 days"
}
```

### Future config

Later on, in a subsequent PR, we might allow syntax like this, for example stating that we want all Guardian updates promptly, but all other artifacts only once they're 7 days old:

```
updates.cooldown = {
  minimumAge: "7 days",
  overrides: [
    { minimumAge: "1 day", [{groupId = "com.gu"}] }
  ]
}
```

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -109,6 +109,15 @@ updates.pin  = [ { groupId = "com.example", artifactId="foo", version = "1.1." }
 # Defaults to empty `[]` which mean Scala Steward will not ignore dependencies.
 updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 
+# Scala Steward will ignore dependency update versions while they are newer than
+# the specified minimum age.
+#
+# Scala Steward records the first time it sees any artefact version and calculates
+# the artefact version's age from that time.
+updates.cooldown = {
+  minimumAge: "7 days"
+}
+
 # The dependencies which match the given pattern are retracted. Their existing pull-request will be closed.
 #
 # Each entry must have a `reason`, a `doc` URL and a list of dependency patterns.
--- a/modules/benchmark/src/main/scala/org/scalasteward/benchmark/UpdatesConfigBenchmark.scala
+++ b/modules/benchmark/src/main/scala/org/scalasteward/benchmark/UpdatesConfigBenchmark.scala
@@ -20,7 +20,8 @@ import java.util.concurrent.TimeUnit
 import org.openjdk.jmh.annotations.{Benchmark, BenchmarkMode, Mode, OutputTimeUnit}
 import org.scalasteward.core.data.*
 import org.scalasteward.core.repoconfig.{UpdatePattern, UpdatesConfig, VersionPattern}
-import org.scalasteward.core.util.Nel
+import org.scalasteward.core.util.{Nel, Timestamp}
+import org.scalasteward.core.coursier.VersionsCache.VersionWithFirstSeen
 
 @BenchmarkMode(Array(Mode.AverageTime))
 class UpdatesConfigBenchmark {
@@ -32,13 +33,14 @@ class UpdatesConfigBenchmark {
     val dependency = CrossDependency(Dependency(groupId, ArtifactId("artifact"), Version("1.0.0")))
     val newerVersions = Nel
       .of("2.0.0", "2.1.0", "2.1.1", "2.2.0", "3.0.0", "3.1.0", "3.2.1", "3.3.3", "4.0", "5.0")
-      .map(Version.apply)
+      .map(v => VersionWithFirstSeen(Version(v), None))
     val update = ArtifactUpdateCandidates(ArtifactForUpdate(dependency), newerVersions)
+    val currentTime = Timestamp(millis = 0)
 
-    UpdatesConfig().keep(update)
-    UpdatesConfig(allow = Some(List(UpdatePattern(groupId, None, None)))).keep(update)
+    UpdatesConfig().keep(update, currentTime)
+    UpdatesConfig(allow = Some(List(UpdatePattern(groupId, None, None)))).keep(update, currentTime)
     UpdatesConfig(allow =
       Some(List(UpdatePattern(groupId, None, Some(VersionPattern(prefix = Some("6.0"))))))
-    ).keep(update)
+    ).keep(update, currentTime)
   }
 }
--- a/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/buildtool/sbt/SbtAlg.scala
@@ -129,7 +129,7 @@ final class SbtAlg[F[_]](defaultResolvers: List[Resolver], ignoreOptsFiles: Bool
   private def latestSbtScalafixVersion: F[Option[Version]] =
     versionsCache
       .getVersions(Scope(sbtScalafixDependency, defaultResolvers), None)
-      .map(_.lastOption)
+      .map(_.lastOption.map(_.version))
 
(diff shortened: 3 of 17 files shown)


modules/core/src/main/scala/org/scalasteward/core/persistence/JsonKeyValueStore.scala
```diff
@@ -54,8 +54,11 @@ final class JsonKeyValueStore[F[_], K, V](storeRoot: File, name: String)(implici
     }
   }
 
-  private def jsonFile(key: K): File =
-    storeRoot / keyEncoder(key) / s"$name.json"
+  private def jsonFile(key: K): File = {
+    val file = storeRoot / keyEncoder(key) / s"$name.json"
+    println(s"file is ${file}")
```
mzuehlke: leftover debug println

mzuehlke (changes requested):
The test are failing and an addition to the default config to document this feature is still missing. 
The feature itself looks good 👍

rtyley:
> The test are failing and an addition to the default config to document this feature is still missing.
> The feature itself looks good 👍

Thanks @mzuehlke - just to be clear, this PR is not yet ready for review, as you've spotted, the tests are still failing and it's not fully implemented - I think @exoego may have set this PR to ready-for-review by mistake!

Looking forward to continuing work on the PR soon.

rtyley:
Added this small issue that @emdash-ie & I discussed while working on this PR:

* https://github.com/scala-steward-org/scala-steward/issues/3779

mzuehlke (approved):

mzuehlke:
Thanks for you support with this and other PRs @emdash-ie

kubukoz:
Hi @rtyley, I'm interested in adding more fine-grained configuration. I'll have a draft PR soon.

kubukoz:
> allowing us to record when a new version was first seen by this Scala Steward instance (the only simple & reliable way to find out how old a Maven artifact is!)

@rtyley 👋 

I'm interested in knowing more about this. Relying on the instance's memory seems quite constraining, maybe we could use a heuristic that tries to read the times from the directory listing / POMs? Then we pick the earliest out of those and the first time we've seen it. (if we haven't seen it before, it's surely more recent than our most recent run).

Even if e.g. a mirror uses a more recent date, this would be somewhat better than just relying on the one instance's state. Thoughts?

kubukoz:
BTW. this is now documented on https://cooldowns.dev/#scala-steward

rtyley:
> Relying on the instance's memory seems quite constraining

Just to clarify, what are the pain points with relying on just the first-seen-store approach? Are you concerned about storage capacity/durability, or perhaps cases where it might be possible to allow an artifact-version sooner if Scala Steward had evidence from other sources?

At the moment, if people are using cooldowns of 1 or 2 weeks, it doesn't seem that awful to wait a bit longer for first-seen to age-up to get the latest update? Are there cases where that's tiresome/painful? Are we generally _that_ eager to get new, non-security-related, updates?

> maybe we could use a heuristic that tries to read the times from the directory listing / POMs? Then we pick the earliest out of those and the first time we've seen it. (if we haven't seen it before, it's surely more recent than our most recent run).

I think in general we don't want to trust any metadata in the pom.xml file itself, as that's under the control of the library publisher, who for the purposes of this check we do not trust.

The directory listing has always seemed a fairly sketchy way of [getting release dates](https://stackoverflow.com/q/69790966/438886) to me: _"The raw directory listing of an artifact version folder - assuming you can access it - will often give a timestamp against each file- however, this has no standard format, and it's position and datetime format will vary depending on how the webserver is set up"_ - there is no real guarantee as to the format of the data here, it's just what that particular version of [`mod_autoindex`](https://httpd.apache.org/docs/current/mod/mod_autoindex.html) or [`ngx_http_autoindex_module`](https://nginx.org/en/docs/http/ngx_http_autoindex_module.html) or whatever is configured to return:

* https://repo1.maven.org/maven2/software/amazon/awssdk/s3/2.17.71/ gives a raw text grid, with `2021-10-29 20:53` in the 2nd column (missing a timezone, which would have been nice)
* https://repo.clojars.org/Hermes/Hermes/0.1.0-SNAPSHOT/ _used_ to give `2020-02-09T14:36:26.000Z` in the 1st column, but now has changed to `2020-02-09 14:36` in the 2nd column

Having said that, it hasn't stopped Dependabot from [using that approach](https://github.com/dependabot/dependabot-core/blob/0353ac2cd58e1632451c8ca6ad2dd9f21960e8fe/maven/lib/dependabot/maven/shared/shared_package_details_fetcher.rb#L158-L172).

First-seen seems ok to me personally, but I may not be feeling the pain points that you are, so I'd like to know more about them.

It is exasperating that there isn't a better standard for recording release-dates in the Maven ecosystem. At this point, a distributed, verifiable release log might be the ideal long-term solution - something like https://certificate.transparency.dev/howctworks/ .

kubukoz:
> Just to clarify, what are the pain points with relying on just the first-seen-store approach? Are you concerned about storage capacity/durability, or perhaps cases where it might be possible to allow an artifact-version sooner if Scala Steward had evidence from other sources?

Mostly worried about internal usage in which we run Steward on a weekly basis. It seems like introducing any non-zero cooldown means that we'll be effectively cooling down by at least a week, even if e.g. it would've been allowed sooner (imagine something gets released on a saturday, we run on friday, so a 6-day release will be seen as fresh). It errs on the side of safety, though, so I guess it's fine.

> I think in general we don't want to trust any metadata in the pom.xml file itself, as that's under the control of the library publisher, who for the purposes of this check we do not trust.

Fair enough, I hadn't considered that this is not validated by Maven.

> First-seen seems ok to me personally, but I may not be feeling the pain points that you are, so I'd like to know more about them.

We're discussing what the optimal config would be internally, and some teams are worried that we'll have to wait too much for updates to propagate.

mzuehlke:
The majority of pull request I have seen are referring to Github releases. 
We already try to get the release for the pul request description. 
Maybe this would be a better source for the real release date ?

kubukoz:
That would make sense! If it's not available, we can always fall back to the "seen" date.

Edit: then again, the releases are probably relying on `scmInfo`, which could be spoofed too. An attacker could make it point to their own repo, with the release tag having been published before the attack.

PR #3765, opened 2025-12-05
Bump max-buffer-size to 65536

kellen:
Spotify's Scio continues to have issues with smaller buffer sizes; this bumps the default max-buffer-size to 65536

cc @clairemcginty

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -53,7 +53,7 @@ Options and flags:
     --disable-sandbox
         Whether to not use the sandbox
     --max-buffer-size <integer>
-        Size of the buffer for the output of an external process in lines; default: 49152
+        Size of the buffer for the output of an external process in lines; default: 65536
     --repo-config <uri>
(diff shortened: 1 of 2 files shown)


rtyley (approved):
Looks like this follows on from https://github.com/scala-steward-org/scala-steward/pull/3694 - see also this issue for a suggestion allowing bigger buffers:

* https://github.com/scala-steward-org/scala-steward/issues/3620

Happy to bump this for now!

Incidentally, the fact that the buffer size is expressed in lines, rather than _bytes_, makes it a bit difficult to judge what a sensible limit should be - obviously, lines can be any length, so while on modern systems 65536 is a small number of bytes, or even kilobytes, I've no idea whether 65536 lines is 'big' or not!

mzuehlke:
If I understand the code correctly the parameter is used for limiting the the length of each line and for limiting the number of lines.
Maybe increasing the limit further is a good idea 👍

PR #3787, opened 2026-01-15
Refactor: Separate modelling of candidate `newerVersions` from `Update`s

### LINKED ISSUE #3781: `PullRequestRepository` unnecessarily persists `newerVersions` - ie the data about what possible versions might have been eligible
`PullRequestRepository` persists `Entry`s that include the `Update` type:

https://github.com/scala-steward-org/scala-steward/blob/19e294bd534f8b42efd46e5e4cb9df2880c0d485/modules/core/src/main/scala/org/scalasteward/core/nurture/PullRequestRepository.scala#L146-L149

...and it's reasonable to want to know what conceptual `Update` a pull request represented! However, it includes more information than necessary (and this becomes a little troublesome/distracting with #3762 ) - `Single` is a sub-trait of `Update`, and it includes the `newerVersions` field:

https://github.com/scala-steward-org/scala-steward/blob/63a047628fcb9a801cf5c2695d03f845d7173f68/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala#L50-L60

...so `PullRequestRepository` persists `newerVersions` - ie it is storing what possible update versions were considered when creating the PR, which is _not_ the most useful thing- we only really want to know what actual version the PR went with.

The reason this is troublesome with #3762 is that that PR adds additional metadata for versions - ie when the Version was first seen - so we are now trying to persist yet _more_ superfluous Version metadata with the `PullRequestRepository`.

rtyley:
_Addresses https://github.com/scala-steward-org/scala-steward/issues/3781 - `Update.Single` doesn't need `newerVersions`, just `nextVersion`_

The Scala Steward codebase deals with the concept of 'updates' in two areas:

* **Multiple `newerVersions`**: Initially, evaluating _candidate_ updates (in `UpdateAlg` & `FilterAlg`), where for any given artifact, there can be many possible versions we might want to upgrade to. The list of versions is reduced down with filtering (based on config & other rules), and eventually a single next version is selected.
* **Single `nextVersion`**: Subsequently, retrieving information on the updates represented by the pull requests Scala Steward created - each artifact being updated to a specific _single_ new version.

...Scala Steward used the `Update` type for _both_ areas (particularly, `Update.ForArtifactId` for the first phase, _requiring_ it to support multiple `newerVersions`).

Looking at a single `Update.ForArtifactId` in code, it's not always clear to the newcomer if the update contains many newer versions or not - ie where we are in the pipeline.

## Changes

* `Update.Single` types (eg `Update.ForArtifactId`) now just have a `nextVersion` field, not `newerVersions`, which better models what they actually signify: the single specific update selected for a particular PR
* Within `UpdateAlg` & `FilterAlg` we change `Update.ForArtifactId` → `ArtifactUpdateCandidates`🆕 (which _does_ have a `newerVersions` field) for the parts where multiple candidate versions _are_ being evaluated
* `Update.ForArtifactId` & `ArtifactUpdateCandidates` both extend `ArtifactUpdateVersions`🆕 to allow `UpdatePattern.findMatch()` to filter lists of either of them:
  * Filtering the candidate versions of `ArtifactUpdateCandidates` in `UpdatesConfig.isAllowed`, etc
  * Filtering the `Update`s of existing PRs in `PruningAlg.newPullRequestsAllowed()` & `RetractedArtifact.isRetracted()`
* The fields from `Update.ForArtifactId` also required by `ArtifactUpdateCandidates` have been placed in the `ArtifactForUpdate`🆕 case class, avoiding possible duplication 

Remaining unchanged:

* `PullRequestRepository` still persists instances of `Update`
* The JSON serialisation format of `Update` remains unchanged for now (the encoder & decoders have been adapted so that they still output the same format, despite the case classes having changed, just with a _single_ 'newerVersion'), so there's currently no need to introduce a new fallback 'legacy' decoder (several of which were recently deleted with #3785). This also means the JSON metadata added to PR descriptions (originally introduced in https://github.com/scala-steward-org/scala-steward/pull/3466) also remains the same for now, though I think it will be reasonable to adjust the format for `Update.ForArtifactId` to something more logical in future.

With this PR, the changes from #3762 will no longer need to change the data stored in `PullRequestRepository`.

### DIFF
--- a/modules/benchmark/src/main/scala/org/scalasteward/benchmark/UpdatesConfigBenchmark.scala
+++ b/modules/benchmark/src/main/scala/org/scalasteward/benchmark/UpdatesConfigBenchmark.scala
@@ -33,7 +33,7 @@ class UpdatesConfigBenchmark {
     val newerVersions = Nel
       .of("2.0.0", "2.1.0", "2.1.1", "2.2.0", "3.0.0", "3.1.0", "3.2.1", "3.3.3", "4.0", "5.0")
       .map(Version.apply)
-    val update = Update.ForArtifactId(dependency, newerVersions)
+    val update = ArtifactUpdateCandidates(ArtifactForUpdate(dependency), newerVersions)
 
     UpdatesConfig().keep(update)
     UpdatesConfig(allow = Some(List(UpdatePattern(groupId, None, None)))).keep(update)
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala
@@ -23,6 +23,43 @@ import org.scalasteward.core.repoconfig.PullRequestGroup
 import org.scalasteward.core.util
 import org.scalasteward.core.util.Nel
 
+case class ArtifactForUpdate(
+    crossDependency: CrossDependency,
+    newerGroupId: Option[GroupId] = None,
+    newerArtifactId: Option[String] = None
+) {
+  private val headDependency: Dependency = crossDependency.head
+  def groupId: GroupId = headDependency.groupId
+  def artifactId: ArtifactId = headDependency.artifactId
+  def currentVersion: Version = headDependency.version
+}
+
+trait ArtifactUpdateVersions {
+  val artifactForUpdate: ArtifactForUpdate
+
+  val refersToUpdateVersions: Nel[Version]
+
+  def show: String
+}
+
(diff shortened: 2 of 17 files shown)


mzuehlke (approved):
I like these more specific classes. 
Everyone will benefit from easier code understanding-

rtyley:
> I like these more specific classes.
> Everyone will benefit from easier code understanding

Many thanks @mzuehlke !

Issue #1481, opened 2020-06-14
Support for projects unlinked from root


blast-hardcheese:
For some time, I was wondering why only some libraries on https://github.com/twilio/guardrail/ were receiving library update PRs.

After digging in a bit, running `stewardDependencies` locally, I discovered that due to `stewardDependencies` collecting `libraryDependencies` from the project it was launched in, if there were any projects that were only accessible fully qualified, they'd be completely left out of analysis.

To put it differently,
```
sbt> stewardDependencies // Only lists a couple dependencies, defined on `root`
sbt> sampleAkkaHttp/stewardDependencies // Finds dependencies that need to be updated
```

The reason for this is surely due to my lack of how to properly structure an SBT project to serve my needs, but the guardrail integration tests are run in subprojects that are only able to be `compile`d during integration tests, as all other times the required source files have not been generated yet.

I was able to play a little shell game with my dependencies, which works pretty decently: https://github.com/twilio/guardrail/pull/664

... though I wonder if there's not a more elegant solution for this problem, perhaps `.steward.conf` containing `projects = foo, bar, baz` or similar.

This seems somewhat related to https://github.com/fthomas/scala-steward/issues/1158 as well, though that issue seems to be with multiple, separate invocation roots, whereas mine is just separate command prefixes plus an aggregation phase.

As always, thank you for this excellent tool, and have a fantastic day writing Scala!

fthomas:
You're right that Scala Steward only updates dependencies that are listed by `libraryDependencies` of the root project. Running that task in the root project will only run it in subprojects that are aggregated by the root project. I'm not aware of a way to list dependencies for all subprojects regardless whether they are aggregated by the root project or not. If it exists I'm open to discussing if Scala Steward should be changed in this direction.

I also think that https://github.com/twilio/guardrail/pull/664 looks like a good workaround when you don't want certain subprojects aggregated in the root project but still want to benefit from Scala Steward's automatic updates. The good thing about this solution is that library definitions are still only in the sbt build and do not need to be duplicated in some Scala Steward specific config file.

eugeniyk:
@blast-hardcheese what we endup with giter8 templates (same - empty root dependencies, but lot's inside) is to regenerate dependencies.gen.sbt inside root that will aggregate them (simple bash scripts ;)

Then Scala steward update in both places - root (gen) and under

exoego:
IIUC, this is covered by repo-specific config
```
[...]

PR #3818, opened 2026-02-16
Support for non-aggregated projects

satorg:
One more step towards #1481 resolution.
Although it probably does not resolve it completely – this PR only tackles on SBT builds.

Introduces a "sub-project" part for the `buildRoots` configuration option:
```
# It is possible to have multiple scala projects in a single repository. In that case the folders containing the projects (build.sbt folders)
# are specified using the buildRoots property. Also, an inner subproject for every build root can be specified after `:` (colon).
# It can be useful when there are subprojects in the build that are not aggregated into the root project. The subproject part cannot be empty.
# If there's no subproject part (the default) then the root project is assumed.
# Note that the paths used there are relative and if the repo directory itself also contains a build.sbt the dot can be used to specify it.
# Default: ["."]
buildRoots=[ ".", "subfolder/projectA", ".:subProjectX", "subfolder/projectB:subprojectY" ]
```

The `:` (colon) symbol is used as a separator between the `relativePath` and `subProject` parts in the configuration.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -152,6 +152,7 @@ lazy val core = myCrossProject("core")
       Dependencies.logbackClassic % Runtime,
       Dependencies.catsLaws % Test,
       Dependencies.circeLiteral % Test,
+      Dependencies.circeTesting % Test,
       Dependencies.disciplineMunit % Test,
       Dependencies.http4sDsl % Test,
       Dependencies.http4sEmberServer % Test,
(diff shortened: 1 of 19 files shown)


satorg:
I'm not sure that non-aggregated (yet embedded) sub-projects are possible in all other build systems. So in order to add it there I may need additional time to find it out. But I'd appreciate your feedback on this part – let me know if I got it in the right way please. Thank you!

PR #3852, opened 2026-03-24
WIP: Group updates by Edit

rtyley:
This fixes two issues:

* **Multiple PRs with the same diff**: Only one PR will raised, rather than several PRs with identical diffs like these ones:
  * https://github.com/guardian/play-json-extensions/pull/13 - Maven groupId: `com.fasterxml.jackson.core`
  * https://github.com/guardian/play-json-extensions/pull/14 - Maven groupId: `com.fasterxml.jackson.datatype`
* **Inaccurate PR descriptions**: PR descriptions are more accurate, in that they reflect _all_ artifacts that are getting updated, fixing https://github.com/scala-steward-org/scala-steward/issues/3780

See also:

* https://github.com/scala-steward-org/scala-steward/issues/25
* https://github.com/scala-steward-org/scala-steward/issues/3803
* https://github.com/scala-steward-org/scala-steward/pull/2019 - previous attempt at addressing this problem

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/data/Update.scala
@@ -19,10 +19,11 @@ package org.scalasteward.core.data
 import cats.Order
 import cats.implicits.*
 import io.circe.{Decoder, Encoder}
+import org.scalasteward.core.coursier.VersionsCache.VersionWithFirstSeen
+import org.scalasteward.core.nurture.UpdatesForGivenEdit
 import org.scalasteward.core.repoconfig.PullRequestGroup
 import org.scalasteward.core.util
 import org.scalasteward.core.util.Nel
-import org.scalasteward.core.coursier.VersionsCache.VersionWithFirstSeen
 
 case class ArtifactForUpdate(
     crossDependency: CrossDependency,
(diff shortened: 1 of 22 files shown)


modules/core/src/main/scala/org/scalasteward/core/nurture/NurtureAlg.scala
```diff
@@ -35,6 +36,18 @@ import org.scalasteward.core.util.{Nel, UrlChecker}
 import org.scalasteward.core.{git, util}
 import org.typelevel.log4cats.Logger
 
+case class ApplicableUpdateSet(edits: List[Substring.Replacement], artifactUpdates: List[Update.ForArtifactId]) {
```
emdash-ie: I’d like to propose `EquivalentUpdateSet` or `UpdatesForGivenEdit` as the name for this instead of `ApplicableUpdateSet`. I find the existing name hard to immediately find a meaning for when I read it, and I think the most important thing about these sets of updates is that they’re produced by the same edits.

rtyley: Ah good suggestion Emily! I like `UpdatesForGivenEdit`, I've done this with commit 1574d0ce74a7b292f4c4294ce1dc5cb89588bae7.

modules/core/src/main/scala/org/scalasteward/core/data/Update.scala
```diff
@@ -260,18 +281,34 @@ object Update {
...
   implicit val SingleOrder: Order[Single] =
     Order.by((u: Single) => (u.crossDependencies, u.nextVersion))
 
+  implicit val GroupedOrder: Order[Grouped] =
+    Order.by((u: Grouped) => (u.name, u.asSingleUpdates))
+
+  implicit val UpdateOrder: Order[Update] = Order.by {
+    case g: Grouped => Left(g)
+    case s: Single  => Right(s)
+  }
```
rtyley, line 315: `org.scalasteward.core.data.Update.groupByGroupId()` delivers sorted updates - now the more general `Update` rather than `Update.Single` - so we need to be able to sort instances of `Update`.

modules/core/src/main/scala/org/scalasteward/core/edit/EditAlg.scala
```diff
@@ -44,7 +44,17 @@ final class EditAlg[F[_]](implicit
...
   def applyUpdate(
+      data: RepoData,
+      update: Update,
+      preCommit: F[Unit] = F.unit
+  ): F[List[EditAttempt]] = update.on(
+    update = applySingleUpdate(data, _, preCommit),
+    grouped = preCommit >> _.updates.flatTraverse(applySingleUpdate(data, _))
+  )
+
+  private def applySingleUpdate(
```
rtyley, line 57: This generalises `org.scalasteward.core.edit.EditAlg.applyUpdate()` to accept the general `Update` type, rather than just the `Update.Single`.

PR #3880, opened 2026-05-13
Add per-pattern overrides to cooldown config

kubukoz:
## Summary

Follows up on #3762 - a single global cooldown is a good start, but I think we may want to be more lenient on some repositories, especially in closed ecosystems.

Adds `updates.cooldown.overrides`, allowing per-`groupId` (or `groupId` + `artifactId`) `minimumAge` values that override the default. The first matching override wins; otherwise the top-level `minimumAge` is used.

```hocon
updates.cooldown = {
  minimumAge = "7 days"
  overrides = [
    { groupId = "org.typelevel", minimumAge = "3 days" },
    { groupId = "org.acme", artifactId = "foo", minimumAge = "1 day" }
  ]
}
```

(I actually didn't see before opening the PR that @rtyley ended up with the same design as what I imagined) :) 

Notes:
- Top-level `minimumAge` remains required as the default.
- Override patterns reuse `UpdatePattern` for consistency with `pin` / `allow` / `ignore`.
- When merging global + repo configs: top-level `minimumAge` keeps the prior repo-wins behavior; `overrides` lists are concatenated with repo entries first, deduplicated by pattern.

## Test plan

- [x] `CooldownConfigTest` covers override match, fallback, first-match-wins, merge ordering, dedup, and one-sided merges
- [x] Existing `FilterAlgTest` and `UpdatesConfigTest` still pass
- [x] Confirm mdoc-generated `docs/repo-specific-configuration.md` regenerates cleanly in CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/repo-specific-configuration.md
+++ b/docs/repo-specific-configuration.md
@@ -114,8 +114,16 @@ updates.ignore = [ { groupId = "org.acme", artifactId="foo", version = "1.0" } ]
 #
 # Scala Steward records the first time it sees any artefact version and calculates
 # the artefact version's age from that time.
+#
+# Optionally, the minimum age can be overridden for specific groupIds (or
+# groupId + artifactId). The first matching override wins; otherwise the
+# top-level `minimumAge` is used as the default.
 updates.cooldown = {
   minimumAge: "7 days"
+  overrides = [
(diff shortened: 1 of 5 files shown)


kubukoz:
Superseded by #3881

PR #3881, opened 2026-05-15
Allow overrding dependency cooldown period

iRevive:
We don't need a cooldown period for internally published dependencies, where we fully control the supply chain.

## Motivation

We use `updates.cooldown` as a global safeguard, but that is too coarse for company-internal dependencies. Internal artifacts often publish frequently and under our control, so we may want a shorter cooldown, or no   cooldown at all, while still keeping a stricter minimum age for external dependencies.

This PR makes cooldown configurable at the dependency-override level, so repos can treat internal dependencies differently without weakening the default policy for everything else.

## Example config

```hocon
  updates = {
    cooldown = {
      minimumAge = "7 days"
    }
  }

  dependencyOverrides = [
    {
      dependency = {
        groupId = "com.my-company"
        artifactId = "platform-core"
        version = "2."
      }
      cooldown = {
        minimumAge = "2 days"
      }
    }
  ]
```

In this example:

- external dependencies still use the default 7 days cooldown
- `com.my-company:platform-core` version 2.x keeps 2 days cooldown because the more specific override matches first


iRevive:
I just noticed another PR addresses issue #3880 😄  Since we approached it differently, feel free to choose the preferred option.

docs/repo-specific-configuration.md
```diff
@@ -192,6 +192,11 @@ dependencyOverrides = [
     dependency = { groupId = "com.example", artifactId = "foo", version = "2." },
     pullRequests = { frequency = "1 day" },
   },
+  {
+    dependency = { groupId = "com.my-company" },
+    pullRequests = {},
+    cooldown = { minimumAge = "0 days" },
```
kubukoz: I like this config style more 👍

kubukoz: but I don't think "0 days" is a good idea even in internal, "controlled" repos. I was going to go with something low, but not _that_ low.

For example, consider this situation:

- you have an internal library depending on http4s
- http4s releases an infected patch release
- you update your internal library manually an hour later and publish
- later today / tomorrow, users of your internal library are infected

so I think it'd be good for us to document a slightly higher amount (since documentation is implicitly a recommendation). Even as little as 1 day can potentially save you from a disaster.

kubukoz:
@iRevive I like your proposal more, can you just add a little bit more detail in the PR description? just for future generations.

mzuehlke:
I am waiting for feedback from @rtyley  because he contributed the base feature

modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
```diff
@@ -87,10 +95,44 @@ final case class UpdatesConfig(
...
-      .getOrElse(Right(update))
+  private def isTooRecent(
+      update: ArtifactUpdateCandidates,
+      currentTime: Timestamp,
+      dependencyOverrides: List[GroupRepoConfig]
+  ): FilterResult =
+    update
+      .filterVersionsWithFirstSeen(
+        keepVersionForCooldown(_, update, currentTime, dependencyOverrides)
+      )
```
rtyley, line 106: Note that this change makes the method `CooldownConfig.filterForAge()` superfluous, as it's no longer called anywhere in implementation code.

This makes sense, as the `filterForAge()` method assumed that all different versions of an artifact would have the same cooldown age:

https://github.com/scala-steward-org/scala-steward/blob/8b69e839642d1847f093777f7cf99b405ceaaf07/modules/core/src/main/scala/org/scalasteward/core/repoconfig/CooldownConfig.scala#L31-L35

...but now we have `dependencyOverrides` we can no longer assume that - and so the method definition should be removed, along with the whole of `CooldownConfigTest`.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
```diff
@@ -87,10 +95,44 @@ final case class UpdatesConfig(
...
+          UpdatePattern.findMatch(List(groupRepoConfig.dependency), update, include = true)
+        Option
+          .when(
+            matchResult.byArtifactId.nonEmpty && matchResult.filteredVersions.contains(
+              versionWithFirstSeen.version
+            )
+          )(groupRepoConfig.cooldown)
+          .flatten
+      }
+      .orElse(cooldown)
```
rtyley, line 135: Because the code here (trying to find out _"what's the effective config for this dependency version?"_) is so similar to the code in `PruningAlg`:

https://github.com/scala-steward-org/scala-steward/blob/19e294bd534f8b42efd46e5e4cb9df2880c0d485/modules/core/src/main/scala/org/scalasteward/core/update/PruningAlg.scala#L169-L177

...I feel like there must be an opportunity to capture some of the commonality in the code (the common stuff being handling the slightly weird interface of `UpdatePattern.findMatch`), tho' I'm not certain what it is just yet - doesn't need to be solved with this PR!

rtyley (commented):
This all looks great to me, tho' as mentioned inline, I think `CooldownConfig.filterForAge()` should be removed with this PR.

I agree that placing this config in `dependencyOverrides`, like this:

```
dependencyOverrides = [
  {
    dependency = { groupId = "com.gu" },
    pullRequests = {},
    cooldown = { minimumAge = "1 day" },
  }
]
```

...makes much more sense, given the pre-existence of `dependencyOverrides`, than where we originally proposed it in PR #3762:

```
updates.cooldown = {
  minimumAge: "7 days",
  overrides: [
    { minimumAge: "1 day", [{groupId = "com.gu"}] }
  ]
}
```

Nice work!

docs/repo-specific-configuration.md
iRevive: Done. I've updated the example and set it to 2 days.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
iRevive: Done

modules/core/src/main/scala/org/scalasteward/core/repoconfig/UpdatesConfig.scala
iRevive: I moved repetitive logic to the `UpdatePattern` helpers.

modules/core/src/main/scala/org/scalasteward/core/repoconfig/RetractedArtifact.scala
```diff
@@ -27,10 +27,7 @@ final case class RetractedArtifact(
 ) {
   def isRetracted(updateSingle: Update.Single): Boolean =
     updateSingle.forArtifactIds.exists { updateForArtifactId =>
-      UpdatePattern
-        .findMatch(artifacts, updateForArtifactId, include = true)
-        .filteredVersions
-        .nonEmpty
+      artifacts.exists(UpdatePattern.matches(_, updateForArtifactId))
```
rtyley: I think unfortunately, while this is more concise, which I love, it's not an improvement in all ways, as `UpdatePattern.findMatch()` _already_ takes a `List[UpdatePattern]` - so it's a bit inefficient to instead iterate with `artifacts.exists()`, then individually wrap each entry in a `List` in `UpdatePattern.matches()`, before handing it back to `UpdatePattern.findMatch()` to look through several single-element `List`s.

Thanks for taking a look at my comment in https://github.com/scala-steward-org/scala-steward/pull/3881/changes#r3249788900, but I think we don't need to solve the problem in this PR!

So, I'm very happy to merge this PR if we can just drop this last commit (7869defcf2dcf137fa294a6f545bb09b8a7c2e5c)  for now - everything else looks great, thank you!

iRevive: Done

rtyley (approved):
Great stuff! I'll merge this now as I don't think we've seen any objections.

kubukoz:
@rtyley can we get this in a release?

rtyley:
> @rtyley can we get this in a release?

Sure, this is now [released](https://github.com/scala-steward-org/scala-steward/actions/runs/26299513579/job/77421521839) as [v0.38.1](https://github.com/scala-steward-org/scala-steward/releases/tag/v0.38.1)!

Just so you know, for every commit merged to the `main` branch of Scala Steward, an _automatic snapshot is immediately released_ - for instance, on 9f2e5067797ae1a67d1faa63fdbb226a10e952b3 (the merge commit for this PR) the [`Publish Artifacts`](https://github.com/scala-steward-org/scala-steward/actions/runs/25989393554/job/76392924554#step:14:65) GHA job published `0.38.0-90-9f2e5067-SNAPSHOT`:

> [info] 	published scala-steward-core_2.13 to https://central.sonatype.com/repository/maven-snapshots/org/scala-steward/scala-steward-core_2.13/0.38.0-90-9f2e5067-SNAPSHOT/scala-steward-core_2.13-0.38.0-90-9f2e5067-SNAPSHOT.jar

If you're using [`scala-steward-action`](https://github.com/scala-steward-org/scala-steward-action) you could use the `scala-steward-version` setting to try out that specific version, like in this PR:

* https://github.com/guardian/scala-steward-public-repos/pull/55

Eg:

```yaml
        uses: scala-steward-org/scala-steward-action@v2.88.0
        with:
          scala-steward-version: 0.38.0-92-cd58e585-SNAPSHOT
```

PR #3895, opened 2026-06-12
Update doobie migration scalafix rule

iRevive:


iRevive:
This one should resolve properly.
cc @jatcwang

mzuehlke:
The path is still wrong
```bash
❯ scalafix --verbose --rules "github:jatcwang/doobie-package-rename-scalafix/DoobiePackageRename?sha=v0.1.2"
error: 404 - not found https://raw.githubusercontent.com/jatcwang/doobie-package-rename-scalafix/v0.1.2/scalafix/rules/src/main/scala/fix/DoobiePackageRename.scala
```

according to the expansion rule for `github` https://github.com/scalacenter/scalafix/blob/main/docs/developers/tutorial.md#using-github
the folder for the file must be different.

If you don't want to move the file around you can use the complete URL
like here https://github.com/iRevive/scala-steward/blob/2f7349ee673993dc9a1539c1689cd597edbef5d1/modules/core/src/main/resources/scalafix-migrations.conf#L32

This would be 
```
https://raw.githubusercontent.com/jatcwang/doobie-package-rename-scalafix/refs/tags/v0.1.2/rules/src/main/scala/fix/DoobiePackageRenameScalafix.scala
```

```bash
❯ scalafix --verbose --rules "https://raw.githubusercontent.com/jatcwang/doobie-package-rename-scalafix/refs/tags/v0.1.2/rules/src/main/scala/fix/DoobiePackageRenameScalafix.scala"
error: No files to fix
```

iRevive:
The https variant should work, right?

mzuehlke:
> The https variant should work, right?

I copy pasted it into the scalafix CLI and it find the rule 👍

mzuehlke (approved):

PR #3904, opened 2026-06-19
Refactor build file. Avoid sbtcrossproject

xuwei-k:


exoego (commented):
Please describe a motivation in PR decription for this change.
It is not obvious to me.

xuwei-k:
https://github.com/portable-scala/sbt-crossproject

> Cross-platform compilation support

scala-steward is JVM only project. I think we should use just `sbt.Project` instead of `CrossProject`.

xuwei-k:
This is a blocker for sbt 2.x migration.

exoego (approved):
context: https://github.com/portable-scala/sbt-crossproject/issues/170

I don't find a good reason to make Scala Steward cross project right now.
Native version could be a performance boost but require much effort seemingly, so such can be added when needed.

PR #3918, opened 2026-07-01
Add --dry-run mode

ptrdom:
## What

Adds a `--dry-run` command-line flag that runs the full Scala Steward
pipeline — cloning repositories, applying updates, and computing diffs —
but skips every write to the forge and to the local pull request cache.

With `--dry-run`, Scala Steward will not:

- push branches or delete remote branches,
- create, update, or close pull requests,
- comment on pull requests, or
- create forks (`--dry-run` implies `--do-not-fork`).

Actions that would otherwise be taken are logged with a `[dry-run]`
prefix. This is useful for verifying configuration or previewing what
Scala Steward would do without making any changes.

## How

The write suppression is implemented with decorators wired in `Context`
when `--dry-run` is set, keeping the change isolated and cross-compiling
on Scala 2.13 and 3:

- `DryRunForgeApiAlg` decorates `ForgeApiAlg`, neutralizing
  `createPullRequest`, `updatePullRequest`, `closePullRequest`,
  `commentPullRequest` and `createFork` while delegating all reads.
- `DryRunKeyValueStore` decorates the pull-request `KeyValueStore` so the
  local PR cache is never written.
- `NurtureAlg` skips `gitAlg.push` and `gitAlg.deleteRemoteBranch` under
  dry-run, logging what it would have done.

Unit tests cover both decorators and the new CLI flag; docs (`help.md`
and the running guide) are updated.

## Note

This PR also includes a small unrelated fix to `MockConfig` (test-only):
the repos-file `Uri` is now built via `path.toUri` instead of directly
from a filesystem path, so the mock test config initializes on Windows,
where a raw path such as `C:\...\repos.md` is not a valid URI. It is a
separate commit and can be dropped if you'd prefer it in its own PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -4,7 +4,7 @@ All command line arguments for the `scala-steward` application.
 
 ```
 Usage:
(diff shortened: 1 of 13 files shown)


fthomas:
How differs this from setting the `updates.limit` repo config to 0: https://github.com/scala-steward-org/scala-steward/pull/1828 ?

ptrdom:
Aha, I really tried to search for a dry-run-like behavior before implementing this PR, but did not discover this existing option. If I understand correctly, the `dry-run` implemented by this PR would do a bit more work than `updates.limit = 0`, that is it actually tries to create branch, apply version bumps, run migrations, commit changes. Are we good to proceed with this PR, or do you think the existing option is good enough?

PR #3922, opened 2026-07-07
add test proving that pin will filter updates attempt on scala 2 to v…

loicl-bbc:
shows that the pin to scala 2.13 disallow updates in scala 3.???


loicl-bbc:
sorry was meant to push to my clone

PR #3971, opened 2026-08-30
feat: update generated GitHub workflows instead of regenerating

raboof:
This is useful for projects that want to refer to actions by hash instead of tag (and manage that through for example dependabot), but still want to use sbt-github-actions to manage the rest of the workflow.

This is a [new feature](https://github.com/sbt/sbt-github-actions/pull/230) in sbt-github-actions 0.32.0

### DIFF
--- a/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
+++ b/modules/core/src/main/scala/org/scalasteward/core/edit/hooks/HookExecutor.scala
@@ -32,6 +32,7 @@ import org.scalasteward.core.git.{gitBlameIgnoreRevsName, Commit, CommitMsg, Git
 import org.scalasteward.core.io.process.SlurpOptions
 import org.scalasteward.core.io.{FileAlg, ProcessAlg, WorkspaceAlg}
 import org.scalasteward.core.repocache.RepoCache
+import org.scalasteward.core.repoconfig.WorkflowTask
 import org.scalasteward.core.scalafmt.{scalafmtArtifactId, scalafmtGroupId, ScalafmtAlg}
 import org.scalasteward.core.util.Nel
 import org.scalasteward.core.util.logger.*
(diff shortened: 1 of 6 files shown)


samueleresca (approved):

exoego (commented):
`githubWorkflowGenerate` regenerates workflow file:
```diff
-        java: [temurin@17]
+        java: [temurin@17, temurin@21]
       - name: Checkout current branch (full)
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v7.0.0
+        uses: actions/checkout@v7
       - name: Setup Java (temurin@17)
-        uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v6.0.0
+        uses: actions/setup-java@v6
+      - name: Setup Java (temurin@21)
+        uses: actions/setup-java@v6        ← 新規ステップはタグ参照
       - name: Setup sbt
-        uses: sbt/setup-sbt@3f1ce6b7f30ee47d0687b53d4bf5087b6f8bb45e # v1.1.11
+        uses: sbt/setup-sbt@v1
```

while `githubWorkflowUpdate` keeps the unchanged parts:
```diff
-        java: [temurin@17]
+        java: [temurin@17, temurin@21]
+      - name: Setup Java (temurin@21)
+        uses: actions/setup-java@dded0888837ed1f317902acf8a20df0ad188d165 # v6.0.0
```

This is generally good in terms of review.

To adopt this without breaking sbt-typelevel and older-plugin repos, consider making the task configurable per repo, defaulting to `githubWorkflowGenerate`.

The existing `enabledByConfig` mechanism (see `scalafmtHook`) supports this with minimal change. Define two built-in hooks and let `.scala-steward.conf` pick
one, e.g. `sbtGithubActions.workflowTask = "update"`.

Since generate and update produce identical output for repos without pinned hashes, keeping generate as default harms nobody, while hash-pinning repos opt in with one line. The default could flip later, once sbt-typelevel ships the task and pre-0.32.0
versions age out.

raboof:
Thanks for having a look! I've updated the PR - is this the general direction you were thinking of?

(not 'actually' tested yet, just in the unit test)

exoego:
@raboof Yes, this is the direction I expects

raboof:
TBH I found it difficult to test my changes. I _think_ it's not picking up the new option when I add `sbtGithubActions.workflowTask = "update"` to my `.scala-steward.conf`. Could you have a look if you can spot what might still be missing?

raboof:
> TBH I found it difficult to test my changes. I _think_ it's not picking up the new option when I add `sbtGithubActions.workflowTask = "update"` to my `.scala-steward.conf`. Could you have a look if you can spot what might still be missing?

Ah no I think it's taking the `.scala-steward.conf` from the wrong commit. I'm going to leave this for a bit, would appreciate if you could have a look :)

PR #3973, opened 2026-09-03
Handle Scala 3.9.x as LTS

mzuehlke:
The LTS versions are hard coded.
so we need to add 3.9 as the latest LTS version


modules/core/src/test/scala/org/scalasteward/core/update/FilterAlgTest.scala
```diff
@@ -343,7 +356,7 @@ class FilterAlgTest extends FunSuite {
     assert(isScala3Lang(update33_24))
 
     val update37_38 =
-      ("org.scala-lang".g % ("scala-library", "scala-library_3").a % "3.8.4" %> Nel.of(
+      ("org.scala-lang".g % ("scala-library", "scala-library_3").a % "3.7.4" %> Nel.of(
         "3.8.1"
```
rtyley, line 360: Looks like a good test fix!

exoego (approved):
LGTM

rtyley (approved):

PR #3987, opened 2026-09-11
Force explicit custom labels, keeping optional only derived labels

kitbellew:
If a repository has explicitly specified labels to be applied, via `pullRequests.customLabels`, let's honor that, regardless of whether `--add-labels` was used.


mzuehlke (approved):
I like the idea to add label, when the repo asks for them

PR #3992, opened 2026-09-14
Do not update from Scala 2.13 to Scala 3

mzuehlke:
With the relesase of the new SCala 3 LTS and the return of the scala-library artifact (in version 3.8.0) we are seeing a huge amount of PR to update Scalla 2.13.18 to Scala 3.9.0.

This step that cab't be done automatically. So in nearly all cases these PRs are noise. This change prevents theres PRs.


exoego (approved):
LGTM

PR #3995, opened 2026-09-16
Add artifact migrations for the Prometheus simpleclient modules

Dichotomia:
The Prometheus Java client was rewritten for its `1.0.0` release and its Maven modules
changed coordinates. The old ones stopped receiving releases at that point:

| coordinate | latest | published |
|---|---|---|
| `io.prometheus:simpleclient` and friends | `0.16.0` | 2022-06-15 |
| `io.prometheus:prometheus-metrics-*` | `1.x`, active | since 2023-09-27 |

Without a migration entry there is nothing newer under the old coordinates, so a build
still depending on `simpleclient` never hears about the client again — which is how a
project can sit on a 2020 release of its metrics library without a single pull request
drawing attention to it.

The four renames added here are the ones the upstream migration guide states explicitly
(`docs/content/migration/simpleclient.md`):

- `simpleclient` -> `prometheus-metrics-core`
- `simpleclient_hotspot` -> `prometheus-metrics-instrumentation-jvm`
- `simpleclient_httpserver` -> `prometheus-metrics-exporter-httpserver`
- `simpleclient_servlet_jakarta` -> `prometheus-metrics-exporter-servlet-jakarta`

I deliberately stopped there. The 0.x line had many more modules, and mapping them from
the directory listing alone would be guesswork: some have no 1.x counterpart at all
(`simpleclient_graphite_bridge`, `simpleclient_hibernate`, `simpleclient_spring_boot`,
`simpleclient_vertx`), and for the ones that look like they do, the guide does not say
so. A wrong migration entry points every build at a coordinate that was never the
successor, which is worse than a missing one.

One thing worth stating plainly, as with any rename across a rewrite: the update this
enables is `0.x` -> `1.x`, a breaking API change — the metric builders, the registry and
the exposition format writer all moved. The pull requests Scala Steward opens from it
will not compile as-is. Upstream ships `prometheus-metrics-simpleclient-bridge` for
projects that need a staged transition.

Checked before opening:

- no `prometheus` or `simpleclient` entry exists in `artifact-migrations.v2.conf`
- the entries follow the artifactId-only shape already used for the `play-*` renames,
  since the groupId is unchanged
- the version ordering across the rename is correct: `Version.versionOrder` compares
  rank by rank, so `[0, 16, 0]` vs `[1, x, y]` is decided on the first component
- `ArtifactMigrationsLoaderTest` asserts `migrations.size > 0` for the defaults rather
  than an exact count, so no test needed updating

One consequence of stopping at the documented four, worth weighing: a build that declares
`simpleclient_common` alongside one of the migrated modules migrates **partially** — the covered
artifact moves to `1.x` while `simpleclient_common` stays on `0.x`. Both generations then sit on the
classpath with separate registries (`io.prometheus.client.CollectorRegistry` versus
`io.prometheus.metrics.model.registry.PrometheusRegistry`), so metrics registered by the migrated
module stop appearing on an endpoint that exports the other one.

That state is supported rather than broken — upstream ships `prometheus-metrics-simpleclient-bridge`
for exactly this, and `SimpleclientCollector.builder().register()` bridges the two registries — but
it is a manual step, so the update this entry enables is one a human has to finish. That seems to me
the normal shape of a migration across a rewrite rather than a reason to withhold the entries, and
`simpleclient_common` has no documented successor because the `1.x` exposition API is a redesign, not
a rename. Inventing a mapping for it would be worse than leaving it out. Happy to adjust if you see
it differently.

Migration guide the four renames come from:
https://github.com/prometheus/client_java/blob/v1.9.0/docs/content/migration/simpleclient.md


exoego (approved):
https://prometheus.github.io/client_java/migration/simpleclient/#refactoring-the-instrumentation-code

LGTM

